PR #2, opened 2015-11-14
Updated library files

OmgImAlexis:
More library files should now be in bower so we shouldn't have to worry about maintaining a newer version of each file ourselves.


OmgImAlexis:
@miigotu @fernandog please don't merge until I'm done.

PR #10, opened 2015-11-15
update confirmations and move to SICKRAGE

OmgImAlexis:


gui/slick/js/core.js
```diff
@@ -24,6 +24,49 @@ var configSuccess = function(){
...
+                dialogClass: "modal-dialog",
+                post: false,
+                confirm: function(e) {
+                    location.href = e.context.href;
+                }
+            };
+
+            $("a.shutdown").confirm({
+                title: "Shutdown",
+                text: "Are you sure you want to shutdown SickRage?"
```
alisonatwork, line 39: This is totally minor, but I guess this should probably be "shut down" when used as a verb.

OmgImAlexis, line 39: [Shutdown isn't a verb](http://notaverb.com/shutdown), I thought the same thing when I was moving it all over to the new confirm setup.

PR #12, opened 2015-11-15
add some unit tests for exceptions helper

alisonatwork:
Well that was rather annoying. I also hit this issue: https://github.com/SiCKRAGETV/sickrage-issues/issues/2358 and could not figure out what was happening. So I started writing unit tests for this class, until I realized the problem happened earlier. Here's how we can recreate it:

```
class Dummy():
    def __str__(self):
        return DummyException()

class DummyException(Exception):
    pass
```

...

```
    def test_bad_string_issue_2358(self):
        try:
            o = Dummy()
            print 'Error : %s' % o
        except TypeError as e:
            self.assertEqual(ex(e), u'lala')
```

I figured we can keep the unit tests instead of throwing them away... Next step is to figure out which string interpolation in helpers.py is getting the object with the broken **str** method. Sigh...

### DIFF
--- a/tests/exceptions_helper_tests.py
+++ b/tests/exceptions_helper_tests.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+
+import sys, os.path
+sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '../lib')))
+sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+
+import unittest
+
+import test_lib as test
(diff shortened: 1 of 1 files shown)


alisonatwork:
Btw I am kind of wondering if we should replace that empty string with at the least the exception classname. Because log entries like this are really unhelpful:

> 2015-11-15 20:14:13 ERROR    POSTPROCESSER :: Exception generated in thread POSTPROCESSER:

miigotu:
:+1:  for more unit tests, will merge after football probably

PR #13, opened 2015-11-15
Fix statusstrings

labrys:
##### Code Changes:
- 21acc29 Change from UserDict to NumDict and add unittests for NumDict and new statusstrings
- ac46158 Streamline common tests and PEPs

This fixes issues with recursion in the UserDict implementation of statusstrings, adds the unit tests required, is Python 3 compatible, and NumDict can be further leveraged to streamline common.Quality in the future.


miigotu:
I like this one much better! Thanks!

PR #34, opened 2015-11-16
Convert to new style class with explicit object subclassing

miigotu:
PEP 203: Replace dict creation with dict literal

PEP 263: Add encoding declaration

Replaces #20


miigotu:
@labrys I think you need to resync your fork before more PR lol

Command line would be:

``` bash
git remote upstream set-url git@github.com:SickRage/SickRage.git
git fetch upstream
git checkout develop
git reset upstream/develop --hard
git push origin develop -f
```

PR #63, opened 2015-11-17
update js references

OmgImAlexis:


OmgImAlexis:
@MGaetan89 can you please lint your apiBuilder, I was trying to do it myself but you've got episodes and commands set outside of the js file.

I'm also not sure how it'll go with using core.js as it doesn't have a controller or action set on the body tag unlike layouts/main.jade

PR #86, opened 2015-11-18
nameparser fix

CristianBB:
PLEASE, DON'T MERGE UNTIL YOU CAN TEST IT (It can cause a real chaos)

This code avoid to use ALL_REGEX when we know type of show to parse,
because some shows are being parsed incorrectly (anime shows being
parsed as normal shows and vice versa). This problem will affect later
to scoring system, giving invalid results.

SR matching works as follows:
- Firstly, SR has a show object with episode searched
- Secondly, for each result obtained from each provider, SR tries to
  construct an object of type show
- If SR has succes building the object, SR compare it with original show
  object. If they are equal gives it as valid result.

For that reason we never should send the original show object to
NameParser method (we learned it by hard way...)

This modification allow us to tell NameParser what type of show we're
parsing from findSearchResults method (generic.py) without affect others
calls to this method (there are several calls to NameParser method...)

Please be careful with this, I'm pretty sure that it's working fine but
would be great if somebody else could test it.

@miigotu @duramato @fernandog could you take a look?

### DIFF
--- a/sickbeard/providers/generic.py
+++ b/sickbeard/providers/generic.py
@@ -364,12 +364,7 @@ class GenericProvider(object):
 
             # parse the file name
             try:
-                if show.is_anime:
-                    method = 'anime'
-                else:
-                    method = 'normal'
-                    
-                myParser = NameParser(parse_method=method)
+                myParser = NameParser(parse_method=('normal', 'anime')[show.is_anime])
                 parse_result = myParser.parse(title)
             except InvalidNameException:
                 logger.log(u"Unable to parse the filename " + title + " into a valid episode", logger.DEBUG)
--- a/sickbeard/providers/generic.py
+++ b/sickbeard/providers/generic.py
@@ -364,7 +364,12 @@ class GenericProvider(object):
 
             # parse the file name
             try:
-                myParser = NameParser(False)
+                if show.is_anime:
+                    method = 'anime'
+                else:
+                    method = 'normal'
+                    
+                myParser = NameParser(parse_method=method)
                 parse_result = myParser.parse(title)
             except InvalidNameException:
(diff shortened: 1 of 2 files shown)


sickbeard/providers/generic.py
```diff
@@ -364,7 +364,12 @@ def findSearchResults(self, show, episodes, search_mode, manualSearch=False, dow
 
             # parse the file name
             try:
-                myParser = NameParser(False)
+                if show.is_anime:
+                    method = 'anime'
+                else:
+                    method = 'normal'
+                    
+                myParser = NameParser(parse_method=method)
```
miigotu: `myParser = NameParser(parse_method=('normal', 'anime')[show.is_anime])`

CristianBB: more elegant :)

miigotu:
Will try and put together some unittest cases for this.

CristianBB:
@miigotu Sure,
It happens with most results. For example for show "The Flash", Kickass as only provider, searching for episode 10, temp 1. First result: "The.Flash.2014.S01E10.Revenge.of.the.Rogues.1080p.WEB-DL.DD5.1.H.264-YFN"

On _parse_string method (from parser.py) you can see how it's being parsed on original code using regex: "standard", "season_only", "bare", "anime_and_normal" and "anime_bare"

Then scoring system assign best result using "standard" regex, that's ok in that case only because this regex has obtained more info than others, however it's not always ok (I tried to search some example but I don't remember what shows are problematic, If I find one I'll tell you)

With that changes, the only valid regex from _parse_string method are "standard", "season_only" and "bare". Then scoring system gives "standard" as best result.

CristianBB:
From SickRageTV, same problem occurs with https://github.com/SiCKRAGETV/SickRage/pull/3091 
This code should fix it

PR #87, opened 2015-11-18
Added a small `v2` so users can see the difference with the old SR

neoatomic:


WebSpider:
I really dont agree with this, but I'm going to have to make time to go to irc and discuss this over.

This is no new and improved SR, it's the same SR it has for a while now.

labrys:
I think I agree with you @WebSpider.  It's a change of repo location, but not a change to SR.

miigotu:
@WebSpider problem is that users should have an easy way to tell if they successfully switched or not. With the fact that echelon as locked the channel, turned off the issue tracker, and even had the forum in maintenance mode for awhile is really giving them no choice.

Along with permanently banning us (at least me) from the forums, and spamming things like [this](https://github.com/SiCKRAGETV/sickrage-issues/commit/e1499a2bda78e7898465739eaaaf631feafbb215) everywhere.

To lie so terribly and then do everything possible to prevent us from telling our side that's just wrong.

We would really like to hear from you on irc, because ech really thinks you are on his side with this and that's not what we got from you the other night. It's sad it had to come to this, we tried avoiding it but this is where we are now and there is no going back.

labrys:
I was more than willing to contribute to both until echelon started posting the stuff about viruses and closing the forums and stuff.

miigotu:
Yeah there just is no middle ground now, you can really only ride one horse. I just want users to stop being neglected and lied to because of this.

Arcanemagus:
The "right" thing would probably be a new name/logo, but that will take more time and I think I agree with what was posted elsewhere in planning that for a much later time after everything has settled.

distracto:
soooo those of us on QNAP running from a .qpkg, how would we switch?

miigotu:
@distracto we are working really hard to get the install packages updated, but if you go look in #3 and #6 there is enough info the get you through it hopefully

PR #106, opened 2015-11-20
Per Show Notifications for Prowl

Amadeus-:
This commit would add per-show Prowl notifications, similar to what we currently have for per-show email notifications.   The current Prowl API setting would serve as a global Prowl API key (i.e. it would work exactly as it did before.)    I also did some clarification edits to the configuration pages to better explain how the settings work for email and Prowl per-show notifications.

To add per-show Prowl notifications possible without changing the DB structure, it was necessary to utilize the notify_list column in the tv_shows table in a different way than it was before.   I believe it is a good solution that will be easily expandable into the future.    The code was written in such a way as to ensure that nothing is broken for current email per-show notification users.

I've tested it over the last couple of days and think that it's working as intended.  I will be keeping a close watch on it.

Other notes/thoughts:
- Added new configuration option to set the 'title' of the notification sent to Prowl
- Added proper sorting to the shows drop-down selection box for both email and prowl per-show-notifications.
- You will notice that I changed some of the logging in the prowl and email notification python files to DEBUG.   This is because it will be pretty typical for a user to want to have per-show notifications without a "global" notification.   Therefore, we don't need to have a constant warning for every show, just because a user is using notifications, but doesn't have a notification assigned for that particular show.


gui/slick/js/core.js
```diff
@@ -942,11 +943,18 @@ var SICKRAGE = {
                 var key = parseInt($('#email_show').val(), 10);
                 $('#email_show_list').val(key >= 0 ? notify_data[key.toString()].list : ''); // jshint ignore:line
             });
+            $('#prowl_show').change(function() {
```
OmgImAlexis: Anything to do with change, click, etc. needs to be using an on click event like so.

```
$('#prowl_show').on('change', function() {
```

gui/slick/js/core.js
```diff
@@ -942,11 +943,18 @@ var SICKRAGE = {
...
+            $('#prowl_show').on('change', function() {
+                var key = parseInt($('#prowl_show').val(), 10);
+                $('#prowl_show_list').val(key >= 0 ? notify_data[key.toString()].prowl_notify_list : ''); // jshint ignore:line
+            });
 
             // Update the internal data struct anytime settings are saved to the server
             $('#email_show').bind('notify', function() {
                 loadShowNotifyLists();
             });
+            $('#prowl_show').bind('notify', function() {
```
OmgImAlexis, line 955: This should be on instead of bind. The email_show one got changed in a PR I did before.

gui/slick/js/core.js
Amadeus-: Thanks!    Yea, I just copied what was there for the #email_change, etc. from before you made the change to .on() recently.     

I committed a fix.

miigotu:
Cant add BOM and change file permissions, must fix that first too.

miigotu:
Label spam! xD

PR #107, opened 2015-11-20
File browser now allows typing a path

VinceVal:
Also added showBrowseButton option to $.fn.fileBrowser().

### DIFF
--- a/gui/slick/js/browser.js
+++ b/gui/slick/js/browser.js
@@ -6,7 +6,8 @@
             title:             'Choose Directory',
             url:               srRoot + '/browser/',
             autocompleteURL:   srRoot + '/browser/complete',
-            includeFiles:      0
+            includeFiles:      0,
+            showBrowseButton:  true
         }
     };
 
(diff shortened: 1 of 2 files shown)


gui/slick/js/browser.js
```diff
@@ -34,7 +35,20 @@
             data = $.grep(data, function (value) {
                 return i++ !== 0;
             });
-            $('<h2>').text(first_val.current_path).appendTo(fileBrowserDialog);
+
+            $('<input type="text" class="form-control input-sm">')
```
OmgImAlexis, line 39: This should be something like this.

```
$('input[type="text"].form-control.input-sm">')
```

gui/slick/js/browser.js
```diff
@@ -166,13 +180,19 @@
...
-            var initialDir = options.field.val() || (options.key && path) || '';
-            var optionsWithInitialDir = $.extend({}, options, {initialDir: initialDir});
-            $(this).nFileBrowser(callback, optionsWithInitialDir);
-            return false;
-        }));
+        options.field.addClass('fileBrowserField');
+        if (options.showBrowseButton) {
+            // append the browse button and give it a click behaviour
+            options.field.after(
+                $('<input type="button" value="Browse&hellip;" class="btn btn-inline fileBrowser">').click(function () {
```
OmgImAlexis: Same for here.

This should also be using `on('click', function() {` instead of `click(function () {`.

fernandog:
What if path does not exist?

VinceVal:
@fernandog If the path doesn't exist the AJAX callback will climb the tree until it finds one that does.

PR #111, opened 2015-11-20
Removed the original "Boxcar" notification settings/functionality.

Amadeus-:
In 2014, Boxcar went from version 1 to version 2.   SickBeard added support for both versions.  Today,
v1 no longer exists; v2 is the only one that is active, available, and functional.

Technical Details:
- Removed configuration code and settings for the original "Boxcar" notification
- The internal code and setting names for what remains is still called "Boxcar2" (in order to preserve
  settings); however, the labels that end-users see on the web-interface now simply say "Boxcar".


Amadeus-:
To clarify, if you go to https://boxcar.io/, you will see that the only option is the app "Boxcar 2"   (The original Boxcar is long gone and no longer supported.)

Personally, I think it's fine for us to call it "Boxcar" in the notifications settings (especially since clicking on the logo takes you to the site); however, if anyone felt strongly that it should be named "Boxcar 2" in the settings, I'm happy to make that change.   Either way works for me.

But, regardless of that point, the original Boxcar stuff should be removed.

OmgImAlexis:
@miigotu can you please add this to news or something so people know what's going on. @Amadeus- I'm guessing people using boxcar2 will have to retype their details or are the fields the same?

Amadeus-:
No, everything will work exactly the same as before for the people that are using Boxcar2 currently.   It will just not say "Boxcar2" on the configurations page, but _internally_, everything is working just as before.

Frankly, whoever decided to support both Boxcar v1 and v2 simultaneously chose the most confusing way possible to implement it.   So, that's why it seems confusing right now trying to fix it.   But, long-story-short, this commit is only removing the old boxcar stuff and doing a bit of renaming from the user-interface standpoint ....internally, nothing that users are CURRENTLY using will be changed.

OmgImAlexis:
Okay, this will definitely need to be added to the news section and at a later date we'll want to move everything internally to the "correct" one by dropping the 2.

Amadeus-:
Perhaps it would be better to leave it labeled as "Boxcar 2" in the web UI?    Would that make it more clear do you think?   (The iOS app is actually called "Boxcar 2" ..even though it's the only Boxcar app available.)

Changing the internal names might be more of a hassle than it's worth just leaving them as they are.   It wouldn't be a big deal in the js/mako files, of course, but might create confusion elsewhere.   Again, the way it was done originally wasn't designed very well for future changes.   Honestly, the v2 stuff should have been included in the same configuration settings, then it would have been easier-to-understand when the old stuff was removed, etc.   (In other words, there are several ways of doing it that would have made fixing it now...easier.)

OmgImAlexis:
@Amadeus- it's actually only called BoxCar2 on 2 screens I think, the app store and the initial loading screen. Everything else references it as BoxCar. I think you might have a point in keeping it as BoxCar2 at least for a little while especially as it won't hurt anything.

miigotu:
Yes, keeping it as BoxCar2 visually will prevent people from being like "wtf my boxcar2 settings disappeared"

gui/slick/views/config_notifications.mako
```diff
@@ -1074,7 +1011,7 @@
                                 <span class="component-title">Enable</span>
                                 <span class="component-desc">
                                     <input type="checkbox" class="enabler" name="use_boxcar2" id="use_boxcar2" ${('', 'checked="checked"')[bool(sickbeard.USE_BOXCAR2)]}/>
-                                    <p>should SickRage send Boxcar2 notifications ?</p>
+                                    <p>should SickRage send Boxcar notifications ?</p>
```
OmgImAlexis, line 1014: This is good because people will slowly realise it's the same thing as boxcar2.

PR #117, opened 2015-11-20
Add test for sickrage module

MGaetan89:
I added some tests for the `sickrage` module.
For the moment, those new tests are not executed on Travis, because I wasn't sure how to update `tests/all_tests.py` to include them.
Currently, I hardcoded the list of tests available in `tests/sickrage/__init__.py`.
I duplicated the `sickrage` module directory structure in the `tests` folder, so the `tests` folder does not get too full of files and it is easier to find a specific test.

Any tip on how to make Travis run those tests as well? How to remove the hardcoded list?


MGaetan89:
Well, it looks like Travis is trying to execute my new tests.
But I don't see why it can not find the file. Is it because the package name are almost the same?

miigotu:
You have to sys.path.insert for the tests. I will look at this later.

MGaetan89:
If I run the tests individually or from `tests/sickrage/__init__.py` it
works.
But from `tests/all_tests.py` it doesn't.
So I guess the `sys.path.insert` are good, but maybe not made properly?

miigotu:
@MGaetan89 you would need to modify them because you added subdirs to tests

MGaetan89:
I printed `sys.path` to check what values are inserted, and they seem to be ok.
I guess the issue is that `sys.path` contains `SickRage/tests` and `Sickrage`. So when looking for something in `sickrage.*`, it looks in `tests` first, but that's probably not where it should look. I don't know if Python looks in every path before failing ?
I will try other solutions to make it work.

miigotu:
It looks in the current parent first, then in order in the sys.path. It stops when the first one is found. so if you are inside "sickrage" and import from "sickrage" it is importing from inside itself

MGaetan89:
But it `tests/sickrage`, it can not find the correct file. So it should look in `sys.path`, where the path to the correct folder exists. So why does it fails?

This PR is getting a mess because I forgot to use a dedicated branch. I guess I will close this and resend a proper PR once it's working.

miigotu:
../.. and ../../lib need to be in the path since you have a subfolder in tests lol, otherwise anything in tests/sickrage/ cannot find sickbeard, lib, or sickrage modules

MGaetan89:
I am always adding those folders in `sys.path`: https://github.com/SickRage/SickRage/pull/117/files#diff-b19000e3602fcef243780abb885a9885R22
Or do I have to do that in `all_tests.py`?

miigotu:
idk I haven't looked into the actual code of this PR much, I'll have to pull it down and try to figure it out

miigotu:
Checked it out. Just rename tests/sickrage to tests/sickrage_tests, the tests trying to import from sickrage end up finding tests/sickrage first.

MGaetan89:
I thought about that, but I thought there could be an other solution.
I will resend a PR with the correct folder name when I'm back from work.

PR #118, opened 2015-11-21
Add abnormal.ws provider

Fen0l:
New provider


duramato:
@JohnDooe could you remake this against the develop branch? Thanks in advance!

Arcanemagus:
Also you should probably start a new branch off of develop instead of committing directly to the branch, that way it makes it a lot easier to handle situations like this as you simply resubmit the PR instead of having to do a lot of trickery with the commits.

PR #123, opened 2015-11-21
Removed unneeded sudo from .travis.yml

OmgImAlexis:


Arcanemagus:
You actually want to keep that in, unless you _need_ root rights in the build. `sudo: false` there actually _disables_ `sudo` in the build, enabling you to build on the container infrastructure which starts about 10x faster than the regular builds: http://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

MGaetan89:
@Arcanemagus :+1:

Arcanemagus:
Although looking over the docs it looks like they have updated them to say this is now the default, so you are right that this isn't strictly necessary.

OmgImAlexis:
@Arcanemagus if you check there's actually `sudo: false` already in there. There was a duplicate so I removed it :stuck_out_tongue:

Arcanemagus:
Ha! That was hidden in the diff, carry on!

PR #142, opened 2015-11-22
Enhance show search

VinceVal:
- If term contains a leading article, remove it.
- If there is a trailing year, enclose it in ().

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -2237,10 +2237,10 @@ class HomeAddShows(Home):
 
         searchTerms = [search_term]
 
-        # If search term ends with a 4 digit number (year), enclose it in ()
-        matches = re.match(r'^(.+) ([12][0-9]{3})$', search_term)
+        # If search term ends with what looks like a year, enclose it in ()
+        matches = re.match(r'^(.+ |)([12][0-9]{3})$', search_term)
         if matches:
-            searchTerms.append("%s (%s)" % (matches.group(1), matches.group(2)))
+            searchTerms.append("%s(%s)" % (matches.group(1), matches.group(2)))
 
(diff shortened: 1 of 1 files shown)


VinceVal:
@miigotu thoughts?

fernandog:
Why not search with and without year? The show name in the another language may not have the year.

OmgImAlexis:
@fernandog we don't really support anything but English so I'm not really sure how this will break anything?

fernandog:
Cant you choose language while adding show?

OmgImAlexis:
"\* This will only affect the language of the retrieved metadata file contents and episode filenames.
This DOES NOT allow SickRage to download non-english TV episodes!"

So this shouldn't effect anything like that.

fernandog:
What are you talking about?!
Im saying the Show name!
In english: The Blacklist (2014)
In french: Blacklist

Just an example

duramato:
@OmgImAlexis you are wrong it searches in the title that is in that language ,I've found myself having to add a scene exception with the English title because it searched the title in PT ..

miigotu:
I think it should be done a little different. It should first search for the inputted string as is, and if no results but contains a|an|the or \d{4}$ search again with the string fixed. Otherwise you will have shows you wouldnt be able to find anymore. Ex: Mystery Science Theater 3000

VinceVal:
Hmm, I think there is some confusion here. It actually creates a list of multiple search terms and searches for all of them. It still searches for exactly what you typed also. Per @miigotu I'll add 'a' and 'an'.

miigotu:
looks great!

miigotu:
Not so concerned with waiting for a squash on this one, just gonna merge it =P

PR #146, opened 2015-11-22
Search again if original search started with an article or ended with…

miigotu:
… a year without parens and returned no results

Necessary to retain previous precision
Closes #142


VinceVal:
We're going to use #142.

PR #152, opened 2015-11-23
Newpct: Torrent download by torrent client now works

CristianBB:
downloadResult was overwrited but it only applies when torrent download
is made by Black Hole.

To make download by torrent client works fine, we must identify when the
snatchEpisode method from search.py is trying to get content of torrent
file (needBytes=True) and apply the same logic used on downloadResult
method.


miigotu:
This provider is getting to be too much overrides. This is the last time I'm letting stuff be _added_ to it

CristianBB:
Sorry for that Miigotu, I have been testing it alone until this weekend that one user writes me to send me some issues...

miigotu:
Its not about the code its that  because they dont have a link on main search results you have to override all this stuff, and because they use wrong naming. Really a sucky provider because of all this.

CristianBB:
@miigotu I agree, in fact I'm sure that they changed the results page cause they detected lots of conexions... If they changes again the result page I think that the best option will be disable it...

I'm testing now a new provider that will be less problematic...

PR #153, opened 2015-11-23
Add "Force subtitle search" in manage

fernandog:
@duramato @medariox @miigotu


duramato:
And wab also  a pause subtitle search button?

PR #160, opened 2015-11-23
Add tests

MGaetan89:
PR following https://github.com/SickRage/SickRage/pull/117
I renamed `tests/sickrage` to `tests/sickrage_tests`, as suggested by @miigotu 
I will try to add more tests in future PRs

### DIFF
--- a/sickrage/show/ComingEpisodes.py
+++ b/sickrage/show/ComingEpisodes.py
@@ -54,77 +54,77 @@ class ComingEpisodes:
         :return: The list of coming episodes
         """
 
-        if not isinstance(categories, list):
-            categories = categories.split('|')
-
-        if sort not in ComingEpisodes.sorts.keys():
-            sort = 'date'
+        categories = ComingEpisodes._get_categories(categories)
+        sort = ComingEpisodes._get_sort(sort)
 
         today = date.today().toordinal()
         next_week = (date.today() + timedelta(days=7)).toordinal()
         recently = (date.today() - timedelta(days=sickbeard.COMING_EPS_MISSED_RANGE)).toordinal()
-        qualities_list = Quality.DOWNLOADED + Quality.SNATCHED + Quality.SNATCHED_BEST + Quality.SNATCHED_PROPER + Quality.ARCHIVED + [IGNORED]
+        qualities_list = Quality.DOWNLOADED + Quality.SNATCHED + Quality.SNATCHED_BEST + Quality.SNATCHED_PROPER + Quality.ARCHIVED + [
+            IGNORED]
 
         db = DBConnection()
         fields_to_select = ', '.join(
-            ['airdate', 'airs', 'description', 'episode', 'imdb_id', 'e.indexer', 'indexer_id', 'name', 'network',
-             'paused', 'quality', 'runtime', 'season', 'show_name', 'showid', 's.status']
+                ['airdate', 'airs', 'description', 'episode', 'imdb_id', 'e.indexer', 'indexer_id', 'name', 'network',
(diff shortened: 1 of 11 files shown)


labrys:
@MGaetan89 I've been looking at the tests myself and one thing I think we could use is some more comprehensive tests.  For example testing for things such as unicode characters in show/episode titles.   None values, even when we shouldn't be passing None, and other tests to really try to break the code so we can find issues before they hit the bug-tracker.   For example simple things like SickRage/sickrage-issues#65 failing to convert a show id to int.

MGaetan89:
I agree. I tried to use various parameters for the methods I tested in this PR to make sure the behavior is always correct. That's why I added some new methods, so it is easier to test and control the parameters. I prevented a couple of issues with those (similar to the one you linked). If your think of other test cases to include, let me know and I will include them.

labrys:
Sure thing.. haven't looked at your test yet but I'm in progress linting :grimacing: , peping :stuck_out_tongue_closed_eyes: , and documenting :sleeping: the tests so I'm sure I'll get to it eventually :grinning:

MGaetan89:
Yeah I saw your PR :)
I updated (and PEPed) `tests/all_tests.py` in this PR to support tests in subfolder. So you might want to way until this gets merged before working on that file. Then I'll probably move to test `sickbeard/common.py`.

labrys:
lol funny you mention that.. i was just in process of fixing up all_tests.py

labrys:
Also I'm working on some changes to common.py so maybe hold off on the changes to testing common.py.   I'm going to convert the Quality strings to NumDicts similar to what I did with StatusStrings except without the need to subclass it.

MGaetan89:
Ok, thanks for the heads-up.
I can start to test it and update my branch once your done and your PR is merged ;)

labrys:
Sounds good... likely won't change your tests too much yet since NumDict is still essentially a dict that does the key conversion for you, but I do have additional changes in mind but those won't be till much later.  Don't want to change to much in common.py at once.

medariox:
Quick rebase please? :)

MGaetan89:
Done :)

sickrage/show/ComingEpisodes.py
```diff
@@ -54,77 +54,77 @@ def get_coming_episodes(categories, sort, group, paused=sickbeard.COMING_EPS_DIS
...
-            'AND e.status NOT IN (' + ','.join(['?'] * len(qualities_list)) + ')',
-            [today, next_week] + qualities_list
+                'SELECT %s ' % fields_to_select +
+                'FROM tv_episodes e, tv_shows s '
+                'WHERE season != 0 '
+                'AND airdate >= ? '
+                'AND airdate < ? '
+                'AND s.indexer_id = e.showid '
+                'AND e.status NOT IN (' + ','.join(['?'] * len(qualities_list)) + ')',
+                [today, next_week] + qualities_list
```
miigotu, line 79: these hanging indents are wrong =P

MGaetan89, line 79: PyCharm did it itself, but I confess I didn't try to find why it did it :P

miigotu:
``` python
x = foo(
    bar,
    blah,
)
```

or

``` python
x = foo(bar,
        blah)
```

I like the first one, less indents help readability.

labrys:
@MGaetan89 I use pycharm and while most of the time it gets the indents right, sometimes its idea is pretty stupid (I'm thinking continuation lines here).  So it always helps to look at any changes to indentation first :)

MGaetan89:
Will change them back later ;)
I'll have to tweak the formatter configuration to prevent this.

PR #161, opened 2015-11-23
Add Bitsnoop provider

duramato:
@miigotu what do you think?


sickbeard/providers/bitsnoop.py
```diff
@@ -0,0 +1,148 @@
...
+from sickbeard.providers import generic
+
+
+class BitSnoopProvider(generic.TorrentProvider):
+    def __init__(self):
+        generic.TorrentProvider.__init__(self, "BitSnoop")
+
+        self.urls = {
+            'index': 'http://bitsnoop.com',
+            'rss': 'http://bitsnoop.com/search/video/',
```
miigotu: This is really the search url, rss is:
`http://bitsnoop.com/browse/video-tv/?sort=dt_reg`

I will mess with this soon, want to remove xmltodict also

OmgImAlexis:
@duramato can you please update this so it can be merged.

miigotu:
I need to make some changes to it, it isnt ready for merge even if it is rebased.

sickbeard/providers/bitsnoop.py
```diff
@@ -0,0 +1,141 @@
...
+
+                            title = item.title.text
+                            assert isinstance(title, unicode)
+                            # Use the torcache link bitsnoop provides,
+                            # unless it is not torcache or we are not using blackhole
+                            # because we want to use magnets if connecting direct to client
+                            # so that proxies work.
+                            download_url = item.enclosure['url']
+                            if sickbeard.TORRENT_METHOD != "blackhole" or 'torcache' not in download_url:
+                                download_url = item.find('magneturi').next.replace('CDATA', '').strip('[]')
```
duramato, line 88: @miigotu what about getting the magnet from the torrent page (like i had it)? that way you get the magnet with trackers
becouse getting it from the rss its a trackerless magnet

miigotu, line 88: Then you need to add our tracker list, because getting a second page causes the search to be unbearably slow and also DDoS the provider.

PR #165, opened 2015-11-24
Enable filtering shows in poster layout

VinceVal:
- Rename .show class to .show-container to avoid conflict with
  bootstrap's .show class.
- Remove name search box from all other layouts since you can use the
  filter row in the table.
- Rename 'Reset Search' button to 'Clear Filter(s)'.


VinceVal:
@OmgImAlexis please take a look at this. It touches core.js.

miigotu:
Might need to remove the option to disable filter row if removing the search from table layouts.

VinceVal:
@miigotu didn't realize that was an option. I can put the search box back, or we can remove the option. I'd almost vote for removing the option to keep things simple.

miigotu:
Im always for less options =P. I see it like if something really NEEDS to be an option it better be super necessary to have it at all. This is necessary, the option isnt.

gui/slick/js/core.js
```diff
@@ -1792,6 +1792,17 @@ var SICKRAGE = {
                 $('table').trigger('filterReset');
             });
 
+            $('#filterShowName').on('input', _.debounce(function (e) {
+                $.each([$('#container'), $('#container-anime')], function () {
```
OmgImAlexis: I think we may want to change this to `$('#container, #container-anime')` as it may be faster but I'm not sure if isotope needs to be bound per instance of a DOM element or if it can handle multiples.

OmgImAlexis:
Other than the possible change I commented on it looks good.

miigotu:
@OmgImAlexis you made conflicts =P

VinceVal:
@OmgImAlexis I changed to using the comma in the selector for isotope and it seems to work.

VinceVal:
Feedback on 955cf4641dcf8ba38e475bca1102a5e8689eb015 which enables FILTER_ROW all the time and removes the option. @OmgImAlexis this touches core.js too.

OmgImAlexis:
@VinceVal was there a reason we even had that option? I'm still not sure what it even did/does.

labrys:
@OmgImAlexis If i'm not mistaken that's the checkbox that hides the filter row in case you don't want the ability to filter.  IMO if you don't want a filter, don't use it, so the option is useless.

OmgImAlexis:
Oh that's a weird option, I guess this can be merged then.

rmbleeker:
I used to use this option to disable the filter row, because I find it visually annoying that it pops up and then disappears again whenever I move my mouse from the show list to the top navigation bar. Can someone please explain to me why it was removed? I'm all for "greater good", but I also like to have options :-)

PS.: I just moved from the old SiCKRAGETV repo to this one today, after my old SickRage installation wasn't working anymore. Can someone perhaps pm me or provide a link to an article explaining what the deal is? As I'm sure you're aware the other repo is still up on Github and claiming nobody should move because they are the official one.

miigotu:
https://github.com/SickRage/sickrage-issues/issues/3

rmbleeker:
Thanks for the link, Google was no help. I'm glad to see you guys up and running again, keep up the good work and I wish you all the best for 2016 and beyond.

As for the filter_row option, it's not a big deal to me just a small UI gripe.

PR #167, opened 2015-11-24
Use is/is not for comparing to True/False/None

miigotu:
Some of these dont even need explicitly compared to True/False/None. Might fix some of those up before merge.


sickbeard/webapi.py
```diff
@@ -862,11 +859,11 @@ def run(self):
         ep_list = []
         if self.e:
             epObj = showObj.getEpisode(self.s, self.e)
-            if epObj == None:
+            if not epObj:
```
miigotu: Pretty sure this FIXES a bug!

when you getEpisode and it isnt found it returns a string iirc, need to check if it is an empty string or not though, not sure, but I think it is.

PR #168, opened 2015-11-24
Add custom_url property to torrent providers

Amadeus-:
This commit will add a "custom_url" property for torrent provider modules.

Although the configuration page will provide examples as to the appropriate format for the URL provided (protocol://IP[:port]/), it will be the responsibility of the module to ensure that the URL is valid.


fernandog:
nice adding. thanks @Amadeus- 

http://feathub.com/SiCKRAGETV/SickRage/+248

PR #169, opened 2015-11-24
The authentication validation method should return false on error.

Amadeus-:
This is a small bug; however, the method that checks to make sure that the api key is set properly should return False so that the rest of the routine doesn't try to process.

It appears that other provider files have the "return False" properly in this method -- so, it was probably just an oversight/typo here.


miigotu:
Thanks, nice find. I probably just missed it when I fixed some other crazy brokenness there not too long ago

PR #171, opened 2015-11-24
Revert "Check for login in BLUETIGERS"

miigotu:
Reverts SickRage/SickRage#85

Breaks BlueTigers according to @adaur


adaur:
:+1: 

throws SEARCHQUEUE-DAILY-SEARCH :: [BLUETIGERS] :: Unable to connect to provider

fernandog:
should  "return False" not "return results"

adaur:
Same error

labrys:
@fernandog results is set to [ ].  An empty list (just like an empty string, zero, or None, etc...) evaluates to False.

PR #173, opened 2015-11-24
Rewrite kat to use bs4 instead of error prone xmltodict

miigotu:
@OmgImAlexis


OmgImAlexis:
I'll check this after.

PR #177, opened 2015-11-24
Changes to the Poster View Home Page - Smaller Posters

duramato:
@hernandito 
Correct https://github.com/SickRage/SickRage/pull/176


hernandito:
Should I close #176? I have never done any of this.

duramato:
@hernandito sure ,since they are invalid and are made against master branch

hernandito:
@duramato I feel terrible. In order to correct the rebase. I deleted my fork, and re-added it clean. Because #176 was on the older fork, it says it cannot close it because it was deleted. I dont know what to do.

duramato:
@hernandito  https://github.com/SickRage/SickRage/pull/176 is already closed

miigotu:
Still needs evaluated before merge, it changes the default look of SR.

duramato:
@hernandito do you have any screen shots of the changes?

hernandito:
@duramato Here is a screen cap. The grey background texture is something I could disable if you guys object. But it will mean tweaking a few other things so the white on white shows up properly.

hernandito:
@duramato Here is a screencap with the grey turned off.


Please let me know if you want me to change something.

OmgImAlexis:
I think we should remove the background image just for the fact that we may not be allowed to use a random image for copyright reasons.
I want to mess around with isotope before we merge as I think we can do this in js meaning we don't have to touch any of the stylesheets which is always better.

VinceVal:
I vote yes on this (sans BG image). The posters look nicer and smaller means you can fit more.

OmgImAlexis:
This can easily be done with javascript and requires a LOT less editing of the stylesheets. I can also set the margin between the items in javascript. We would need to rename the layout options though as this would be called "small poster" and currently we have one called "small" which is actually "small posts in a row layout".
screen shot 2015-11-25 at 1 36 06 pm
screen shot 2015-11-25 at 1 35 58 pm

VinceVal:
Why don't we just replace poster with this?

hernandito:
I have been tweaking the css to get closer to the original "look" of Sickrage, without background. Take a look at how many items I can see at once when by browser is a little smaller than 1920x1080 resolution.I get all the same info from the larger poster view.

OmgImAlexis:
@hernandito I think the best bet would be to use jquery like I said that way it's easier to maintain than using CSS and really it's a layout issue not a style issue so CSS isn't really what should be used anyway.

miigotu:
Also, making them TOO small isn't really a great idea.

OmgImAlexis:
@miigotu I tried 150px wide and they looked fine.

Am I okay to rename the fields?

miigotu:
I dont see why not?

labrys:
@OmgImAlexis how easy would it be to just add a size option to the posters layout?  Then you could select small, medium, large.

OmgImAlexis:
@labrys really it should all be redone, we should be doing all of the layout using js instead of reloading the page just to change it. The use a post/get request to tell sickrage to save the change on the backend.

miigotu:
ajax!

hernandito:
Resolution I am used in my layout was 126px wide by 190px high. I did have to reduce the font sizes of the show name, air date, quality, and the tv network image resolution. 

Guys, I dont have the skills to help code this in anything other than tweaking css... I hope one of you pros can take this over.... : /

VinceVal:
Any interest in something like #205?

PR #179, opened 2015-11-24
Make Newpct quality correct and make it more scene "friendly"

duramato:
@CristianBB what do you think?

### DIFF
--- a/sickbeard/providers/newpct.py
+++ b/sickbeard/providers/newpct.py
@@ -225,21 +225,26 @@ class newpctProvider(generic.TorrentProvider):
         title = title[22:]
 
         # Quality
-        title = title.replace('[HDTV]', '[720p HDTV x264]')
-        title = title.replace('[HDTV 720p AC3 5.1]', '[720p HDTV x264]')
-        title = title.replace('[HDTV 1080p AC3 5.1]', '[1080p HDTV x264]')
-        title = title.replace('[DVDRIP]', '[DVDrip x264]')
-        title = title.replace('[DVD Rip]', '[DVDrip x264]')
-        title = title.replace('[DVDrip]', '[DVDrip x264]')
-        title = title.replace('[DVDRIP-AC3.5.1]', '[DVDrip x264]')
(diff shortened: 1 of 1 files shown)


CristianBB:
@duramato nice! =P
However I have to test if new format is procesed correctly, newpct regex uses the marker "[" so probably don't work anymore (really I'm with @miigotu , this provider sucks...). I'm out of my house so let me test it tonight, I'll post the result

CristianBB:
@duramato I've testing it and, as I suppose, SR don't parse results correctly. 
We have some options:
- Adapt newpct regex 
- Format results returned by provider to construct an string that SR can correctly parse
- Disable, burn and throw to trash this provider....

After some weeks with newpct regex, I'm not so sure if this regex is being good to SR. This regex works fine, but this type of naming is so similar to anime naming patterns that we can have problems with them.
I prefer the second option, we could delete newpct regex
What do you think?

In addition, there is a issue with:
title = title.replace('[Español Castellano]', 'SPANISH')

I've tested with "Heroes Reborn - Temporada 1 [HDTV][Cap.108][Español Castellano]" but replace instruction is not working... maybe a charset issue?

sickbeard/providers/newpct.py
```diff
@@ -225,21 +225,26 @@ def _processTitle(title):
...
+        title = title.replace('[BRrip]', '720p BlueRay x264')
+        title = title.replace('[BDrip]', '720p BlueRay x264')
+        title = title.replace('[BluRay Rip]', '720p BlueRay x264')
+        title = title.replace('[BluRay 720p]', '720p BlueRay x264')
+        title = title.replace('[BluRay 1080p]', '1080p BlueRay x264')
+        title = title.replace('[BluRay MicroHD]', '1080p BlueRay x264')
+        title = title.replace('[MicroHD 1080p]', '1080p BlueRay x264')
+        
+        title = title.replace('[Spanish]', 'SPANISH')
+        title = title.replace('[Español Castellano]', 'SPANISH')
```
miigotu, line 245: `title = title.replace(u'[Español Castellano]', 'SPANISH')`

PR #185, opened 2015-11-25
Fix bitcannon

Amadeus-:
This commit fixes bitcannon as a torrent provider.   There are also some linting in the torrent test file.

### DIFF
--- a/sickbeard/providers/bitcannon.py
+++ b/sickbeard/providers/bitcannon.py
@@ -17,7 +17,8 @@
 # You should have received a copy of the GNU General Public License
 # along with SickRage.  If not, see <http://www.gnu.org/licenses/>.
 
-from urllib import quote_plus
+import traceback
+from urllib import urlencode
 
 from sickbeard import logger
 from sickbeard import tvcache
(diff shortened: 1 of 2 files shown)


miigotu:
Except you removed the sorting and didn't stick to the provider format lol....

OmgImAlexis:
@miigotu does it need to be sorted since you can sort using the API?

miigotu:
We want all providers to follow the same layout, logs, result format, etc.

This is sort of out of the ordinary where you would normally try to optimize each individual one as best you could.

But we want this because we have so many providers to maintain (and it is growing) that having a common format not only makes it simpler to maintain and always understand what is going on regardless of which provider you are looking at, but also because we want to be sure that someone picking a random provider as a base for a new provider will start with a proper base format.

Amadeus-:
I have to admit I did not worry about the sorting since the default sorting from the provider is the same as the other ones, but I can add that.

Otherwise, I thought it was using the same format as the other providers.  I'll compare with kat.py again and update this PR.

Amadeus-:
@miigotu  -- does the new commit look better?   I tried to model after kat.py, which I was thinking was yours.   (It ends up having one more tab level out due to the api_key/auth check.)

@OmgImAlexis -- can you test this again?   I didn't change anything really, but had to shift a bunch of things around and added try/exception blocks to match the model provider I was using.  The pylint shows everything as good, but I don't really use this provider to actually test it in "real life" use.

OmgImAlexis:
```
2015-11-26 21:25:57 SEARCHQUEUE-MANUAL-257655 :: [BitCannon] :: Unknown exception in getURL /api/search?q=&category=tv&apiKey= Error: "Invalid URL u'/api/search?q=&category=tv&apiKey=': No schema supplied. Perhaps you meant http:///api/search?q=&category=tv&apiKey=?"
2015-11-26 21:25:57 SEARCHQUEUE-MANUAL-257655 :: [BitCannon] :: Traceback (most recent call last):
  File "/Users/xo/code/SickRage/sickbeard/helpers.py", line 1617, in getURL
    resp = session.get(url, timeout=timeout, allow_redirects=True, verify=session.verify)
  File "/Users/xo/code/SickRage/lib/requests/sessions.py", line 476, in get
    return self.request('GET', url, **kwargs)
  File "/Users/xo/code/SickRage/lib/requests/sessions.py", line 450, in request
    prep = self.prepare_request(req)
  File "/Users/xo/code/SickRage/lib/requests/sessions.py", line 381, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/Users/xo/code/SickRage/lib/requests/models.py", line 304, in prepare
    self.prepare_url(url, params)
  File "/Users/xo/code/SickRage/lib/requests/models.py", line 361, in prepare_url
    "Perhaps you meant http://{0}?".format(url))
MissingSchema: Invalid URL u'/api/search?q=&category=tv&apiKey=': No schema supplied. Perhaps you meant http:///api/search?q=&category=tv&apiKey=?
```

PR #190, opened 2015-11-25
[FIX] Lost of Subtitle Plugins selection

eucalyptus30:
https://github.com/SickRage/sickrage-issues/issues/76


fernandog:
@OmgImAlexis

PR #196, opened 2015-11-25
NZB splitter updates

labrys:
@miigotu If you wouldn't mind taking a look at this before merging.  There are a couple of regex's that still need cleaning up and double check my logic for stupid mistakes since I'm pushing this while tired :P

### DIFF
--- a/sickbeard/nzbSplitter.py
+++ b/sickbeard/nzbSplitter.py
@@ -1,5 +1,6 @@
+# coding=utf-8
 # Author: Nic Wolfe <nic@wolfeden.ca>
-# URL: https://sickrage.github.io
+# URL: https://SickRage.GitHub.io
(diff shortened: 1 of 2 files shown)


MGaetan89:
Can you rename the file in a separate PR (or don't rename it at all)? Because now it is not easy to review what you did from GitHub.
And @miigotu will probably complain because you have 6 commits in your PR :P

labrys:
It ain't rebased yet.  And sorry git-extensions and PyCharm see it as a rename so the rename wasn't an issue here.  I'll change that shortly.

labrys:
Pushed again without the file rename.

miigotu:
Other than that regex likely failing on specific filenames it looks good for rebase and merge.

PR #205, opened 2015-11-26
Add slider to poster layout to resize posters.

VinceVal:
This also incorporates the styling suggested by @hernandito in #177.


miigotu:
:+1:   This is so damn awesome, please finish it! lol

OmgImAlexis:
I'd suggest you change the CSS to something like this, you should notice we're using max-width now. Then to change the size of the posters just make the width change with `$('.show-container');` and then reflow isotope with `$('#container, #container-anime').isotope('reloadItems');`

```
.show-container {
    margin: 12px;
    width: 188px;
    background-color: #DFDACF;
    border: 1px solid #111;
    border-radius: 6px;
}

.show-image {
    overflow: hidden;
    max-width: 100%
    border-top-left-radius: 5px;
    border-top-right-radius: 5px;
}
```

VinceVal:
@OmgImAlexis thank you, I will do that.

gui/slick/js/core.js
```diff
@@ -1762,6 +1762,25 @@ var SICKRAGE = {
                 });
             }, 500));
 
+            $('#slider').slider({
```
OmgImAlexis: I'd change this to something like so. 125 being the min as anything smaller breaks the layout.

```
$('#slider').slider({
    min: 125,
    max: 250,
    value: 186,
    change: function (e, ui) {
        $('.show-container').css({
          width: ui.value + 2,
        });
        $('#container, #container-anime').isotope();
    }
});
```

OmgImAlexis: Also we don't need CSS on the image as it should be done using the `max-width:100%;`.

VinceVal:
@miigotu @OmgImAlexis please take a look again.

gui/slick/css/dark.css
```diff
@@ -265,13 +265,13 @@ home.mako
     color: #fff;
 }
 
+.loading-spinner {
```
OmgImAlexis, line 268: Isn't this already set using javascript?

duramato:
Do you have any SS?

gui/slick/js/core.js
```diff
@@ -1862,16 +1916,16 @@ var SICKRAGE = {
                 sortAppend: [[2,0]]
             });
 
-            // @TODO we need to check if doing a $('') with a comma would be quicker than a seperate
-            //       isotope function for each as it does here
-            $.each([$('#container'), $('#container-anime')], function (){
-                this.isotope({
+            var $gridContainers = $('.show-grid');
+            $gridContainers.imagesLoaded(function () {
```
OmgImAlexis: There's no reason to use an extra var here.

VinceVal: @OmgImAlexis I usually store the jQuery object if I end up using the same selector more than once or twice, but I can definitely change it if that style is not used/encouraged here.

gui/slick/css/dark.css
VinceVal, line 268: @OmgImAlexis I'll take a look. I did that quickly because I was tired of seeing all the posters show up in a single vertical column and then move around. :)

gui/slick/js/core.js
OmgImAlexis: Even if it does get used I'd much prefer a non PHP style var name, so drop the `$`.

VinceVal: The `$` indicates that it is a jQuery object:

http://stackoverflow.com/a/553734

I'm down to adapt to whatever convention you guys want; this is just how I've done things in the past. ;)

OmgImAlexis: "answered Feb 16 '09 at 15:42", this doesn't get used anymore :stuck_out_tongue:

miigotu:
pretty broken here atm, something with vendor.min.js (yes developer == 1 lol)

labrys:
Yeah doesn't work here either.

labrys:
Friggin awesome!

PR #207, opened 2015-11-26
Add custom_url option for kat and tpb, set them like "http://kat.cr" …

miigotu:
Use the official urls by default, can type in a mirror url in the "Custom url" field of the provider's setting if the official one is blocked for you.

Fixes https://github.com/SickRage/sickrage-issues/issues/7

This is the initial step towards having a list that autoselects the best mirror.

This does not mask the custom url from logs (intentionally) and also updates the href url you are taken to when clicking on the provider icon.

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -76,10 +76,14 @@ $('#config-components').tabs();
                                     continue
 
                                 curName = curProvider.getID()
+                                if hasattr(curProvider, 'custom_url'):
+                                    curURL = curProvider.custom_url or curProvider.url
+                                else:
+                                    curURL = curProvider.url
                             %>
                             <li class="ui-state-default ${('nzb-provider', 'torrent-provider')[bool(curProvider.providerType == "torrent")]}" id="${curName}">
                                 <input type="checkbox" id="enable_${curName}" class="provider_enabler" ${('', 'checked="checked"')[curProvider.isEnabled() is True]}/>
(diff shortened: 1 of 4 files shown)


OmgImAlexis:
It'd be a good idea to either link to https://kastatus.com/ or list the official mirrors and others we know that work with SickRage on the wiki.

miigotu:
Eventually we can either do one of two things:

Include a preset list of mirrors, maybe 5, selectable with a dropdown and have the last option as "Custom"

Include a list internally and autoselect the best one every time one fails. (no gui setting)

OmgImAlexis:
A mix of both would be good. We could try to auto select and if all fail then log it as an error and ask them to try a custom one.
If they have a custom one we'd use that though.

miigotu:
Sure, just still unsure of a reliable, clean, and stable way of implementing the auto-select just yet. When the vision comes to me I will create that masterpiece =P

OmgImAlexis:
Sounds good.

fernandog:
What about adding a test button with a fixed search string?

SickRage2:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked
- [ ] Include basic information: Branch/Commit, OS, What you did, What happened, What you expected
- [ ] Enable debug logging (be sure to disable after the bug is fixed)
- [ ] Post debug logs, either inline (for smaller logs) or using gist)

Please make sure you also read [how to create an issue](https://github.com/SickRage/sickrage-issues/blob/master/README.md) and followed all of the steps. 

The title should describe your issue. Having "SR not working" or "I get this bug" for 100 issues, isn't really helpful. We will close issues if there isn't enough information.

Sometimes the devs may seem like grunts and respond with short answers. This isn't (always) because the dev hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let us know, and this bot may get updated to automatically answer you.

Thanks!

miigotu:
Ouch, might need to run separate bots

PR #212, opened 2015-11-26
Newpct: Several improvements

CristianBB:
@duramato that's the option that I was talking about on
https://github.com/SickRage/SickRage/pull/179

I've also introduced several improvements:
- Added RSS support using original result page filtered by date
- Make results more scene "friendly" as suggest by @duramato
- Removed newpct regex, this regex is so similar to anime regex and
  sometimes it causes parse errors, with this change newpct provider will
  format the results to make them totally parseable by SR

@miigotu do you agree with this changes? If yes you can close PR  #179

### DIFF
--- a/sickbeard/name_parser/regexes.py
+++ b/sickbeard/name_parser/regexes.py
@@ -67,6 +67,18 @@ normal_regexes = [
      -(?P<release_group>[^- ]+([. _-]\[.*\])?))?)?$              # Group
      '''),
 
+    ('newpct',
+     # American Horror Story - Temporada 4 HDTV x264[Cap.408_409]SPANISH AUDIO -NEWPCT
+     # American Horror Story - Temporada 4 [HDTV][Cap.408][Espanol Castellano]
+     # American Horror Story - Temporada 4 HDTV x264[Cap.408]SPANISH AUDIO –NEWPCT)
+     r'''
+     (?P<series_name>.+?).-.+\d{1,2}[ ,\.]       # Show name: American Horror Story
+     (?P<extra_info>.+)\[Cap\.                   # Quality: HDTV x264, [HDTV], HDTV x264
+     (?P<season_num>\d{1,2})                     # Season Number: 4
+     (?P<ep_num>\d{2})                           # Episode Number: 08
+     ((_\d{1,2}(?P<extra_ep_num>\d{2}))|.*])     # Episode number2: 09
+     '''),
+
     ('fov',
      # Show_Name.1x02.Source_Quality_Etc-Group
      # Show Name - 1x02 - My Ep Name
@@ -115,17 +127,6 @@ normal_regexes = [
      (?!264)                                     # don't count x264
      (?P<season_num>\d{1,2})                     # 1
      (?P<ep_num>\d{2})$                          # 02
-     '''),
-
-    ('newpct',
-     # Example: Sobrenatural - Temporada 10 [HDTV][Cap.1023][Espanol Castellano]
-     r'''
-     (?P<series_name>.+?)                      # Showw_Name: "Sobrenatural"
-     (?:.-.+\d{1,2}.\[)                        # Separator and junk: " - Temporada 10 ["
-     (?P<extra_info>.+)                        # Quality: "HDTV"
-     (?:\]\[.+\.)                              # junk: "][Cap."
-     (?P<season_num>\d{1,2})                   # Season number: "10"
-     (?P<ep_num>\d{2})(?:])                    # Episode number: "23"
      '''),
 
     ('verbose',
--- a/sickbeard/providers/newpct.py
+++ b/sickbeard/providers/newpct.py
@@ -45,37 +45,29 @@ class newpctProvider(generic.TorrentProvider):
 
         self.urls = {
             'base_url': 'http://www.newpct.com',
-            'search': 'http://www.newpct.com/buscar-descargas/'
+            'search': 'http://www.newpct.com/index.php'
         }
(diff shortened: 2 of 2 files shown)


miigotu:
What about newpct1.com ? someone said newpct.com was no longer updated?

CristianBB:
Both pages are updated at same time, newpct1.com is a new version to avoid some restrictions by ISP (some of them blocked original newpct.com).
In any case, newpct1.com is less friendly (even!) that newpct.com

duramato:
@CristianBB and what about the actual file names?

CristianBB:
@miigotu you can look this result page: http://www.newpct.com/index.php?l=doSearch&q=&bus_de_=semana&category_=767&idioma_=1

Last torrents are added today

CristianBB:
@duramato Until now this provider had no RSS so there should be nothing on cache

Edit: you mean to process files after downloading?

duramato:
i mean when they are PP since you add the regex to the provider only

CristianBB:
You're right, that could affect even renaming results cause downloaded torrent have their own files with their own format name. Let me do some test with that

CristianBB:
Indeed it isn't working fine, post processing fails.... 
Downloaded files has the following format:
Dir name: "American Horror Story - Temporada 4 [HDTV][Cap.408][Español Castellano]"
File in dir: "AHS408 [www.newpct1.com].avi"

Parsing by file name will never worked (until now it's being parsed by dir name), here is the problem:

1 - with changes by @duramato (scene names more friendly), SR always returns NO RESULTS cause current newpct regex don't work with this changes.
2 - If I adapt newpct regex to new format, then Post Processing will fail cause Dir Name of torrent downloaded has original format

If I keep the current newpct regex (problem 2 solved) I'll still have to convert results from provider, otherwise problem 1 still remains, but I think that it's sooooo ugly (to keep same regex on different places....)

Please, some new idea? I'm pretty disgusted with this provider :@

CristianBB:
Forgot it, I think that I can adapt this regex to work with both names (original and scene friendly)

CristianBB:
Something like this should work in both cases, included multi episode torrents:

"(?P<series_name>.+?)(?:.-.+\d{1,2} )(?P<extra_info>.+)(?:[Cap.)(?P<season_num>\d{1,2})(?P<ep_num>\d{2})((?:])|((?:_)(?P<season_num2>\d{1,2})(?P<ep_num2>\d{2})))"

Possible cases:
American Horror Story - Temporada 4 [HDTV][Cap.408][Espa\xa4ol Castellano]
American Horror Story - Temporada 4 HDTV x264[Cap.408]SPANISH AUDIO -NEWPCT
American Horror Story - Temporada 4 HDTV x264[Cap.408_409]SPANISH AUDIO -NEWPCT

Please don't merge it, I need to make several test with it.... I'll tell you today

CristianBB:
@miigotu @duramato After a lot of test, that's the result:
- New regex process correctly the results previously formatted to be more friendly
- Post processing (with original file names) are working ok with every show that I've tested
- As new regex include a literal uncommon ("[Cap.")  it looks that it isn't causing more problems with anime.
- Multi episode results now works.
- Quality conversion has been modified by regular expressions, in my test I've see a lot of differents results that only vary a few letters. Now it's converting pretty fine.

Some variants that now are working fine are:
American Horror Story - Temporada 4 HDTV x264[Cap.408_409]SPANISH AUDIO -NEWPCT
American Horror Story - Temporada 4 [HDTV][Cap.408][Español Castellano]
American Horror Story - Temporada 4 HDTV x264[Cap.408]SPANISH AUDIO –NEWPCT
Masters Of Horror - Temporada 2 [HDTV][Cap.213 FINAL][Spanish] 
Modern.Family.-.Temporada.7.[720p.HDTV.x264][Cap.706][Español.Castellano]
American.Horror.Story.-.Temporada.4.[720p.HDTV.x264][Cap.408][Español.Castellano]
American.Horror.Story.-.Temporada.2.[720p.HDTV.x264][Cap.211][Español.Castellano]
The.Walking.Dead.-.Temporada.6.[720p.HDTV.x264][Cap.607][[AC3.5.1.Español.Castellano]

Please if you see any issue let me know

sickbeard/name_parser/regexes.py
```diff
@@ -67,6 +67,22 @@
      -(?P<release_group>[^- ]+([. _-]\[.*\])?))?)?$              # Group
      '''),
 
+    ('newpct',
+     # American Horror Story - Temporada 4 HDTV x264[Cap.408_409]SPANISH AUDIO -NEWPCT
+     # American Horror Story - Temporada 4 [HDTV][Cap.408][Espanol Castellano]
+     # American Horror Story - Temporada 4 HDTV x264[Cap.408]SPANISH AUDIO –NEWPCT)
+     r'''
+     (?P<series_name>.+?)                        # Show name: American Horror Story
+     (?:.-.+\d{1,2}[ ,.])                        # Junk info
```
miigotu: should explicitly catch Temporada here?

CristianBB: @miigotu Of course not, I used that to be sure what I'm capturing when I was testing this regex, sorry

CristianBB:
@miigotu I've removed all unnecesary captures from regex

sickbeard/name_parser/regexes.py
miigotu: yeah, I misinterpreted. I missed that .+, was thinking Temporada wouldn't be caught there and end up in series_name

PR #213, opened 2015-11-26
Try to fix backlogOverview not showing "low quality" snatches

fernandog:
@Amadeus- @medariox @miigotu @duramato @labrys can you test this please?

It fixed for me.
Tested with shows with options "archive first match" enabled/disabled

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1289,6 +1289,9 @@ class TVShow(object):
 
     def getOverview(self, epStatus):
 
+        anyQualities, bestQualities = Quality.splitQuality(self.quality)  # @UnusedVariable
+        epStatus, curQuality = Quality.splitCompositeStatus(epStatus)
+        
         if epStatus == WANTED:
             return Overview.WANTED
         elif epStatus in (UNAIRED, UNKNOWN):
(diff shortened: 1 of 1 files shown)


miigotu:
This breaks the ep counts on home and displayshow. I told you this isnt an easy fix -.-
screenshot from 2015-11-26 06 16 47

fernandog:
@miigotu 

so in your opinion then the counter should display:
wanted: 0
low quality: 6 
download: 3
snatched: 3 ?

is that it ?

fernandog:
this is wrong isn't?

            <label for="good"><span class="good"><input type="checkbox" id="good" checked="checked" /> Downloaded: <b>0</b></span></label>

if should be "downloaded" not "good"

fernandog:
IMO "Downloaded" should be "Higher Quality" or "Best Quality" - to be aligned with "Low quality"
then the counter will be correct. The label is the ep line is already "Downloaded"

today we are mixing quality with status

labrys:
I think something needs to be done about this, but I haven't looked at your solution yet, and would need to think about it for a while.... Working on something else atm, hope to get back to this soon though.

labrys:
Now that I've got a minute, my quick two cents worth:  I think you are right Fernandog that we really need to rework how this is done, but I think your proposed change here isn't thorough enough and may have implications in other parts of the code.  This is definitely part of the user interface that should be improved, but it might be worth it to hold off on this sort of change until we've cleaned up the code some and added some test cases to deal with this sort of scenario before we go live with it.  The current functionality isn't optimal, but neither is it broken.

fernandog:
been playing around with this:

**Screenshot 1**
image


**Screenshot 2**
image

fernandog:
Check my proposal with screenshots:
https://github.com/fernandog/SickRage/commit/1c71c83959ca7df6833d263c482a81f0ac32f2ac#commitcomment-14644215

fernandog:
Or for now we apply this PR and rename the tag "Downloaded:" to "Best quality"
in the code the "downloaded" is Overview.GOOD, so this is a quality tag.

and later we improve the new layout with Quality vs Status

pkoutsias:
IMHO https://github.com/fernandog/SickRage/commit/1c71c83959ca7df6833d263c482a81f0ac32f2ac#commitcomment-14644215 is the way to go, although it really needs to be tested a lot. I would also suggest to replace the "best quality" with "wanted quality" or "target quality" or something similar.

miigotu:
just....leave...it....alone.....
we discussed this and concluded that it needs to be parked for awhile. overview does more than just visual gui.

PR #215, opened 2015-11-26
Move more functions to sickrage.helper.common and add more tests

MGaetan89:
Move `notTorNZBFile` (as `is_torrent_or_nzb_file`) and `pretty_filesize` (as `pretty_file_size`) to `sickrage.helper.common`.

The behavior of `pretty_file_size` changed a bit:
- If the provided size if too big, it returns the provided size (instead of `None` in the current implementation)
- The type of the parameter is checked. So a string containing a float is not parsed and returns an empty string. If the string contains an int, it works fine


MGaetan89:
Rebased.

PR #218, opened 2015-11-26
Moving isSyncFile and sanitizeFilename to sickrage.helper.common

MGaetan89:
There are still 5/10 functions I want to extract from `sickbeard/helpers.py` and test, but I will do that in an other PR.

### DIFF
--- a/sickrage/helper/common.py
+++ b/sickrage/helper/common.py
@@ -16,6 +16,9 @@
 # You should have received a copy of the GNU General Public License
 # along with SickRage.  If not, see <http://www.gnu.org/licenses/>.
 
+import re
+import sickbeard
+
 dateFormat = '%Y-%m-%d'
 dateTimeFormat = '%Y-%m-%d %H:%M:%S'
 media_extensions = [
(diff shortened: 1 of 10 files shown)


sickrage/helper/common.py
```diff
@@ -26,17 +29,32 @@
...
     :param filename: The filename to check
     :return: ``True`` if the ``filename`` is a NZB file or a torrent file, ``False`` otherwise
     """
 
-    if filename is None:
+    if not isinstance(filename, (str, unicode)):
         return False
 
-    return filename.endswith('.nzb') or filename.endswith('.torrent')
+    return filename.rpartition('.')[2].lower() in ['nzb', 'torrent']
```
miigotu, line 57: Doesnt make a huge difference, but endswith takes a tuple optionally too:
`filename.lower().endswith(('.nzb', '.torrent'))`

MGaetan89, line 57: I didn't know that.
I just chances it so if looks like the other functions in the file.

PR #219, opened 2015-11-26
Fix api builder error

labrys:
Missed some allowedValue refactors.


labrys:
closes sickrage/sickrage-issues#87

PR #224, opened 2015-11-27
Split Overview.SNATCHED into seperate SNATCHED_BEST and SNATCHED_PROP…

miigotu:
…ER, add selector and SNATCHED to backlogOverview

Still need to handle when show doesnt have any preferred qualities (would only be set as SNATCHED not SNATCHED_BEST I think)
Possible to also assign different css to these to color them different, but need to be careful not to break the selector id's on displayShow
Replaces #213

### DIFF
--- a/gui/slick/views/manage_backlogOverview.mako
+++ b/gui/slick/views/manage_backlogOverview.mako
@@ -20,12 +20,15 @@
 % endif
 
 <%
+    hasPreferredQual = lambda x: Quality.splitQuality(x.quality)[1]
+
     totalWanted = totalQual = totalQualSnatched = 0
     backLogShows = sorted([x for x in sickbeard.showList if showCounts[x.indexerid][Overview.QUAL] + showCounts[x.indexerid][Overview.WANTED] + showCounts[x.indexerid][Overview.SNATCHED]], key=lambda x: x.name)
     for curShow in backLogShows:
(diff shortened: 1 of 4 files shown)


miigotu:
@fernandog this is how it needs to be done, however there are still some small things to do for this (or at least check)

I might move .QUAL to another value so that SNATCHED can carry it's real common.SNATCHED value as well.

EDIT: Did those two things, test it out please. Could still change the Overviewstrings for proper and best and add some css to distinguish them

fernandog:
@miigotu 

Seems ok The low snatched appeared OK.
But on show is showing as "Snatched (Low Quality)" but it has archived first match enabled, so its not low quality.

About the color I would keep the yellow too and keep the pink for "last" snatch. what do you think?

miigotu:
Yellow doesnt belong for anything snatched regardless of quality. yellow means you have a file already to watch.

fernandog:
Ok. Do you need more info about the other one? 
Allowed 1080 web
Preferred 1080p hdtv
Archive first match enabled
Ep status: snatched 1080 web

miigotu:
@fernandog and now?

fernandog:
Its ok now! Maybe others should test too.
Are you going to change displayShow too or just backlogOverview?

miigotu:
displayshow just shows them as they are, no real reason to split them without css for special colors, but that would also add 2 more selectors at the top of the page.

PR #229, opened 2015-11-27
Better subtitles!

medariox:
- subtitles.py uses the native subliminal function to detect subtitles now!
- subtitles won't be refreshed if they have not changed
- linted everything I could lint without impacting the rest of the code (hopefully)
- removed outdated/unnecessary functions

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -18,7 +18,6 @@
 # You should have received a copy of the GNU General Public License
 # along with SickRage.  If not, see <http://www.gnu.org/licenses/>.
 
-import io
 import os
 import re
 import datetime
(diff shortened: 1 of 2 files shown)


miigotu:
Still think there is more code that can be left in subliminal, but this is better =P

medariox:
For example?

miigotu:
save_subtitles and a few other methods, too tired to really dig into it atm, but subliminal handles all that, and you even pass the desired directory directly to it in the new subliminal

medariox:
save_subtitles is the only function that could eventually be replaced by anything in subliminal afaik. We replicate it in subtitles.py to sickbeard.helpers.chmodAsParent() and to sickbeard.helpers.fixSetGroupID(). Otherwise we would need an addition loop for that. It was like that in the past and has been changed for that reason. I can do either ways. Maybe we should keep using the subliminal one.

miigotu:
I think is a lot less code to use the subliminal one and then listdir on the subtitles_path and set mode and group that way.

Plus, the more code you can leave in subliminal, the easier it is to adapt your code on the next subliminal update.

medariox:
You are right. Will use the subliminal one.
Edit: done.

sickbeard/subtitles.py
```diff
@@ -238,110 +259,37 @@ def getSubtitlesPath(video_path):
...
-    for language in subtitle_languages:
+    result_list = []
+
+    if not video.subtitle_languages:
+        return result_list
+
+    if not sickbeard.SUBTITLES_MULTI:
+        if len(wantedLanguages()) == 1 and Language('und') in video.subtitle_languages:
+            video.subtitle_languages.remove(Language('und'))
+            video.subtitle_languages.add(fromietf(wantedLanguages()))
```
miigotu: Shouldnt this be:
`fromietf(wantedLanguages()[0])`

PR #90, opened 2015-11-18
add more explicit dependencies so it builds better from clean node install

alisonatwork:
- use devDependencies instead of dependencies since bower is handling the actual dependencies
- add grunt-cli for if you don't have it installed globally
- be more explicit on our bootstrap version so bower doesn't prompt which one to choose

PR #233, opened 2015-11-27
Fixes issue #90

labrys:
This fixes sickrage/sickrage-issues#90.  Api was only showing show results if paused=1.

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -2798,8 +2798,8 @@ class CMD_Shows(ApiCall):
         shows = {}
         for curShow in sickbeard.showList:
 
-            if not self.paused and not curShow.paused:
-                continue
+            if not self.paused and curShow.paused:  # If we're not including paused shows, and the current show is paused
+                continue  # continue with the next show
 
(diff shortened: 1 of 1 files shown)


tresni:
This is still broken, now it will never return paused shows :)  Additionally, it's an optional parameter and according to Sickbeard: `if not set then both paused and non paused are shown`

50ee052c599160 has the right code... It was "fixed" in a4374681c87477ece402e889 and it's been broken ever since...

This should reverted to the old code to get the right behavior.

tresni:
@labrys not sure you saw the previous comment, but want to make sure you do.  There is currently no way to get paused shows at all.  You will always get _only_ unpaused shows.

tresni:
So to be explicit (and based on the documentation of the [Sickbeard API](http://sickbeard.com/api/):
- `paused=1` should show _only_ paused shows
- `paused=0` should show _only_ unpaused shows
- If `paused` is not defined/present, then it should show _both_ paused and unpaused shows.

labrys:
No I must have missed the previous comment.  I'll take a look at this right now.

PR #239, opened 2015-11-27
Lint 9.95/10

labrys:


labrys:
Branch was incorrectly named - should have been lint-helpers_tests

PR #240, opened 2015-11-27
Add tests to helpers_tests

labrys:
Added missing tests to local **main** loop.

### DIFF
--- a/tests/helpers_tests.py
+++ b/tests/helpers_tests.py
@@ -707,3 +707,27 @@ if __name__ == '__main__':
 
     SUITE = unittest.TestLoader().loadTestsFromTestCase(HelpersTests)
     unittest.TextTestRunner(verbosity=2).run(SUITE)
+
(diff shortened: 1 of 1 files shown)


tests/helpers_tests.py
```diff
@@ -707,3 +707,27 @@ def test_pretty_time_delta(self):
 
     SUITE = unittest.TestLoader().loadTestsFromTestCase(HelpersTests)
     unittest.TextTestRunner(verbosity=2).run(SUITE)
+
```
miigotu, line 710: Maybe you could save some typing by using a for loop here...

```
cases = [HelpersTests, HelpersConnectionTests, HelpersDirectoryTests, ...]
for case in cases:
    SUITE = unittest.TestLoader().loadTestsFromTestCase(case)
    unittest.TextTestRunner(verbosity=2).run(SUITE)
```

labrys:
yeah planned on it.. but a copy paste was faster at the time and had to get to the store :)

PR #242, opened 2015-11-28
Don't log IP when access is granted in API

fernandog:
@miigotu @MGaetan89


miigotu:
You should leave this. You should be able to see when someone logs in even with a correct api key so you can know if it wasnt you....

PR #246, opened 2015-11-28
Bugfix 73

miigotu:
This fixes credits, writers, and genre in KODI 12+ nfo files, and adds country tag.
I still have a few assertion errors from old repo here, will continue in another PR.

### DIFF
--- a/sickbeard/metadata/kodi_12plus.py
+++ b/sickbeard/metadata/kodi_12plus.py
@@ -19,6 +19,7 @@
 
 import re
 import datetime
+from babelfish import Country
 
 import sickbeard
 from sickbeard.metadata import generic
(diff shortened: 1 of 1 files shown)


MilhouseVH:
Will this now create:

```
<credits>WriterA / WriterB</credits>
```

or

```
<credits>WriterA</credits>
<credits>WriterB</credits>
```

in the NFO?

I know the former works, but haven't tested the latter...

miigotu:
@MilhouseVH It will create the second style. Both work as far as I can tell, and actually Kodi itself uses the second style when exporting as separate files.

I also referred to examples on the kodi.wiki (seems outdated), and nfo's generated with ember, mediaelch, and tiny media manager.

If they dont work as expected, let me know.

PR #248, opened 2015-11-28
Fix notify_list variable in core.js

Amadeus-:
Basically, the "notify_list" variable that was assigned a value in loadShowNotifyLists() was never declared as a global variable, but was being used (by other methods in the file) as such.    As you might imagine, this creates great awkwardness and makes the linter unhappy.

Since core.js has very few global variables (something JS folks seem to hate anyway), I decided to alter the routines in the file to avoid it.   

One could argue that using a global variable might be more efficient; however, given the very small amount of data being returned by the AJAX call, and given that this section of the code/site is very seldom used, I'm thinking that you all would prefer to use this solution.


Amadeus-:
Note:  The "jshint ignore:line" directive on line 885 is due to the variable prowl_notify_list.   That variable is used in python and in JS ....so, one must choose which linter one is going to appease when it comes to variable naming.....

Also @OmgImAlexis :   grunt updated vendor.min.js   ....I assume that this is expected (I was just surprised since it seemed to change more than core.min.js  (and, I wasn't really expecting my changes to core.js to affect any other .min.js files.)   Anyway, let me know if I did something wrong with that.

OmgImAlexis:
All seems good, I've been going over it since the PR was first sent. :stuck_out_tongue: 

I'm still not sure what effected the vender file to update. I might look into it. Could you do a `rm -rf bower_components && rm -rf node_modules` and then `npm install && bower install && grunt` and see if that changes anything. Since grunt builds from the files in the bower_components directory it may be that you had old files and that's what got added to the vender file.

@miigotu this may have happened to you too.

gui/slick/js/core.js
```diff
@@ -868,14 +868,23 @@ var SICKRAGE = {
...
+                    var notifyData = $.parseJSON(data);
+                    if (notifyData._size > 0) {
+                        $('#email_show_list').val(key >= 0 ? notifyData[key.toString()].list : '');
+                    }
+                });
             });
             $('#prowl_show').on('change', function() {
                 var key = parseInt($('#prowl_show').val(), 10);
-                $('#prowl_show_list').val(key >= 0 ? notify_data[key.toString()].prowl_notify_list  : ''); // jshint ignore:line
+                $.get(srRoot + "/home/loadShowNotifyLists", function(data) {
```
miigotu: wouldnt this be more efficient as a `$.getJSON` ?

Amadeus-: Let me test and see :)    (I was just reusing the same logic as was already present in the file, but I think you may be right.)

Amadeus-:
OK, I removed vendor.min.js from the PR.    I'll work on miigotu's suggestion and will comment here (and ping in IRC) when PR is ready for consideration again.

Amadeus-:
@miigotu -- I implemented your suggestion and it's working as intended.

miigotu:
@OmgImAlexis he removed bower_components

Amadeus-:
@miigotu I'm sorry for the confusion -- I meant that I implemented your suggestion in the code (i.e., changing from .get to .getJSON).    I tried the suggestion that @OmgImAlexis gave, but when I ran it after deleting those directories, it still altered vendor.min.js

For now, I just didn't stage the vendor.min.js, so it's not in this PR.   I'm sure the issue will resolve itself soon.

gui/slick/js/core.js
OmgImAlexis: I wish I had caught this earlier, it's actually better to use $.get since we have more control over it and if we add ajax setup stuff later $.getJSON isn't effected from what I can remember.

PR #257, opened 2015-11-28
Subtitles

miigotu:


miigotu:
@medariox please look over the 2 commits I added, it cleans up a lot and makes it easier to follow imho.

Also, I think we can remove the hack in subtitles.py L268-271 if we check before downloading now.

```
# after scanning the video
    if not sickbeard.SUBTITLES_MULTI:
        if len(wanted_languages()) == 1 and Language('und') in video.subtitle_languages:
            # dont try and download
```

This handles all cases without making an assumption

PR #269, opened 2015-11-29
Fix poster shows '?' when only unaired episodes.

labrys:
Applies same fix as #252 to posters.

### DIFF
--- a/gui/slick/views/home.mako
+++ b/gui/slick/views/home.mako
@@ -33,7 +33,7 @@
     % if sickbeard.HOME_LAYOUT == 'poster':
     &nbsp;
     <span> Poster Size:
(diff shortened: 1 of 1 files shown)


gui/slick/views/home.mako
```diff
@@ -33,7 +33,7 @@
     % if sickbeard.HOME_LAYOUT == 'poster':
     &nbsp;
     <span> Poster Size:
-        <div style="width: 100px; display: inline-block; margin-left: 7px;" id="posterSizeSlider"></div>
```
VinceVal, line 36: Did you mean to delete this?

labrys:
No and I'm not sure how it got deleted... hmm good catch

OmgImAlexis:
All that was removed was the margin, check the diff.

labrys:
Yeah added it back in #270

PR #274, opened 2015-11-29
When posters are small, display a hover popout

VinceVal:
Thoughts on this?


labrys:
Working good for the most part.  Couple issues, pop-outs near edges get clipped.  Also a slightly longer delay before popping out my be better.

hernandito:
@VinceVal  Thank you for this. It looks terrific. I am not sure if you saw my new feature request... It would be great if the user could (in Settings) specify either a color or upload an image to be used as the background for the main body of the site.

I prefer a medium grey, to make the show "postcards" pop more in the home page. I realize this would also have to include the footers, and any other backdrop on the site.

Thanks again!

PR #275, opened 2015-11-29
ek every os.path.* and os.* that touches the file system

miigotu:
@SickRage/collaborators @SickRage/moderators 

This needs tested as much as possible **with** some actual files existing, PP, adding/removing shows, creating metadata, downloading subs, etc

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -70,6 +70,8 @@ from sickbeard.webserveInit import SRWebServer
 from sickbeard.event_queue import Events
 from configobj import ConfigObj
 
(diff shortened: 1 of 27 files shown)


neoatomic:
Looks ok for me now. But let us test for at-least a week. And before merging to master, confirming it has been tested on multiple platforms.

miigotu:
I manually tested almost every method I could find and forced it to operate on non-ascii paths, so hoping its good.

labrys:
It's unicode of course its not :stuck_out_tongue_closed_eyes:

PR #289, opened 2015-11-29
Fix pretty_file_size...

labrys:
...and lint common_tests 10/10

Fixes SickRage/sickrage-issues#104

### DIFF
--- a/sickrage/helper/common.py
+++ b/sickrage/helper/common.py
@@ -63,11 +63,9 @@ def pretty_file_size(size):
     :param size: The size to convert
     :return: The converted size
     """
-
-    if isinstance(size, str) and size.isdigit():
+    if isinstance(size, (str, unicode)) and size.isdigit():
         size = float(size)
-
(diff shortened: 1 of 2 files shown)


MGaetan89:
What's the point if splitting the tests like that?

Arcanemagus:
Maybe so it's easier to pinpoint whether it's failing due to a bug in the logic, or due to unicode handling?

labrys:
@MGaetan89 I split the dict because the unicode strings weren't being tested.  The dict had the same key as both a unicode and non-unicode string, which it sees as the exact same key.  While not all of the unicode strings may be duplicates in all of the places I split the tests, I did so to ensure that if additional tests were added later we didn't run into the same issue.

labrys:
If you run the test prior to the `pretty_file_size` fix you will see that the test fails, and if you run the test prior to me splitting them, you will see the test does not fail, because of the duplicate key issue (which is related to how Python 2 does some behind the scenes conversion of strings).

PR #308, opened 2015-11-30
Add option to notify on new login to webserver

fernandog:
Will keep this FR open and rebasing until we can merge new FR like this

This does not work for API because each api command would send a new nofity.

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -78,13 +78,17 @@ NOTIFY_DOWNLOAD = 2
 NOTIFY_SUBTITLE_DOWNLOAD = 3
 NOTIFY_GIT_UPDATE = 4
 NOTIFY_GIT_UPDATE_TEXT = 5
+NOTIFY_LOGIN = 6
+NOTIFY_LOGIN_TEXT = 7
 
 notifyStrings = NumDict({
     NOTIFY_SNATCH: "Started Download",
     NOTIFY_DOWNLOAD: "Download Finished",
     NOTIFY_SUBTITLE_DOWNLOAD: "Subtitle Download Finished",
     NOTIFY_GIT_UPDATE: "SickRage Updated",
-    NOTIFY_GIT_UPDATE_TEXT: "SickRage Updated To Commit#: "
+    NOTIFY_GIT_UPDATE_TEXT: "SickRage Updated To Commit#: ",
+    NOTIFY_LOGIN : "SickRage new login",
+    NOTIFY_LOGIN_TEXT : "New login from IP: {0}. http://geomaplookup.net/?ip={0}"
 })
 
 # Episode statuses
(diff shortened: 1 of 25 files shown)


sickbeard/common.py
```diff
@@ -72,6 +74,8 @@
 notifyStrings[NOTIFY_SUBTITLE_DOWNLOAD] = "Subtitle Download Finished"
 notifyStrings[NOTIFY_GIT_UPDATE] = "SickRage Updated"
 notifyStrings[NOTIFY_GIT_UPDATE_TEXT] = "SickRage Updated To Commit#: "
+notifyStrings[NOTIFY_LOGIN] = "SickRage new login"
+notifyStrings[NOTIFY_LOGIN_TEXT] = "New login from IP: {0}. http://geomaplookup.net/?ip={0}"
```
OmgImAlexis: Shouldn't we be using some sort of anonURL here? This basically sends your IP to a service that we don't control. Don't really like the idea of that.

fernandog: sure. will change it

fernandog:
maybe add this to helpers to no notify private IPs:

```
def is_ip_private(ip):
    priv_lo = re.compile("^127\.\d{1,3}\.\d{1,3}\.\d{1,3}$")
    priv_24 = re.compile("^10\.\d{1,3}\.\d{1,3}\.\d{1,3}$")
    priv_20 = re.compile("^192\.168\.\d{1,3}.\d{1,3}$")
    priv_16 = re.compile("^172.(1[6-9]|2[0-9]|3[0-1]).[0-9]{1,3}.[0-9]{1,3}$")

return priv_lo.match(ip) or priv_24.match(ip) or priv_20.match(ip) or priv_16.match(ip)
```

duramato:
and local host ?

fernandog:
In my tests accessing from local host I only got IP in the notification

OmgImAlexis:
@duramato localhost should use 127.0.0.1

miigotu:
@fernandog rebase and fix conflicts please? I think this should go in now. We have a bunch of major changes going in whats one small one =P

fernandog:
@miigotu any suggestions ? I couldn't make it work with API

fernandog:
@SickRage/collaborators last time some f**\* tried to access my SR. Look your logs.
I only got this error because when someone tries to access SR without 'https' i receive this warning

already did `sudo iptables -A INPUT -s 61.0.0.0/8 -j DROP`

image

miigotu:
phone have tab open?

fernandog:
no. that IP is from china. Its not the first time. in a few weeks ago I already blocked another IP. Now I blocked range

miigotu:
which is why mine isnt accessible from the internet =P

PR #317, opened 2015-11-30
Fix skipping test and PEP 8

labrys:
Fix skipping test
PEP 8: Optimize imports

### DIFF
--- a/tests/exceptions_helper_tests.py
+++ b/tests/exceptions_helper_tests.py
@@ -4,16 +4,16 @@
 Test exceptions helpers
 """
 
-import sys
-import os.path
-import unittest
-
 # pylint: disable=line-too-long
+
+import os.path
+import sys
+import unittest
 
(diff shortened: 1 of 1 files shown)


tests/exceptions_helper_tests.py
```diff
@@ -4,16 +4,16 @@
...
+import sys
 import unittest
 
-# pylint: disable=line-too-long
-
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '../lib')))
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 
-from sickrage.helper.exceptions import ex
+from sickbeard import ex
```
miigotu, line 16: ?

labrys, line 16: Was trying to keep it consistent, some files were importing ex from sickbeard, some from sickrage.helper.exceptions, some were importing ek from one and ex from the other.

PR #321, opened 2015-11-30
Fix SickRage URL in comment and PEP rest of tests

labrys:
PEP 8: Optimize imports

### DIFF
--- a/tests/encoding_tests.py
+++ b/tests/encoding_tests.py
@@ -6,18 +6,17 @@ Test encoding
 
 # pylint: disable=line-too-long
 
-import sys
-import os.path
 import locale
+import os.path
+import sys
 import unittest
 
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '../lib')))
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 
 import sickbeard
+from sickbeard import ek, ex
 from sickrage.helper.common import sanitize_filename
-from sickrage.helper.encoding import ek
-from sickrage.helper.exceptions import ex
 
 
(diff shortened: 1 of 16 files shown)


tests/encoding_tests.py
```diff
@@ -6,18 +6,17 @@
...
 import unittest
 
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '../lib')))
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 
 import sickbeard
+from sickbeard import ek, ex
 from sickrage.helper.common import sanitize_filename
-from sickrage.helper.encoding import ek
-from sickrage.helper.exceptions import ex
```
miigotu, line 20: this isnt pep8 =P, 1 import per line

tests/xem_tests.py
```diff
@@ -17,23 +17,23 @@
...
+import unittest
 
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '../lib')))
 sys.path.insert(1, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
 
-import tests.test_lib as test
-import sickbeard
 from sickbeard.tv import TVShow
+import sickbeard
+import tests.test_lib as test
```
miigotu, line 36: If you import test_lib first, you dont need to sys.path.insert here

labrys, line 36: I had removed it locally and it broke running the tests individually, although all_tests still worked.

tests/encoding_tests.py
labrys, line 20: Actually it is when doing selective imports from a single module.
https://www.python.org/dev/peps/pep-0008/#imports
=P

PR #322, opened 2015-11-30
Fixes SickRage/sickrage-issues#113

labrys:
Fixes SickRage/sickrage-issues#113


fernandog:
@labrys
Why not logger.ERROR so user can submit?

labrys:
Because a None type object isn't really an error, it's invalid input.  Basically for the same reason that we downgraded SickRage/sickrage-issues#53 SickRage/sickrage-issues#54 SickRage/sickrage-issues#55 to a debug statement.

fernandog:
Ah ok!

labrys:
Ideally adba would gracefully handle a None type object and return a None, but that's not what happens.  We could plausibly raise an issue-request to their repository, but it may be their intended design.

fernandog:
can't remember that last time we updated adba.... lol

PR #330, opened 2015-12-01
Add total show size column to homepage

miigotu:
@Thraxis thanks, dunno how I overlooked using SUM
@OmgImAlexis

### DIFF
--- a/gui/slick/views/home.mako
+++ b/gui/slick/views/home.mako
@@ -243,7 +243,7 @@
             <th>Network</th>
             <th>Quality</th>
             <th>Downloads</th>
-            ## <th>Size</th>
+            <th>Size</th>
             <th>Active</th>
             <th>Status</th>
         </tr>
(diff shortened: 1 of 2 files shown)


labrys:
@miigotu gives you sum-thing to think about :wink:

OmgImAlexis:
@VinceVal The js will also need fixing since this changes the column count again.

fernandog:
@miigotu when SR calculates the size? every page load or when displayShow is open or when SR starts?

labrys:
@fernandog I would guess if its added to the mako then every page load, but I haven't looked at the commit yet.

Thraxis:
@fernandog the sizes are grabbed from the database at the same time it grabs the air-date for display on the index.  All of the episodes have the file sizes stored in the database and an SQL query SUM is very fast.  

This is an on the fly query that's done each time the index page is displayed

fernandog:
So it get updated when: refresh show or show update threads. Nice. Thanks

Thraxis:
http://feathub.com/SickRage/SickRage/+301

fernandog:
@sandervdhoek

PR #333, opened 2015-12-01
add filter to manage

OmgImAlexis:


fernandog:
http://feathub.com/SickRage/SickRage/+291

@syrys

syrys:
@OmgImAlexis  @fernandog not sure if the featurehub comments alert you, but there is a slight bug/issue on the implementation, check the last comment on featurehub. but thanks so much for the feature.

labrys:
@syrys They are Y or N.  We're looking at implementing tooltips to make this more clear.

syrys:
@labrys ahhh, awesome. i tried yes/no, true/false, but didnt think to try y/n.

I think the dropdown solves the problem much better than having tool tips doesnt it? Probably a lot simpler to implement too :P

syrys:
also might be related to this update, but the bottom right submit button is no longer working, the top one works fine. Not a big deal, but confused me a bit just earlier till i realized that there is now a submit button at the top.

OmgImAlexis:
The bottom one "should" be working. :disappointed: Open an issue for this please.

OmgImAlexis:
Also a dropdown would just clutter the page and really can't be added because of the way the table is generated.

PR #334, opened 2015-12-01
Allow to download subtitles again when search is forced

fernandog:
http://feathub.com/SickRage/SickRage/+81

@miigotu @medariox is that it ?

Can I remove `# pylint: disable=unused-argument` ?


fernandog:
I don't think we should pass force=true here otherwise SR will redownload ALL subs from ALL eps:
https://github.com/SickRage/SickRage/blob/develop/sickbeard/webserve.py#L3547 - also not pass force=true in subtitle force search  in "manage searches"

We should only pass force=true for single episodes.

miigotu:
This doesnt do anything except fake the notification/logging that new subs were downloaded.

PR #337, opened 2015-12-01
Add tests

MGaetan89:
This PR moves `sickbeard.client.http_error_code`, `sickbeard.helpers.tryInt` and `sickbeard.helpers. codeDescription` to the `sickrage` module.

`sickbeard.client.http_error_code` had a TODO about duplicate key handling. I fixed that by setting a list to the HTTP status code that can have multiple values.

The log line in `http_code_description` (previously named `codeDescription`) is commented because there is a cyclic dependency. This can be restored once the logger has been rewritten.


sickrage/helper/common.py
```diff
@@ -21,6 +21,84 @@
...
+    409: 'Conflict',
+    410: 'Gone',
+    411: 'Length Required',
+    412: 'Precondition Failed',
+    413: 'Request Entity Too Large',
+    414: 'Request-URI Too Long',
+    415: 'Unsupported Media Type',
+    416: 'Requested Range Not Satisfiable',
+    417: 'Expectation Failed',
+    418: 'Im a teapot',
```
duramato, line 53: Do webservers even return this error code? lol

MGaetan89, line 53: I don't think so :p

Arcanemagus, line 53: There are implementations: https://en.wikipedia.org/wiki/Hyper_Text_Coffee_Pot_Control_Protocol :stuck_out_tongue:

miigotu:
@MGaetan89 need some tests for ex, some ex calls are returning None (I think when it is just a string with no args?) such as when github api limit is hit or username/password are wrong, and some errors in postProcessor (actually processTV I think)

If I change the ex call to a str call they show just fine.

MGaetan89:
OK, I will see if I can write proper tests for that in my next PR.

sickrage/helper/common.py
MGaetan89, line 53: O.o

labrys:
@MGaetan89 When you add to the test, you might also want to add unicode test cases, such as unicode in the response, etc.

sickrage/helper/common.py
duramato, line 53: lol @ http://error418.net/

Arcanemagus, line 53: Wow, that page actually responds with the number, but not the text:
image

MGaetan89:
@labrys I always add Unicode test cases

labrys:
I've noticed, I just didn't see any in the http responses, but I was tired, might have missed them. I appreciate that someone else is giving our testing some TLC. :+1:

labrys:
@MGaetan89 I'd like to export test cases out of the tests and put them in a file (likely json) so they can be easily updated and added to as well as reused between tests.

MGaetan89:
I thought about that, but as it's my first tests experience in Python, I just followed the structure that was already in place.
If you want/know to change that, I support you :)

labrys:
I've been emulating whats there myself, but its been bugging (pun intended) me :)

PR #339, opened 2015-12-01
Add buttons to top of Mass Update page

labrys:
This adds a line break to some of the column headers and adds buttons to the top of the page for `Edit Selected` and `Submit`.  This allows the user to perform the mass update without having to scroll to the bottom to initiate it, particularly when selecting an entire column.

In addition, it adds buttons to `Select Columns` and `Clear filter(s)`.  @OmgImAlexis if you wouldn't mind doing your magic on the last two so that they work properly, it would be appreciated.

### DIFF
--- a/gui/slick/views/manage.mako
+++ b/gui/slick/views/manage.mako
@@ -9,11 +9,32 @@
 </%block>
 <%block name="content">
 <%namespace file="/inc_defs.mako" import="renderQualityPill"/>
-% if not header is UNDEFINED:
-    <h1 class="header">${header}</h1>
-% else:
-    <h1 class="title">${title}</h1>
-% endif
+
+
+<table style="width: 100%;" class="home-header">
+    <tr>
+        <td nowrap>
+            % if not header is UNDEFINED:
+                <h1 class="header" style="margin: 0;">${header}</h1>
+            % else:
+                <h1 class="title" style="margin: 0;">${title}</h1>
+            % endif
+        </td>
+        <td align="right">
+            <div>
+                <input class="btn btn-inline" type="button" value="Edit Selected" id="submitMassEdit" />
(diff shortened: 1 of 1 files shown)


miigotu:
@labrys would help @OmgImAlexis to help you if this was a branch upstream =P

labrys:
@miigotu why yes, yes it would :no_mouth:

fernandog:
@OmgImAlexis is it possible to not wrap show name in mass edit? thanks!

"archive first match" is the larger one and it's not warping

labrys:
@fernandog `Archive first match` should wrap i inserted a `<br>` after `Archive`.  Make sure you ctrl-f5.

gui/slick/views/manage.mako
```diff
@@ -25,8 +46,8 @@
             <th class="col-legend">Sports</th>
             <th class="col-legend">Scene</th>
             <th class="col-legend">Anime</th>
-            <th class="col-legend">Season folders</th>
-            <th class="col-legend">Archive first match</th>
+            <th class="col-legend">Season<br>folders</th>
+            <th class="col-legend">Archive<br>first match</th>
```
OmgImAlexis: These need to be reverted, see https://github.com/SickRage/SickRage/pull/342

fernandog:
@labrys the "Select columns" still need to add the JS? here it's not working. Cleared cache.
we should hide columns with caution because of the "action" columns. Otherwise user could select "force update" but in fact the code is "delete"

miigotu:
The js isnt added yet, but omg(or anyone else) couldnt work on it easily, so it is merged and can be completed today.

fernandog:
i think we can remove the action from the columns and add them as buttons
so the actions would only be executed in the selected shows

miigotu:
https://github.com/SickRage/SickRage/pull/383

PR #340, opened 2015-12-01
PEP 8: Convert from 'not x in y' to 'x not in y'

CristianBB:
This provider was created after @labrys fixes all providers with PEP 8, then he couldn't review it


miigotu:
Trivial in this case =P

PR #346, opened 2015-12-02
Fix SickRage/sickrage-issues/issues/123

fernandog:
This new message will appear like this:

"Could not find season 4 (unaired/deleted) in the indexer theTVDB. Removing existing records from database"


fernandog:
https://github.com/SickRage/sickrage-issues/issues/123

sickbeard/tv.py
```diff
@@ -524,11 +525,11 @@ def loadEpisodesFromDB(self):
                 curEp.loadFromIndexer(tvapi=t, cachedSeason=cachedSeasons[curSeason])
                 scannedEps[curSeason][curEpisode] = True
             except EpisodeDeletedException:
-                logger.log(u"Tried loading an episode from the DB that should have been deleted, skipping it",
+                logger.log(u"%s: Tried loading %s S%02dE%02d episode from the DB that should have been deleted, skipping it" % (curShowid, curShowName, curSeason or 0, curEpisode or 0),
```
Amadeus-: I think you can remove the word 'episode'  so that it would read something like "...Tried loading SHOW S05E06 from the DB..."

This also applies to line 513 above.

PR #348, opened 2015-12-02
e

Fen0l:


fernandog:
@JohnDooe use AlphaRatio as base for example. see log messages and the needed commands in doSearch like the name of variables, lists, the discarding torrent check and etc

fernandog:
compare which def exists and which not. like the image def it's not necessary

sickbeard/providers/__init__.py
```diff
@@ -26,6 +26,7 @@
 from sickbeard.providers import btn, newznab, womble, thepiratebay, torrentleech, kat, iptorrents, torrentz, \
     omgwtfnzbs, scc, hdtorrents, torrentday, hdbits, hounddawgs, nextgen, speedcd, nyaatorrents, animenzb, bluetigers, cpasbien, fnt, xthor, torrentbytes, \
     freshontv, titansoftv, libertalia, morethantv, bitsoup, t411, tokyotoshokan, shazbat, rarbg, alpharatio, tntvillage, binsearch, torrentproject, extratorrent, \
```
duramato, line 28: @JohnDooe This file cant be touched ,it should only inclued your provider..

Fen0l, line 28: I dont use frequently git but at the beginning, it was something like that :) https://github.com/SickRage/SickRage/pull/118/files Dont know why it does that

duramato, line 28: @JohnDooe the sickbeard/providers/**init**.py changed format ,so you need to update it

miigotu:
The code in this provider is totally broken and would not behave properly. Without an account to use for grabbing html or some page source I can't even fix it up correctly.

Closing until someone can write it correctly or provide the necessary html or account.

PR #351, opened 2015-12-02
Remove duplicate function definition

MGaetan89:
The `to_int` function is redundant, as we have a `try_int` function, which is tested.
I also used the `try_int` function in a couple more places.

### DIFF
--- a/sickbeard/config.py
+++ b/sickbeard/config.py
@@ -28,6 +28,7 @@ from sickbeard import logger
 from sickbeard import naming
 from sickbeard import db
 
+from sickrage.helper.common import try_int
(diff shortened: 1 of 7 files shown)


miigotu:
Only curious about the to_int in config, are we sure that is right?

MGaetan89:
I guess it is.
What's bugging you about it?

miigotu:
Idk, the to_int in config comes from the lib for handling config. Just felt like thats the one it should use lol. Its ok though.

MGaetan89:
I didn't know that it came from a lib.
`to_int` and `try_int` have almost the same code (just syntactic differences). So it should still work as before.

PR #357, opened 2015-12-03
merge recommendedShow and trendingShow templates and add more trakt l…

Thraxis:
…ists support
- remove addShows_recommendedShows.mako
- remove recommendedShows code from webserve.py
- add traktList option to trendingShows functions in webserve.py
- allow for any of the 7 trakt lists to be called and displayed
- modify trendingShows.mako to display trakt placeholder image for shows
  without a poster

Lists supported:

Trending - Returns all shows being watched right now.
Popular - Returns the most popular shows.
Played - Most played shows in the past week
Watched - Most watched shows in the past week
Collected - Most collected shows in the past week
Anticipated - Most anticipated show based on the number of lists it
appears in
Recommended - Personalized show recommendations for a user.


Thraxis:
Please add a Do Not Merge tag to this.

@OmgImAlexis This is the start of the template merge and support for all the trakt lists

Note to self:  Need to look into trakt auth errors when asking for pages that don't need auth.

OmgImAlexis:
Looks good so far.

Thraxis:
It needs some luvin' for passing the traktList to the trendingShows.mako in the javascript code

Thraxis:
OK... I think I got everything all covered now except for a core.min.js

trending is now "Trakt Lists" and lets you choose any of the lists from Trakt.  if you have your Trakt setup it will show recommended.

image

image

gui/slick/views/addShows.mako
```diff
@@ -19,32 +19,22 @@
...
+
     <br><br>
-    <a href="${srRoot}/addShows/trendingShows/" id="btnNewShow" class="btn btn-large">
+    <a href="${srRoot}/addShows/trendingShows/?traktList=anticipated" id="btnNewShow" class="btn btn-large">
         <div class="button"><div class="icon-addtrendingshow"></div></div>
         <div class="buttontext">
-            <h3>Add Trending Show</h3>
-            <p>For shows that you haven't downloaded yet, this option lets you choose from a list of current trending shows with ratings to add, creates a directory for its episodes, and adds it to SickRage.</p>
+            <h3>Add From Trakt Lists</h3>
+            <p>For shows that you haven't downloaded yet, this option lets you choose from a show from one of th Trakt lists to add to SickRage .</p>
```
OmgImAlexis: "show from one of th Trakt lists", missing an "e".

OmgImAlexis:
Do you have grunt and everything setup to build the min file?

Thraxis:
No I don't have grunt

OmgImAlexis:
You should squash those commits into one.

https://github.com/SickRage/sickrage-issues/wiki/How-to-setup-a-development-enviroment

Thraxis:
The letter e is fixed.

miigotu:
@OmgImAlexis you see how the layout is not centered? Is that easy to fix? It is the same way on imdbPopular

OmgImAlexis:
@miigotu that's because of the stupid layout. For some reason we have divs within divs within div... :-1: 

The easiest way would to be add something like this so the container is resized and then we can just use something like `margin: 0 auto;` to center the div itself.

```
var width = Math.floor($(window).width() / $('.track_show').width()) * $('.track_show').width();
$('#container').width(width);
```

Thraxis:
The other question is do we really need the Manage Directories and Customize Options at the top of those pages?

OmgImAlexis:
@Thraxis yes because when you add a show from them it uses the settings on the page, if we didn't have it then the user couldn't do anything.

Thraxis:
It doesn't use those settings other then that's another place to over-ride the defaults.

The addShow just uses whatever your default show settings are.

Thraxis:
This should be complete now (core.min.js added)

@miigotu @OmgImAlexis The space on the right of the layout is from isotope, the container div itself is already centered, isotope left aligns the items it re-flows.

OmgImAlexis:
@Thraxis yes I know this is why I was saying to change the size of the div itself. I've use the same method before on my Tumblr blog. Basically you just need to make it so the div is exactly the size of the divs within it and then center the div on the page.

sickbeard/webserve.py
```diff
@@ -2602,7 +2609,7 @@ def existingShows(self):
 
     def addShowByID(self, indexer_id, show_name, indexer="TVDB"):
 
-        if indexer != "TVDB":
+        if indexer is not "TVDB":
```
miigotu: This is wrong, leave it !=

sickbeard/webserve.py
```diff
@@ -2530,20 +2520,37 @@ def getTrendingShows(self):
...
-            if sickbeard.TRAKT_BLACKLIST_NAME is not None and sickbeard.TRAKT_BLACKLIST_NAME:
-                not_liked_show = trakt_api.traktRequest("users/" + sickbeard.TRAKT_USERNAME + "/lists/" + sickbeard.TRAKT_BLACKLIST_NAME + "/items") or []
+            if sickbeard.TRAKT_ACCESS_TOKEN != '':
+                library_shows = trakt_api.traktRequest("sync/collection/shows?extended=full") or []
+                if sickbeard.TRAKT_BLACKLIST_NAME is not None and sickbeard.TRAKT_BLACKLIST_NAME:
+                    not_liked_show = trakt_api.traktRequest("users/" + sickbeard.TRAKT_USERNAME + "/lists/" + sickbeard.TRAKT_BLACKLIST_NAME + "/items") or []
+                else:
+                    logger.log(u"Trakt blacklist name is empty", logger.DEBUG)
+
+            if traktList is not "recommended":
```
miigotu: need to use ==/!= for string comparison

PR #358, opened 2015-12-03
Enable strict_undefined for mako

miigotu:
Fix a typo in a mako variable name
Fix a few missed imports and arguments that were not passed


miigotu:
Merging this for now, will do the other stuff in another PR.

PR #359, opened 2015-12-03
Only show downloaded and archived in status selections instead of every quality possible. Quality will …

miigotu:
…come from the file

Cons: User will no longer be able to change the quality of an episode (Downloaded(SD) => Downloaded(720p HDTV)), only the status. This might be a good thing though, and since we determine the quality accurately from the file, this should be ok?

Uncertain about this, but it makes the status dropdowns so short and pretty.

TODO: logFailed when setting status to failed (why didnt it do this before?)

### DIFF
--- a/gui/slick/views/displayShow.mako
+++ b/gui/slick/views/displayShow.mako
@@ -230,14 +230,12 @@
     <div class="pull-left" >
         Change selected episodes to:</br>
         <select id="statusSelect" class="form-control form-control-inline input-sm">
-        <% availableStatus = [WANTED, SKIPPED, IGNORED, FAILED] %>
+        <% availableStatus = [WANTED, SKIPPED, IGNORED, FAILED, DOWNLOADED, ARCHIVED] %>
         % if not sickbeard.USE_FAILED_DOWNLOADS:
         <% availableStatus.remove(FAILED) %>
         % endif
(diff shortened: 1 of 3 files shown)


fernandog:
what if user lost his collection or deletes files after watch? or he downloads a file manually that SR couldn't snatch and it's not in his quality list - like "this episode I want to watch bluray"?
I vote no for this.

miigotu:
This doesnt prevent you from changing statuses. It only auto-sets the QUALITY, so you dont have that option anymore. This eliminates a lot of problems, and SR autodetects the right quality quite well. If you changed the file manually.. click rescan

You can still change downloaded to archived or wanted or w/e. If it is an archived status it will use the quality it was downloaded as.

This cleans up lot of code, makes the drop downs much smaller and easier to navigate, and stops people from marking things as the wrong quality and then complaining when things arent going right.

Besides all of that, this is still in progress.

duramato:
@miigotu when a file is unknown sr only detects for mp4 and mkv ,that a big handicap and ome of the biggest reasons i use it to set the downloaded qualities

duramato:
@miigotu what could be done is move the archived and download maybe to another dropdown,that way it would clean that one

miigotu:
will just leave it as is for now

PR #364, opened 2015-12-03
 Minor Style Changes to Tighten Poster View Shows 

hernandito:
I have made additional tweaks to the styling of the new Poster View.
These changes, tighten up the info shown on each show. This tightening,
allows you to display more shows at the same poster size, at the same
browser resolutions.

Please refer to screen capture comparison on the below image:


OmgImAlexis:
Since this changes the CSS we'd rather not merge this at the moment. Like I've said before changing the CSS can cause a lot of issues and we're trying to stick to just JS for changing the layout.

hernandito:
@OmgImAlexis The edits I did were slight number changes to what was done by VinceVal to the CSS to get the current adjustable poster sizes.

I just changed a few of his padding, margin and position values to tighten up the content in the white space below the poster. The changing of the layout and the JS does NOT do the positioning of the elements I adjusted. VinceVal is doing this in the css and I only adjusted what he did.

medariox:
@hernandito 
Please squash the PR.
@OmgImAlexis 
What's wrong with changing the CSS? Can you let us know please?

Personally, I like the current version more, but we should still discuss what to do together.

hernandito:
My idea about tightening the stuff in the white area below, is that one can see more shows at the same time while viewing at the same poster size and browser window size. It looks decent on tablets when its smaller.

OmgImAlexis:
@medariox we're trying to stay away from too many UI changes because we need to refactor a LOT of it.

This will stay closed for now.

PR #383, opened 2015-12-05
Fix 'Clear filters' button and columnSelector on mass edit page

miigotu:
Select columns still does not work there. I think the table needs merged with the other one and the widget added? @OmgImAlexis can you handle that part?


miigotu:
nvm @OmgImAlexis I got it xDD

miigotu:
@OmgImAlexis can you please review that my js changes were correct, and then merge ASAP? Works here

PR #385, opened 2015-12-05
fix wonky br tags

alisonatwork:
Was thinking of going through the inputs and imgs with trailing slashes too, but it got a bit too big for now.


OmgImAlexis:
I've done this a few times already. :+1:
We've got a LOT of old code that's still using HTML4 standards. :-1:

PR #386, opened 2015-12-05
upgrade beautifulsoup to 4.4.1 and use html5lib for parsing everywhere

alisonatwork:
Okay this is a pretty epic PR and it needs a BUNCH of testing. Because for some reason BeautifulSoup is also wrapped by another class (BS4Parser) this affects pretty much every single scraping solution in the whole app.

It all started because adding shows via IMDB was broken for me (parse error). But when I changed the parser to html5lib, BeautifulSoup couldn't find anything deeper than about 5 divs, and the table we want is about 10 nested elements in. I am not sure if this is a Windows-only bug, or if this has been broken for a while.

After writing code to manually walking down the node tree myself checking attributes one at a time, i realized the tree was built okay, but the find()/find_all() method just couldn't handle it for some reason. The only explanation was a bug in BeautifulSoup. Sure enough, I copied in the latest version and it suddenly was okay again.

@miigotu I have never upgraded a Python library before, so I don't know if this is the idiomatic way to do it (I just copied it over from the source tarball). Please can you review and make sure this is okay.

In any case, consider this a proposal for now...


alisonatwork:
Ahhh crap I just noticed I didn't update all the BS4Parser usages to replace the non-existent "features" argument to simply html5lib as string. This is gonna cause tons of warnings. Hang a sec...

alisonatwork:
I am not sure if this is a related error, but I have started to see these:

```
C:\SickRage\SickRage\lib\imdb\parser\http\utils.py:472: UserWarning: unable to u
se "lxml": No module named lxml.html
  warnings.warn('unable to use "%s": %s' % (mod, str(e)))
C:\SickRage\SickRage\lib\imdb\parser\http\utils.py:463: UserWarning: falling bac
k to "beautifulsoup"
  warnings.warn('falling back to "%s"' % mod)
```

It still works because BeautifulSoup is fine. It appears to more specifically be a problem around not finding lxml any more somehow.

miigotu:
There must still be a way to set "persmissive" right? Without it half of the privte providers might not work, because they have broken html sometimes.

Why added bs5 ?

Also that error, need to find a way to monkeypatch that down. It is unrelated and comes from the recently update imdb lib..

alisonatwork:
Ahhh crap, bs5 was my backup directory for the old bs4. Will delete.

Regarding "permissive", I can't find any reference to this on the BeautifulSoup site, I don't think it is something that is part of any recent version. Or is it perhaps an additional argument to html5lib? If it is, then the format is definitely different, because from 4.4.0 it starts throwing warnings all over the place. It might be that we can leave the old "features" arg there and just add 'html5lib' as the second arg to avoid warnings. I only use a couple of the providers myself so I haven't been able to test this as thoroughly as it should be.

alisonatwork:
Re this blog post: http://www.crummy.com/2011/02/20/0 it looks like "permissive" was a sort of generic option introduced many years ago that isn't really used any more. All the recent docs I could find recommend using html5lib and that's it.

miigotu:
If html5lib is explicitly designated as the builder, the features are defaulted with "permissive"
`features = ['html5lib', PERMISSIVE, HTML_5, HTML]`

However, the way you are calling it does not specify a builder, but rather the features because of the argument order:
`def __init__(self, markup="", features=None, builder=None,`

This leaves `builder` as `None`, which later in the `__init__`:

```
        if builder is None:
            if isinstance(features, basestring):
                features = [features]
            if features is None or len(features) == 0:
                features = self.DEFAULT_BUILDER_FEATURES
            builder_class = builder_registry.lookup(*features)
```

And:   `DEFAULT_BUILDER_FEATURES = ['html', 'fast']`

So your call should `builder='html5lib'` OR set `features=` like we did before I think.

Documentation and blogs should probably always be assumed to be outdated imho =P

There are a few providers that intentionally or unintentionally have broken html that BS4 cannot parse without permissive.

miigotu:
About the lxml warnings:
At the top of SickBeard.py you can do:

```
import warnings
warnings.simplefilter('ignore')
```

Or add it to the bs4_parser, and anywhere bare bs4 is used.

alisonatwork:
I see the code you are talking about, but I don't think it makes any difference to the actual treebuilder that gets chosen. As indicated in the docs and supported in the code (see bs4/builder subdirectory), there are only three treebuilders - html5lib, htmlparser and lxml. If you specify a string as the second argument, then it explicitly picks one of those treebuilders, and skips the feature-detection code.

The problem is that since 4.4.0, if you do not specify an explicit treebuilder as a string on the second argument, this warning will pop up:

```
            if not (original_features == builder.NAME or
                    original_features in builder.ALTERNATE_NAMES):
                if builder.is_xml:
                    markup_type = "XML"
                else:
                    markup_type = "HTML"
                warnings.warn(self.NO_PARSER_SPECIFIED_WARNING % dict(
                    parser=builder.NAME,
                    markup_type=markup_type))
```

So, although I could change the code back to trying to find the treebuilder based on a feature list, it avoids a warning to choose it explicitly. And even if we did indicate "permissive", the builder_lookup would end up returning html5lib as the treebuilder anyway, since that's what was happening silently (without warning) in 4.3.x.

sickbeard/providers/elitetorrent.py
```diff
@@ -65,7 +65,11 @@ def __init__(self):
 
         }
 
+<<<<<<< HEAD
     def _do_search(self, search_strings, search_mode='eponly', epcount=0, age=0, epObj=None):
+=======
+    def _doSearch(self, search_strings, search_mode='eponly', epcount=0, age=0, epObj=None):
+>>>>>>> update parser selector for all the BS4Parser usages
```
alisonatwork: ooops hold up

miigotu:
@alisonatwork `permissive` isnt a tree builder, it is an option for html5lib.

alisonatwork:
Yes, I realize that, but my point is that the "features" argument is overloaded. If you pass a string, it will be wrapped in an array, and then will be used to look up the builder. But "features" argument is also stored in "original_features", and there the original string (non array) is kept and used to compare if the builder you got back is the same as the one you requested. If that is not the case, then a warning will be printed. This is new behavior in 4.4.0, presumably done precisely to avoid the case where people where still using fuzzy logic circa 2011 to select the builder instead of explicitly naming it, which appears to be the recommended method going forward.

Sorry if I can't explain this well, it would be much easier in person looking at the code.

alisonatwork:
Here is the changelog entry from their readme:

```
* Added a warning when you instantiate a BeautifulSoup object without
  explicitly naming a parser. [bug=1398866]
```

miigotu:
If you dont want to pass it as `features=` you have to explicitly pass it as `builder='html5lib'` because `features` is the second arg, `builder` is the third. So if you pass just `soup(data, "html5lib")` it overrides the default features of `html5lib` and disables the `permissive` option of `html5lib` and it will then barf parsing any html5 that is not strictly perfect.

Its just the way it has to be, without permissive enabled in the html5lib module half of our providers do not work. I have done extensive testing and this exact issue has caused many problems in SR before.

Regardless of the version of bs4, the parameter order and handling of the features has not changed. The warnings have, but that is a totally separate issue.

For the warnings add this to SickBeard.py:

```
import warnings
warnings.simplefilter('ignore')
```

We cant have random output to stdout outside of our control or it can break pythonw on windows.

alisonatwork:
I'm sorry, but I don't think you are looking at the current version of the code. The features argument is only used to select which builder is used. It is not used to alter the behavior of the builder. "permissive" just means "pick one of the builders that has the permissive flag on it". Our code worked by accident because we specified both "html5lib" and "permissive", so "permissive" would cut the available parsers down to two (html5lib and lxml), and then "html5lib" would select html5lib from those two.

Please read this bug to see why they changed this behavior: https://bugs.launchpad.net/beautifulsoup/+bug/1398866

They want you to explicitly to specify the builder always, and not try to detect it via fuzzy logic, because fuzzy logic detection is unpredictable on different platforms and can break things mysteriously.

miigotu:
Big problem with that is that not every system we support has html5lib, or lxml. We dont use a setup.py or requirements.txt (not all environments we support even support pip or distutils) so we can't explicitly require html5lib for the project.

It's no sense in going back and forth over it, if it breaks it will just be on you to fix it anyways =P

alisonatwork:
You are right this is opening us up to the possibility of undefined behavior on systems without html5lib. I mean, we could stay on the old soup, or use fuzzy detection and hide the warning, but then there is a continued risk of strange scraping bugs popping up that we can't easily recreate. I don't think there is a perfect answer for this, hence listing this as a proposal and not specifically a bugfix :) If you and the rest of the "core team" is happy with it then awesome, but if you feel it's too risky I can either close this or modify it and hide the warning as you suggested.

miigotu:
Im just not really sure if it will break or not, hate being in that state of mind lol.

As for the actual parsing, I have always tried to write my parsing code in such a way that it should work with both parsers, and let the fuzzy work its wuzzy to select a builder that exists on the system (preferring the faster lxml would be best but meh).

There are a lot of places however that were using only html5lib to begin with so it might be ok.

alisonatwork:
I just noticed we include a copy html5lib inside our lib/ dir anyway, and html5lib is pure Python so shouldn't need any OS-specific libraries. Feeling more confident now :) I will keep running PR this on my "real" install and test install for a while and let you know if I hit any problems.

fernandog:
@alisonatwork I have some invites from private trackers: TD, AR and HDT. If you want just send you email via PM in IRC

miigotu:
@alisonatwork I was unaware we had html5lib in libs so I guess all of my anxiety was for naught....

miigotu:
Since we just released, I think now is a good time to let the other team/users on develop try this out. That will give it the longest incubation possible in develop.

PR #387, opened 2015-12-05
Changed move associated files to delete associated files, adds possib…

medariox:
…ilty to discard srt files.

_What is this change about?_
SickRage's default behavior is to move all associated files. That's why the option "move associated files" didn't really make any sense, since it would already do it by default.
Now the option is called "delete associated files". It allows, in combination with the "Keep associated file extensions" option, to let the users actually choose which associated files to keep. E.g. by adding "srt,nfo" to the input field, srt and nfo files will be moved during post-processing. The empty input field would mean that all associated files should be deleted. This also adds the ability to move only the video file, which previously wasn't possible (subtitles were always moved).

### DIFF
--- a/gui/slick/views/config_postProcessing.mako
+++ b/gui/slick/views/config_postProcessing.mako
@@ -41,7 +41,7 @@
                             </label>
                             <label class="nocheck" for="process_automatically">
                                 <span class="component-title">&nbsp;</span>
-                                <span class="component-desc"><b>NOTE:</b> Do not use if you use an external PostProcessing script</span>
+                                <span class="component-desc"><b>NOTE:</b> Do not use if you use an external Post Processing script</span>
                             </label>
                         </div>
                         <div class="field-pair">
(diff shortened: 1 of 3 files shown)


sickbeard/__init__.py
```diff
@@ -283,7 +283,7 @@
 TV_DOWNLOAD_DIR = None
 UNPACK = False
 SKIP_REMOVED_FILES = False
-ALLOWED_EXTENSIONS = "nfo,srr,sfv"
+ALLOWED_EXTENSIONS = "srt,nfo,srr,sfv"
```
fernandog: @medariox can

gui/slick/views/config_postProcessing.mako
```diff
@@ -138,18 +138,18 @@
                         <div class="field-pair">
                             <input type="checkbox" name="move_associated_files" id="move_associated_files" ${('', 'checked="checked"')[bool(sickbeard.MOVE_ASSOCIATED_FILES)]}/>
                             <label for="move_associated_files">
-                                <span class="component-title">Move Associated Files</span>
-                                <span class="component-desc">Move srr/sfv/etc files with the episode when processed?</span>
+                                <span class="component-title">Delete associated files</span>
+                                <span class="component-desc">Delete srt/srr/sfv/etc files while post processing?</span>
```
fernandog, line 142: @medariox can't test because of the dev issue. PP page don't load

maybe we should remove this and change the other too
<span class="component-title">Delete all associated files while post processing</span>

PR #388, opened 2015-12-06
Add nzbindex and nzbfriends providers

Comptezero:
backported from https://github.com/cytec/SickRage


miigotu:
Both of these are in the wrong format for SickRage. CopyPasta doesn't work with SR anymore.

mcarbonneaux:
what the good format ?

miigotu:
the same as all of our other providers...

labrys:
@Comptezero do you plan on updating this to the new format?

miigotu:
@labrys I just left this open to remind me and be able to use it as a reference, it is based on master ^

miigotu:
nzbindex rss no longer works, and nzbfriends is random on what words in the q term it uses and is not giving proper nzb names. I have an nzbfriends provider here that may work better in the future.

noodle-:
Just wanted to mention that nzbindex + rss is working smoothly now. No more 503's!

I'll try to tinker with the provider script if nobody else has made one by next week.

PR #389, opened 2015-12-06
Update showUpdater to use theTVDB update files

Thraxis:
- Stores last_update time in cache.db lastUpdate table
- If value is not there it inserts it with a value of current time -
  86400 seconds
- Compares update time to last update time to select which update XML
  file to call from the TVDB
  - Less than 25 hours will pull the updates_day.xml
  - Over 25 hours but less than 8 days will pull the updates_week.xml
  - Anything over 8 days will pull the update_month.xml
- Basic support for multiple indexers is in place
- Array for bad indexers (aka tvRage) included to post warning messages
  in the log for shows when an indexer is removed
- Will update any show regardless of status (Ended / Continuing)

### DIFF
--- a/sickbeard/showUpdater.py
+++ b/sickbeard/showUpdater.py
@@ -16,6 +16,9 @@
 # You should have received a copy of the GNU General Public License
 # along with SickRage.  If not, see <http://www.gnu.org/licenses/>.
 
+import xml.etree.ElementTree as ET
+import requests
+import time
 import datetime
 import threading
 import sickbeard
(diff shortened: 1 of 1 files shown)


Thraxis:
I'm going to call this a Step 1 to the showUpdater 

Next steps:
-  follow refreshShow() path and see what exactly it does and what can be cleaned up
-  follow updateShow() path and see what exactly it does and what can be cleaned up
-  implement TheTVDB APIv2 support for update

Note: I expect these steps to be modular as in not having to wait for everything to be commited.  This commit should lower the number of Forced Updates that are done each night during the showUpdate process.

Thraxis:
Per discussion in IRC....  I will be removing the refreshShow calls from showUpdater later in a PR update.  These calls are where the bulk of time in the showUpdate process actually goes.

PR #396, opened 2015-12-06
Fix named-parameter

MGaetan89:
https://github.com/SickRage/sickrage-issues/issues/164

### DIFF
--- a/sickbeard/providers/bitcannon.py
+++ b/sickbeard/providers/bitcannon.py
@@ -46,7 +46,7 @@ class BitCannonProvider(TorrentProvider):
             'apiKey': ''
         }
 
-    def _do_search(self, search_strings, search_mode='eponly', epcount=0, age=0, epObj=None):
+    def _do_search(self, search_strings, search_mode='eponly', age=0, ep_obj=None):
         # search_strings comes in one of these formats:
         #      {'Episode': ['Italian Works S05E10']}
         #      {'Season': ['Italian Works S05']}
@@ -55,7 +55,7 @@ class BitCannonProvider(TorrentProvider):
         items = {'Season': [], 'Episode': [], 'RSS': []}
 
         # select the correct category (TODO:  Add more categories?)
-        anime = (self.show and self.show.anime) or (epObj and epObj.show and epObj.show.anime) or False
+        anime = (self.show and self.show.anime) or (ep_obj and ep_obj.show and ep_obj.show.anime) or False
         self.search_params['category'] = ('tv', 'anime')[anime]
 
         # Set API Key (if applicable)
--- a/sickbeard/providers/kat.py
+++ b/sickbeard/providers/kat.py
@@ -62,12 +62,12 @@ class KATProvider(TorrentProvider):
 
         self.cache = KATCache(self)
 
-    def _do_search(self, search_strings, search_mode='eponly', epcount=0, age=0, epObj=None):
+    def _do_search(self, search_strings, search_mode='eponly', age=0, ep_obj=None):
         results = []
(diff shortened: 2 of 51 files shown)


sickbeard/providers/bitcannon.py
```diff
@@ -55,7 +55,7 @@ def _do_search(self, search_strings, search_mode='eponly', epcount=0, age=0, epO
         items = {'Season': [], 'Episode': [], 'RSS': []}
 
         # select the correct category (TODO:  Add more categories?)
-        anime = (self.show and self.show.anime) or (epObj and epObj.show and epObj.show.anime) or False
```
MGaetan89, line 58: Is there a reason for that `or False`?

sickbeard/providers/kat.py
```diff
@@ -62,12 +62,12 @@ def __init__(self):
 
         self.cache = KATCache(self)
 
-    def _do_search(self, search_strings, search_mode='eponly', epcount=0, age=0, epObj=None):
+    def _do_search(self, search_strings, search_mode='eponly', age=0, episode=None):
         results = []
         items = {'Season': [], 'Episode': [], 'RSS': []}
 
         # select the correct category
-        anime = (self.show and self.show.anime) or (epObj and epObj.show and epObj.show.anime) or False
```
MGaetan89, line 70: Is there a reason for that `or False`?

sickbeard/providers/bitcannon.py
miigotu, line 58: Doesnt look like it, faintly remember something about that though, just doesnt make sense though xD

miigotu:
The epObj param cant be renamed to episode, it needs to keep the designation that it is an instance of tv.TVEpisode and not an episode number. ep_obj or similar if you must change it.

sickbeard/providers/bitcannon.py
MGaetan89, line 58: Not really an issue, but I was surprised that my IDE does not warn me that this can be simplified :P

MGaetan89:
Right, I did not think of that. I will change to `ep_obj`.

miigotu:
BTW, really hating pyCharm for telling you guys to rename all of these variables. Thats like the last thing you should ever worry about doing, especially during moving code around. It only makes it harder to keep track of the changes and keep bugs out of the refactor, and more annoying for people trying to write new code when they are used to the old variables. Not only that pyCharm doesnt even find them all like it should (even a normal find in files would have found them all)

How many bugs would you have had with the first PR if it werent for renaming variables? =P

MGaetan89:
We're trying to do a too good job by normalizing the code. But as you told me a couple times already, we could do it in multiple PRs instead of one big one.

I change the `episode` to `ep_obj`.

miigotu:
I think we need to document some guidelines for stages of refactors and a specific style guide for the future. I dont mind all the changes, its just too chaotic when stages are jumbled all together. I understand you know whats going on with your changes, it's just so hard for anyone else to follow it that way.

Im not really getting all over your PR here or saying you are doing something wrong. Its just the fact that we dont have a long pre-production stage like most projects/companies, we are almost direct to market. 

More and more with every big PR the need for some "refactoring roadmap" for each refactoring task and some guidelines is becoming more and more obvious. I really think it will help bring clarity to any PR for all contributors, and result in better productivity and code with less bugs (even in the short term)

miigotu:
Found and fixed a bug (preexisting to the providers PR) while testing this. Fixes a serious backlog/manual/propers/failed search bug. I dont think it was working at all? lol

Going to merge this now.

PR #397, opened 2015-12-06
Add provider danishbits

twjacobsen:
nextgen has closed, so this might be a great replacement for some

### DIFF
--- a/sickbeard/providers/danishbits.py
+++ b/sickbeard/providers/danishbits.py
@@ -1,0 +1,219 @@
+# Author: seedboy
+# URL: https://github.com/seedboy
+#
+# This file is part of SickRage.
+#
+# SickRage is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# SickRage is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


romanocry:
If i may, don't know if it's the right place, but Libertalia is close 
too...

miigotu:
This needs slightly modified, our providers layout is changing. I will pull it and fix it in the enxt few days.

sickbeard/providers/danishbits.py
```diff
@@ -0,0 +1,219 @@
...
+                    logger.log(u"Failed parsing provider. Traceback: %s" % traceback.format_exc(), logger.ERROR)
+
+            # For each search mode sort all the items by seeders if available
+            items[mode].sort(key=lambda tup: tup[3], reverse=True)
+
+            results += items[mode]
+
+        return results
+
+    def _convertSize(self, size):
```
MGaetan89, line 188: This method is defined in a bunch of provider and the if always (or almost always) the same.
Maybe we should move it to `GenericProvider` or add a method in `sickrage.helper.common`.

miigotu, line 188: The method is really new, and varies slightly on a lot of the providers. It needs to be slightly modified to handle them all and moved to generic, but I was waiting until it was implemented  in more and I knew all of the variations it would need to handle.

twjacobsen, line 188: Note: This specific piece of code is still untested, although it should be OK

twjacobsen, line 188: OK, so now I had time to test (and modify!) this provider, and it's fully functional

miigotu:
search_params and epcount was removed from _do_search paramteter format, _do_search was renamed to just search, and _do_login was renamed to login.

PR #399, opened 2015-12-06
Remove Libertalia and NextGen

duramato:
They are both dead 
Proof for nextgen ,http://pastebin.com/dqhnG43t


miigotu:
Oh, but "please keep changing the url for us, there is no legal troubles it is perfectly safe"...

Guess next time I follow my gut lol.

PR #401, opened 2015-12-06
Fix subs download button not showing for subs without language

medariox:
@miigotu There could be a better way to do this, but it works fine in my tests so far.

Related issue: Missed subtitle management does not work for subtitles without language. We should either count 'und' as a valid sub or remove the feature if MULTI SUBS is unchecked.

Other pseudo-related issues (also keeping this as a tracker for me):
- srRoot needs to be added here: https://github.com/SickRage/SickRage/blob/81aa99c0b4bd39fbf333decd6e4f6c155d9401ef/gui/slick/js/core.js#L2388 or flags won't show in subtitleManagement.
- Some html problems in missed subtitle management (not related to the missing flags):
  issue


miigotu:
You should have just improved the logic in needs_subtitles, so that other things that use it also work.  However, just because the subtitles field is empty for an episode does not automatically mean you should show the search, because if a user doesnt have any languages selected it will still show the icon.

medariox:
It won't show. I actually don't think that there is anything wrong with the code.

PR #407, opened 2015-12-07
Try to fix issue spam.

fernandog:
@miigotu what do you think?


miigotu:
Done differently in https://github.com/SickRage/SickRage/pull/409

PR #411, opened 2015-12-07
Set "min" and/or "step" value to input fields

fernandog:
@miigotu

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -404,7 +404,7 @@ $('#config-components').tabs();
                             <label for="${curTorrentProvider.get_id()}_ratio">
                                 <span class="component-title" id="${curTorrentProvider.get_id()}_ratio_desc">Seed ratio:</span>
                                 <span class="component-desc">
-                                    <input type="number" step="0.1" name="${curTorrentProvider.get_id()}_ratio" id="${curTorrentProvider.get_id()}_ratio" value="${curTorrentProvider.ratio}" class="form-control input-sm input75" />
+                                    <input type="number" min="-1" step="0.1" name="${curTorrentProvider.get_id()}_ratio" id="${curTorrentProvider.get_id()}_ratio" value="${curTorrentProvider.ratio}" class="form-control input-sm input75" />
                                 </span>
                             </label>
                             <label>
@@ -421,7 +421,7 @@ $('#config-components').tabs();
                             <label for="${curTorrentProvider.get_id()}_minseed">
                                 <span class="component-title" id="${curTorrentProvider.get_id()}_minseed_desc">Minimum seeders:</span>
                                 <span class="component-desc">
-                                    <input type="number" name="${curTorrentProvider.get_id()}_minseed" id="${curTorrentProvider.get_id()}_minseed" value="${curTorrentProvider.minseed}" class="form-control input-sm input75" />
+                                    <input type="number" min="0" step="1" name="${curTorrentProvider.get_id()}_minseed" id="${curTorrentProvider.get_id()}_minseed" value="${curTorrentProvider.minseed}" class="form-control input-sm input75" />
                                 </span>
                             </label>
                         </div>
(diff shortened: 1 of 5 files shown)


OmgImAlexis:
Before changing anything like this it's a good idea to check [caniuse.com](http://caniuse.com/#search=number), there's known issues with the number type.

fernandog:
We already have that others fields. Just added to the missing ones.
What issue can cause?

OmgImAlexis:
Check the known issues tab on the page I linked too.

fernandog:
I did and didnt understand.

OmgImAlexis:
Well this is a major one. "The .stepUp and .stepDown methods incorrectly invoke InvalidStateError exceptions in IE 10 and 11." This means we could end up with the inputs that use these not being able to be used with any js including ajax on IE10 and IE11.

miigotu:
Need to change all of these to jquery ui spnner I think
https://jqueryui.com/spinner/

fernandog:
But we dont use those in JS isnt?
Webserve reads directly from the component isnt?

gui/slick/views/config_providers.mako
```diff
@@ -432,7 +432,7 @@ $('#config-components').tabs();
                             <label for="${curTorrentProvider.get_id()}_minleech">
                                 <span class="component-title" id="${curTorrentProvider.get_id()}_minleech_desc">Minimum leechers:</span>
                                 <span class="component-desc">
-                                    <input type="number" name="${curTorrentProvider.get_id()}_minleech" id="${curTorrentProvider.get_id()}_minleech" value="${curTorrentProvider.minleech}" class="form-control input-sm input75" />
+                                    <input type="number" min="1" step="1" name="${curTorrentProvider.get_id()}_minleech" id="${curTorrentProvider.get_id()}_minleech" value="${curTorrentProvider.minleech}" class="form-control input-sm input75" />
```
duramato: This should not be enforced

fernandog: We already do that for new installs. Default value

duramato: Default value is not min value

miigotu:
huh? you apply that class on the input...

fernandog:
Avoid negative numbers. Float numbers. Etc

gui/slick/views/config_providers.mako
miigotu: default value for minleech is 0.

OmgImAlexis:
@miigotu can you have a look through the caniuse page, @fernandog doesn't seem to understand it.

Basically the fields shouldn't be used until IE10 and IE11 support is added otherwise we could end up breaking functionality on there.

miigotu:
@OmgImAlexis I will then, but I think ie10/11 just treat them as a text input (no fancy controls)

Arcanemagus:
That's correct, when a browser doesn't understand the type given to an input element it defaults it to text, so you don't get the fancy controls and built in validation, but the end result sent back to the server is the exact same.

gui/slick/views/config_search.mako
```diff
@@ -346,7 +346,7 @@
                                 <label>
                                     <span class="component-title">NZBget host:port</span>
                                     <span class="component-desc">
-                                        <input type="text" name="nzbget_host" id="nzbget_host" value="${sickbeard.NZBGET_HOST}" class="form-control input-sm input350" autocapitalize="off" />
+                                        <input type="text" min="1" step="1" name="nzbget_host" id="nzbget_host" value="${sickbeard.NZBGET_HOST}" class="form-control input-sm input350" autocapitalize="off" />
```
miigotu, line 349: type="number"

Fix that and we can merge this for now and worry about jquery ui spinner later. This actually will make it easier to regex for replace.

miigotu:
Will remake PR for you, cant use numeric for host =P

gui/slick/views/config_search.mako
fernandog, line 349: Sorry!  I thought it was only the port.

PR #416, opened 2015-12-08
Set "min" value to input fields

miigotu:
Fixup #411 for @fernandog

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -404,7 +404,7 @@ $('#config-components').tabs();
                             <label for="${curTorrentProvider.get_id()}_ratio">
                                 <span class="component-title" id="${curTorrentProvider.get_id()}_ratio_desc">Seed ratio:</span>
                                 <span class="component-desc">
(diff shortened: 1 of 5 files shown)


gui/slick/views/config_providers.mako
```diff
@@ -404,7 +404,7 @@ $('#config-components').tabs();
                             <label for="${curTorrentProvider.get_id()}_ratio">
                                 <span class="component-title" id="${curTorrentProvider.get_id()}_ratio_desc">Seed ratio:</span>
                                 <span class="component-desc">
-                                    <input type="number" step="0.1" name="${curTorrentProvider.get_id()}_ratio" id="${curTorrentProvider.get_id()}_ratio" value="${curTorrentProvider.ratio}" class="form-control input-sm input75" />
+                                    <input type="number" min="-1" step="0.1" name="${curTorrentProvider.get_id()}_ratio" id="${curTorrentProvider.get_id()}_ratio" value="${curTorrentProvider.ratio}" class="form-control input-sm input75" />
```
fernandog, line 407: The only issue here is user can set -0.5 which is wrong. This didnt fix it but fix users setting values <-1 like -2

miigotu, line 407: because step is a float. ofc we need float for ratio, because stepping by ones is crazy there.
Solution: handle it in webserve when saving. if it is < 0 set it to -1

PR #436, opened 2015-12-09
Hacky way to fix funky names quality on tvchaosuk

duramato:
When the files get PPed SR gets the quality from the file but on parsing it sets as unknow
This works becouse almost alll files have x264 codec and would be safe to assume that..

### DIFF
--- a/sickbeard/providers/tvchaosuk.py
+++ b/sickbeard/providers/tvchaosuk.py
@@ -154,7 +154,7 @@ class TVChaosUKProvider(TorrentProvider):
                             freeleech = torrent.find('img', alt=re.compile('Free Torrent'))
                             if self.freeleech and not freeleech:
                                 continue
-                            title = torrent.find(attrs={'class':'tooltip-target'}).text.strip()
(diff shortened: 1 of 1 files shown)


duramato:
@whipcracker can you give an example?

whipcracker:
@duramato , for example, the SD release is "strictly-come-dancing-s13e22-week-11-5-dec-2015-mp4.torrent_[tvcuk].torrent"

I found that SR has a hard time reading the file name and understanding what quality the file is. Thoughts?

whipcracker:
Basically, I added "Strictly Come Dancing" and asked to search for "Unknown" quality, as well as SDTV or SD DVD. If it doesn't find either SDTV or SD DVD, it'll revert to "Unknown" and grab said file above or if there's a 720p version (where the file is actually named with 720p HDTV in the file name) it'll grab that instead.

I don't normally have a problem getting 720p versions of this show, but they're for my wife and quite honestly, she's watching them on a tablet, so 720p isn't REALLY required and she couldn't care less whether it was SD or HD lol...

Hope this helps :)

PR #439, opened 2015-12-10
Add API test suite.

labrys:
TODO: Implement API tests.

### DIFF
--- a/tests/api_v1_tests.py
+++ b/tests/api_v1_tests.py
@@ -1,0 +1,257 @@
+# coding=utf-8
+
+"""
+Test the SR API
+"""
+
+import unittest
+
+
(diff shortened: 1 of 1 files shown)


MGaetan89:
What's the points of adding TODOs?

OmgImAlexis:
@MGaetan89 I'm assuming so it's easier to go through and just add the tests.

labrys:
It gives you an idea in the unittest report how many sections are being skipped.  Similar to what coveralls provides, it also gives people an idea of what exactly still needs to be fleshed out.  I wanted to add some more commits that did some actual testing in there, but the test I was working on I didn't care for so I scrapped it.

PR #441, opened 2015-12-10
Let user choose to download perfect subs or not

fernandog:
@medariox 

score 111 matches only show + season + episode


duramato:
Shouldnt it come enabled by default? Makes more sense..

fernandog:
it's enabled by default

```
SUBTITLES_PERFECT_MATCH = bool(check_setting_int(CFG, 'Subtitles', 'subtitles_perfect_match', 1)) 
```

PR #451, opened 2015-12-11
Partialy fix SickRage/sickrage-issues/issues/196

fernandog:
Avoid thread crash if anything goes wrong with sql results

added try:except:continue

@miigotu @medariox ok?

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -439,6 +439,7 @@ class SubtitlesFinder(object):
 
         rules = self._get_rules()
         now = datetime.datetime.now()
+
         for ep_to_sub in sql_results:
 
             if not ek(os.path.isfile, ep_to_sub['location']):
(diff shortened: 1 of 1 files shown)


fernandog:
@miigotu @labrys so how come the thread crashes?

with this PR I get the log error and FINDSUBTITLES continues:

```
FINDSUBTITLES :: Error while searching subtitles for The Originals S02E08. Error: "time data 'al (imdb display ti' does not match format '%Y-%m-%d %H:%M:%S'"
```

without I get:
image

miigotu:
What does that have to do with our comments on the PR?

Just remove these lines

``` python
                # http://bugs.python.org/issue7980#msg221094
                # I dont think this needs done here, but keeping to be safe (Recent shows rule)
                datetime.datetime.strptime('20110101', '%Y%m%d')
```

miigotu:
Now, when that exception is thrown, you should update that episode's last search time to something like:

``` python
from sickrage.helper.common import dateTimeFormat
datetime.datetime(2015, 1, 1).strftime(dateTimeFormat)

```

That way the next time the subtitle finder is ran, the subs for that ep will be searched.

fernandog:
@miigotu can you commit this in this PR? need more code than that isn't?

PR #464, opened 2015-12-12
Fix archived status episodes not showing missed subs

fernandog:
@medariox @miigotu 
If ep with archived status has location in DB, video filename still exists.
Logs show that subs are missing but they didn't show in 'manage missed subtitles'


miigotu:
soon no archived eps will have location but this is ok for now.

PR #465, opened 2015-12-13
Add video file name to log to compare with subtitles filaname + score

fernandog:
We don't show video filename anywhere while searching for subs

@medariox  what was @Labrys suggestion? Where should I add "candidates" to log and changed downloading>verifying

update: found!


miigotu:
@fernandog agree with that commit I added?

fernandog:
Sure. Better.

PR #471, opened 2015-12-13
Move Torrent RSS specific code from sickbeard.providers.__init__ to sickbeard.providers.rsstorrent

MGaetan89:
In `sickbeard.provider.__init__.py`, some code are related to a specific provider/providers type.
This PR move the code dedicated to RSS torrent into the existing provider class (`sickbeard.providers.rsstorrent`).
I will send an other PR for the code dedicated to `NewznabProvider`.
The remaining code is generic enough to stay there (at least until all providers are moved into `sickrage.providers` module).

### DIFF
--- a/sickbeard/providers/rsstorrent.py
+++ b/sickbeard/providers/rsstorrent.py
@@ -64,6 +64,21 @@ class TorrentRssProvider(TorrentProvider):
             self.enable_backlog
         )
 
+    @staticmethod
+    def get_providers_list(data):
+        providers_list = [x for x in [TorrentRssProvider._make_provider(x) for x in data.split('!!!')] if x]
+        seen_values = set()
+        providers_set = []
+
+        for provider in providers_list:
+            value = provider.name
+
+            if value not in seen_values:
+                providers_set.append(provider)
+                seen_values.add(value)
+
+        return [x for x in providers_set if x]
+
     def image_name(self):
         if ek(os.path.isfile, ek(os.path.join, sickbeard.PROG_DIR, 'gui', sickbeard.GUI_NAME, 'images', 'providers', self.get_id() + '.png')):
             return self.get_id() + '.png'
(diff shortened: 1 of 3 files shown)


SickRage2:

Thanks!

sickbeard/providers/rsstorrent.py
```diff
@@ -64,6 +64,21 @@ def configStr(self):
...
+        providers_set = []
+
+        for provider in providers_list:
+            value = provider.name
+
+            if value not in seen_values:
+                providers_set.append(provider)
+                seen_values.add(value)
+
+        return [x for x in providers_list if x]
```
MGaetan89: I am not sure why we compute `providers_set` above. We don't use it anywhere.
Should it be used in the return statement, instead of `providers_list`? Or can I simply remove the `providers_set` logic?

labrys: Looks like the intention was to make a set from the providers_list to prevent duplicates.  Instead that could be done with a set comprehension:

``` python
providers_list = {x for x in [TorrentRssProvider._make_provider(x) for x in data.split('!!!')] if x}
```

labrys:
``` python
@staticmethod
def get_providers_list(data):
    providers = [TorrentRssProvider._make_provider(x) for x in data.split('!!!')]
    return list({x for x in providers if x})
```

Although the above doesn't check by name

labrys: How about:

``` python
@staticmethod
def get_providers_list(data):
    providers = [TorrentRssProvider._make_provider(x) for x in data.split('!!!')]
    provider_dict = {x.name: x for x in providers if x}
    return provider_dict.values()
```

MGaetan89: Your last suggestion good to me.
I'll update the PR accordingly in a minute.

MGaetan89: It's updated.

miigotu: Providers are sorted, you cant use a set

MGaetan89: So I should just change `providers_list` with `providers_set` in the `return`  statement of the original code?

labrys: Or use an ordered dict instead.

miigotu: `So I should just change providers_list with providers_set in the return statement of the original code?`

I think this is the best solution. I dont like code changes mixed with refactors, it makes it difficult to spot potential mistakes later if issues are created during the refactor.

labrys: @miigotu makes sense

MGaetan89: I will update the PR soon.

MGaetan89: Done.

PR #476, opened 2015-12-14
Remove archive on first match, it is replaced by comprehension method…

miigotu:
… of 2 quality lists

If qualities are set in both lists, a quality in the first list (allowed) will continue looking for a quality
in either list until a quality in the second list (preferred) is reached, and then will not search/replace it again
If there is only qualities in one list, or a predefined quality set is used, the first download will be final, and no more searches/replacements will be made
TODO: Rename the remainder of anyQualities, bestQualities to allowed_qualities, preferred_qualities and add some test suites


SickRage2:

Thanks!

fernandog:
@miigotu 
In my case. Most shows are:
1080p WEB in Allowed
1080p HDTV in preferred
Archived first match enabled.

SR download only once.

With this PR what will happen?

labrys:
@fernandog instead you would just have the qualities selected in a single list to get that functionality.

miigotu:
@fernandog unless you changed your quality settings it would try to get one of your preferred.

fernandog:
So before merge this we need to add it to NEWs explaining so users can have time to change all shows. Otherwise we will have tons of issues about redownloading.

fernandog:
And what if user deletes media. The downloaded status will remain? Cause today user can select deleted media default status'

What about all archived ep status that already exists in the database? We will convert it to downloaded?

miigotu:
archived quality/status has nothing to do with archive_on_first match, that was the majority of the confusion.

archive on first match was only needed before there were 2 quality lists, and simplifying it this way (to always stop searching for a better one if you have a preferred list and the episode is in preferred, or if you only have one list and have one of the qualities in that list) removes the need for the "archive_on_first_match" idea.

The next step in untangling this mess will be to convert archived episodes back to downloaded if the location is set and the file exists, because archived status is really only supposed to be for files that are deleted/missing after they were downloaded. This will take more care though, and will not be a part of this PR.

Deleted status still still work the same after this PR, and even after the changes to archived quality above ever makes it in.

fernandog:
ok.
shouldn't we fix first statuses and user database before merging or together  this PR?

what will happen if user have lots of snatched (best) and PP runs ? It may break. We have to convert snatched (best) to snatched

next DB upgrade we will remove/re-use the archive first match column from DB?

miigotu:
this PR does not effect snatched best. Trust me there is nothing you could think about that I havent already thought of with this. I have thought about this for almost 6 months and talked with a lot of people about it because no one seems to have the same understanding of it. The behavior after merging this will be identical to before without the confusion and unreliability.

fernandog:
ok!

EwolBash:
I don't like this method for a few reasons.

I want sickrage to prioritize 720 WEB-DL, but if the first found is 720 HDTV I don't want it to upgrade it to WEB-DL later. Having an option to automatically mark as archived on first download would be great.

Currently there seems no easy way to achieve this without manually selecting series as 'archived'. Excluding 720 HDTV is not an valid option as not every show gets WEB-DL releases.

Apologies if I'm misunderstanding something I only recently discovered sickrage.

miigotu:
If you just put both in the `allowed` list and nothing in the preferred list, it ill only grab one. This thread is over a year old.

EwolBash:
Yeah I tried that but it just seemed to download HDTV.

I know the thread is old, but github is easily searchable and I don't want to create new threads for every little issue I run into.

miigotu:
If it searches and the only thing it finds is HDTV then thats what its going to DL.  If it finds both HDTV and WEB-DL it will DL the WEB-DL.  Not all shows are released with WEB-DL anyways.

PR #481, opened 2015-12-14
Bypass subtitle search limits if we do a force search. Otherwise SR w…

fernandog:
…ill never download subs for those episodes.

@miigotu @medariox we already filter episodes in the IF. Doesn't need to filter the query. Agree?

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -110,7 +110,7 @@ def enabled_service_list():
 
 def wanted_languages(sql_like=None):
     wanted = frozenset(sickbeard.SUBTITLES_LANGUAGES).intersection(subtitle_code_filter())
-    return (wanted, '%' + ','.join(sorted(wanted)) + '%')[bool(sql_like)]
+    return (wanted, '%' + ','.join(sorted(wanted)) + '%' if sickbeard.SUBTITLES_MULTI else '%und%')[bool(sql_like)]
 
 
 def get_needed_languages(subtitles):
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

miigotu:
It filters in the query to prevent a result set with every episode in the database. Some users can have 10k episodes. Without limiting the results it is irresponsible with i/o and memory management.

fernandog:
hum ok. but that query will only return episodes without subs isn't? I tested in sqlite browser
So how to enable forced subtitle search for those episodes that search limit was reached?

```
2015-12-14 11:35:15 INFO     FINDSUBTITLES :: Subtitle search limit reached for * S03E09
2015-12-14 11:35:15 INFO     FINDSUBTITLES :: Subtitle search limit reached for * S03E08
2015-12-14 11:35:15 INFO     FINDSUBTITLES :: Subtitle search limit reached for * S03E10
```

miigotu:
will have to think about it

fernandog:
ok thanks

@miigotu about limits:
For legendastv. uploader/teams **MUST** upload all subtitles within 10 days from show airdate otherwise the team will loose the "trust" badge and another team will be the official team for that show (they fight about it)

IMO SR should keep searching for subs until the 10th day - providers uses cache to avoid hammering provider. then search once a week for one month. Then we lock the search. After that only with forced search
95% of all popular shows get subtitles in 24 hours. All subs from ltv goes to opensubtitles also.
web-dl releases get a little longer to be added as they take a little longer to be available - but 10th day limit is still valid)

The issue that @medariox @labrys and I was discussing is that we need to base the limit when the show was added to library, not based in the airdate. Otherwise old shows won't get subtitles.
What do you think?

miigotu:
I think it can still be based on the airdate, because if it is even more than 7 days in the past it still has to search for them 7 times, and like you said they only have 10 days to get it up.

however, I was thinking about extending it so it keeps searching once a week or so for old missed subs.

will have more input when I wake up, is like 7:30 am now trying to fall asleeeeep lol

miigotu:
From looking at it the sql select and the if dont mesh together the way it is intended. We need to rethink/rework the select anf the if and the "rules" and come up with a simpler more reliable way to do it.

miigotu:
rewriting the logic now, will add a commit later

fernandog:
@miigotu so if airdate diff < 10 it will search for subs based in subtitle search frequency setting?

top30 shows get subs really fast. Withing hours. TWD for example. 3-4am we already have subs.

miigotu:
@fernandog I added my commits. It doesnt bother with daydiff now, and the whole thing is much simpler. What do you think?

felipesalomao:
I think best solution is search for subtitles every hour until 10 days, if more than 10 days, could search 1 time each day, after 20 days, could search 2 times per week..  after 1 month, stop searching..  something like this.. There is no sense seach for subs each 8 hours after only 24 hours, most of subs are reseased 1-4 days after epsodes aired, just some top tv shows release less than 24 hours after epsodes aired..

miigotu:
This is already an extreme increase of searches on providers over what it was before this PR. We cannot increase it any more (it might be too much now already) because of the strain thousand of sickrage users searching for hundreds of episodes each DDoSing providers. Just one episode can be more than 20 api requests.

labrys:
@fernandog 200 / 24 hours is the limit on OpenSubtitles.org.  I think the frequency here is likely too high myself.

labrys:
@miigotu do we obey the limits when reached?  For example OpenSubs returns a  `407 Download limit reached`.

miigotu:
subliminal handles the api limits properly internally

felipesalomao:
Subtitle module isn't good with those limits right now.. I am getting subtitle search limit for all my TV shows.. Supergirl s01e08 was downloaded at 15/12 00:18, Subtitle was published on legendas.tv at 16/12 at 01:37 and subs only was downloaded at 16/12 12:06, almost 11 hours after.. And mine SR is configured to search it each 1 hour.. Legendas.tv don't have limit.. I hadn't this subs limit before.. 

And another strange thing is that it was downloaded but continue list it as pending on missed subs..
image
image

miigotu:
Except you arent even using a version of SR that includes this PR yet, because it is still in develop.

How do I know? That log line doesnt even exist after this PR.

sickbeard/subtitles.py
```diff
@@ -465,48 +461,54 @@ def run(self, force=False):  # pylint: disable=unused-argument,too-many-statemen
...
+                               % (ep_to_sub['show_name'], ep_to_sub['season'], ep_to_sub['episode']), logger.DEBUG)
+                    continue
 
-                    new_subtitles = frozenset(episode_object.subtitles).difference(existing_subtitles)
-                    if new_subtitles:
-                        logger.log(u'Downloaded %s subtitles for %s S%02dE%02d'
-                                   % (', '.join(new_subtitles), ep_to_sub['show_name'],
-                                      ep_to_sub["season"], ep_to_sub["episode"]))
-                else:
-                    logger.log(u"Subtitle search limit reached for %s S%02dE%02d"
```
miigotu, line 508: @felipesalomao see that log line was removed here.

PR #575, opened 2015-12-25
Fixed logging for calls to/from TVShow.wantEpisode

Amadeus-:
The wantEpisode method of the TVShow class currently returns false for several reasons; however, in some instances across SR, when the function returned false, a log entry was added indicating that it was because the quality was incorrect.

This commit adjusts wantEpisode so that it creates detailed INFO log messages on a false return.  This commit also removes the bogus log messages from the calling function.

So, previously, the log would have looked like this:  http://www.hastebin.com/ivebafuyoc.vbs       But, with this commit, the log file will look like this:  http://hastebin.com/vapapovodo.vbs       As you can see, the first one gives the quality as the reason for all of the false returns ...while the second log snippet gives the actual reason for each.

### DIFF
--- a/sickbeard/nzbSplitter.py
+++ b/sickbeard/nzbSplitter.py
@@ -205,9 +205,6 @@ def split_result(obj):
         want_ep = True
         for ep_num in parsed_obj.episode_numbers:
             if not obj.extraInfo[0].wantEpisode(season, ep_num, obj.quality):
-                # pylint: disable=no-member
-                logger.log(u"Ignoring result " + new_nzb + " because we don't want an episode that is " +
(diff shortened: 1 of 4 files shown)


SickRage2:

Thanks!

Amadeus-:
I know that the beauty of the having the log message in the calling function was that it provided the full name of the file; however, the only way to keep that particular feature would be to add another parameter/argument to wantEpisode() or have two messages in the log (which seemed spammy when I tried it.)    So, I figured this was the most elegant solution.

But, if you prefer to add a string parameter that can be passed to the wantEpisode() method that provides the name of the file, then that would make the new log match the old one, but with the correct reason.   Let me know if you prefer this and I can adjust it.

PR #577, opened 2015-12-26
Fixed Custom Scene Numbering

Amadeus-:
Custom Scene Numbering (as described here: https://github.com/SiCKRAGETV/sickrage-issues/wiki/Scene-exceptions-and-numbering) was broken.   This commit fixes the issues directly related to scene numbering.

### DIFF
--- a/sickbeard/scene_numbering.py
+++ b/sickbeard/scene_numbering.py
@@ -47,22 +47,25 @@ def get_scene_numbering(indexer_id, indexer, season, episode, fallback_to_xem=Tr
     :param fallback_to_xem: bool If set (the default), check xem for matches if there is no local scene numbering
     :return: (int, int) a tuple with (season, episode)
     """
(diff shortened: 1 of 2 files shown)


SickRage2:

Thanks!

Amadeus-:
The changes in scene_numbering.py (lines 50-68) looks like a bigger change than it is.   All I did was change the order of things.   All I did was have it check find_scene_numbering() first, then it still does the other checks thereafter.      The reasoning for this is because the custom scene numbering code doesn't set showObj.scene.   So, this solution gets the same results as before, but avoids extensive code changes elsewhere.

Amadeus-:
Hmmm ....Actually, do not merge this yet.  It works as it should; however, when SR snatches the file (based on the custom scene numbering) it's creating an entry for it in the wrong place.   

Let me see if I can fix that bug and then I'll update.

PR #578, opened 2015-12-26
Fixed logging for calls to TVShow.wantEpisode

Amadeus-:
The wantEpisode method of the TVShow class currently returns false for several reasons; however, in some instances across SR, when the function returned false, a log entry was added indicating that it was because the quality was incorrect.

This commit adjusts wantEpisode so that it creates detailed INFO log messages on a false return. This commit also removes the bogus log messages from the calling function.

So, previously, the log would have looked like this: http://www.hastebin.com/ivebafuyoc.vbs But, with this commit, the log file will look like this: http://hastebin.com/vapapovodo.vbs As you can see, the first one gives the quality as the reason for all of the false returns ...while the second log snippet gives the actual reason for each.

### DIFF
--- a/sickbeard/nzbSplitter.py
+++ b/sickbeard/nzbSplitter.py
@@ -205,9 +205,6 @@ def split_result(obj):
         want_ep = True
         for ep_num in parsed_obj.episode_numbers:
             if not obj.extraInfo[0].wantEpisode(season, ep_num, obj.quality):
-                # pylint: disable=no-member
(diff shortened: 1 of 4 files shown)


SickRage2:

Thanks!

Amadeus-:
Note:  The new code in tv.py, lines 1238-1246 is from https://github.com/SickRage/SickRage/pull/577  ...it's directly related to the wantEpisode function (and a pretty heinous bug), so I included it here to make the merge easier.

PR #610, opened 2015-12-30
Move validation inside getOverview...

labrys:
and clean up TV.py a little bit.
- Moved validation inside `getOverview` to simplify usage.
- Added a debug entry if `getOverview` fails to get a valid overview
- Streamlined `TVShow` properties: `is_anime`, `is_sports`, `is_scene`
- Refactored any and best qualities in common.py to allowed and preferred for consistency

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1277,6 +1277,13 @@ class TVShow(object):
         return False
 
     def getOverview(self, epStatus):
+        """
+        Get the Overview status from the Episode status
+
+        :param epStatus: an Episode status
+        :return: an Overview status
+        """
+
         ep_status = try_int(epStatus) or UNKNOWN
 
         if ep_status == WANTED:
@@ -1296,7 +1303,7 @@ class TVShow(object):
         elif ep_status in Quality.SNATCHED_BEST:
             return Overview.SNATCHED_BEST
         elif ep_status in Quality.DOWNLOADED:
-            allowed_qualities, preferred_qualities = Quality.splitQuality(self.quality)  # @UnusedVariable
+            allowed_qualities, preferred_qualities = Quality.splitQuality(self.quality)
             ep_status, cur_quality = Quality.splitCompositeStatus(ep_status)
 
             if cur_quality not in allowed_qualities + preferred_qualities:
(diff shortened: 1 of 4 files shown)


SickRage2:

Thanks!

sickbeard/webapi.py
```diff
@@ -1159,7 +1159,7 @@ def run(self):
 
             for curResult in sql_results:
 
-                cur_ep_cat = curShow.getOverview(int(curResult["status"] or -1))
+                cur_ep_cat = curShow.getOverview(curResult["status"])
```
fernandog: If status is -1,   INT will fail. We had tons of that issue in past

sickbeard/tv.py
```diff
@@ -149,24 +149,15 @@ def __init__(self, indexer, indexerid, lang=""):
 
     @property
     def is_anime(self):
-        if int(self.anime) > 0:
-            return True
-        else:
-            return False
+        return True if int(self.anime) > 0 else False
```
MGaetan89, line 152: `return int(self.anime) > 0`? Same for the two below.

sickbeard/tv.py
```diff
@@ -1277,26 +1268,34 @@ def wantEpisode(self, season, episode, quality, manualSearch=False, downCurQuali
...
     def getOverview(self, epStatus):
+        """
+        Get the Overview status from the Episode status
+
+        :param epStatus: an Episode status
+        :return: an Overview status
+        """
 
-        if epStatus == WANTED:
+        ep_status = try_int(epStatus) or UNKNOWN
```
MGaetan89, line 1278: `try_int(epStatus, UNKNOWN)`?

fernandog:
Unknown = -1 afaik

sickbeard/tv.py
labrys, line 1278: No because then it could return a 0 which we don't want

labrys:
Correct UNKNOWN is -1

sickbeard/tv.py
labrys, line 152: Was going to do that but I was rushing out the door, this was more benign and didn't require thought :).  Obviously they are equivalent tho and should be changed to the more concise form

PR #630, opened 2015-12-31
fix multithreading issues related to row_factory in DBConnection

coderbone:
db.py: DBConnection

```
manipulation of row_factory on sql connection was not done in a thread safe manner

factory methods should be static to avoid holding a reference
```

I can provide additional supporting changes to demonstrate the problem, this PR contains only the fix

### DIFF
--- a/sickbeard/db.py
+++ b/sickbeard/db.py
@@ -59,29 +59,36 @@ class DBConnection(object):
                 db_locks[self.filename] = threading.Lock()
 
                 self.connection = sqlite3.connect(dbFilename(self.filename, self.suffix), 20, check_same_thread=False)
-                self.connection.text_factory = self._unicode_text_factory
+                self.connection.text_factory = DBConnection._unicode_text_factory
 
                 db_cons[self.filename] = self.connection
             else:
                 self.connection = db_cons[self.filename]
 
-            if self.row_type == "dict":
-                self.connection.row_factory = self._dict_factory
-            else:
-                self.connection.row_factory = sqlite3.Row
+            # start off row factory configured as before out of
+            # paranoia but wait to do so until other potential users
+            # of the shared connection are done using
+            # it... technically not required as row factory is reset
+            # in all the public methods after the lock has been
+            # aquired
+            with db_locks[self.filename]:
+                self._set_row_factory()
+
         except Exception as e:
             logger.log(u"DB error: " + ex(e), logger.ERROR)
             raise
 
-    def _execute(self, query, args):
-        try:
-            if not args:
-                return self.connection.cursor().execute(query)
-            return self.connection.cursor().execute(query, args)
-        except Exception:
-            raise
+    def _set_row_factory(self):
+        """
+        once lock is aquired we can configure the connection for
+        this particular instance of DBConnection
+        """
+        if self.row_type == "dict":
+            self.connection.row_factory = DBConnection._dict_factory
+        else:
+            self.connection.row_factory = sqlite3.Row
 
-    def execute(self, query, args=None, fetchall=False, fetchone=False):
+    def _execute(self, query, args=None, fetchall=False, fetchone=False):
         """
         Executes DB query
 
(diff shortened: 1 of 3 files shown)


SickRage2:

Thanks!

miigotu:
Will need examples of what this fixes, some supporting info and whatnot. Can't accept such a major core change without some solid demonstration so we can understand what is going on.

coderbone:
OK, I mentioned that in the comment but wasn't sure how to provide that
short of putting it in the PR.

Basically, there is a race condition.  The row_factory set on a _shared_
sql connection during the DBConnection creation... if that connection
doesn't finish all it's actions before a connection is created with a
different row_type the results will be of the wrong type.

coderbone:
just added some commits to the PR:

webapi.py... a new command called 'simload' that starts a thread which makes a bunch of queries using a DBConnection that uses the default row_factory... the intent is to reliably put SR into a mode where there are a bunch of queries happening in the same was as the various threads do.

tests/db_row_type_problem.py... start the SR server without my proposed changes to db.py then run this on a command line in a separate shell.... this script has two threads, one that keeps making requests to the simload command mentioned above, the other to the 'backlog' command which makes a sequence of queries with row_type == 'dict'

if the code had no bugs then the backlog command would only see dict objects and the simload command would never see any dict objects but due to the race condition the simload command sees dict objects frequently... the test script will log some stuff about dicts being seen and end with one False and one True being printed

with the proposed fixes to db.py the test script will just print True/True because each connection is only receiving the type of objects it's expecting

coderbone:
doh, in the test script I've revealed my webapikey... of course you'll need to update that file to reflect your local testing environment

miigotu:
Ive ran the test here with your changes I get:
(I changed curl for requests ofc)

``` bash
dusted@PC:~/Projects/SickRage: python tests/db_row_type_problem.py
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
False
False
```

Then I did `git checkout origin/develop sickbeard/db.py` and ran the test again after restarting SR and get the exact same result:

``` bash
dusted@PC:~/Projects/SickRage: python tests/db_row_type_problem.py
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
<Response [200]>
False
False
```

coderbone:
perhaps you have no backlog?  this relies on backlog queries actually returning data.  in my case there are about 1000 records returned by the backlog query (a topic for another day)

miigotu:
Changing to .content from requests yields this for both db.py:
```
{"data": {"n_dicts": 1456}, "message": "", "result": "success"}
{"data": {"n_dicts": 1522}, "message": "", "result": "success"}
{"data": {"n_dicts": 1580}, "message": "", "result": "success"}
{"data": {"n_dicts": 1647}, "message": "", "result": "success"}
{"data": {"n_dicts": 1705}, "message": "", "result": "success"}
{"data": {"n_dicts": 1771}, "message": "", "result": "success"}
{"data": {"n_dicts": 1833}, "message": "", "result": "success"}
{"data": {"n_dicts": 1895}, "message": "", "result": "success"}
{"data": {"n_dicts": 1957}, "message": "", "result": "success"}
{"data": {"n_dicts": 2022}, "message": "", "result": "success"}
False
True
```

coderbone:
well you've certainly reproduced the output as I see it... now why is it that my repaired db.py has not had an effect?  Obviously you have to restart the SR server with the new db.py as that's server side, not client side.

happy to talk through the db.py changes line by line... short of pep8 or other stylistic details it looks pretty solid

miigotu:
Ok after some tinkering I got it to True,True. PR seems pretty straight forward.

miigotu:
bugging me that we shouldnt have that test as it is right now in SR, but I think we should also do some proper unit test for this scenario.

Can you remove that testing commit and squash?

coderbone:
I can try... we've seen that I'm not exactly a whiz when it comes to manipulating git :)

miigotu:
I can do it from here, and make a new PR

coderbone:
how am I ever going to learn if you do it for me?

miigotu:
lol,

```
git fetch upstream
git rebase -i upstream/develop
(comment unwanted commits and commits that arent yours/part of the PR, change pick to fixup or squash or whatever you need, first commit should always be pick or reword, save+quit)
git push origin branchname -f
```

coderbone:
I will add that to my cheat sheet.... I'm lucky enough at my day job that the dev is highly siloed and branching/merging/rebasing is just something that never happens... I appreciate that this is vital for the management of a project like SR... learning this stuff is part of the reason to participate in SR and I'm grateful for the guidance along the way.

but right now I have a head full of Lithuanian pear brandy (happy new year) so who knows if any of this will stick

miigotu:
=P np

PR #696, opened 2016-01-08
Remove ! from the notes since it was removed for some reason Oo

duramato:
@SickRage/collaborators & @SickRage/moderators  
Should we add back the support for this option? Would be good for when a site is in maintance for example ,or like when TPB has issues ie)
OR remove it with this PR

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -61,7 +61,6 @@ $('#config-components').tabs();
 
                         <div>
                             <p class="note">* Provider does not support backlog searches at this time.</p>
-                            <p class="note">! Provider is <b>NOT WORKING</b>.</p>
                         </div>
                     </div>
 


SickRage2:

Thanks!

labrys:
@duramato I think we should eventually add back support for this, as well as possibly a "Test Provider" method similar to the notifications tests.

miigotu:
I think we should keep this and start using it.

duramato:
Ok ,the ! will need to be added back to the ui as it was removed sometime in the past

fernandog:
this is how sickbeard does:

`#if $curProvider.name == "EZRSS" then "**" else ""#`
https://github.com/midgetspy/Sick-Beard/blob/master/data/interfaces/default/config_providers.tmpl#L52-L74

Its a manual thing.

duramato:
It would be better to have somethinh dyn like we do with the provider color the padlock and the if it supports only daily

fernandog:
only shazbat don't have it. and @miigotu said he will add it. so don't need that.

but I would add an icon for "dailysearch enabled" and "backlogsearch enabled"

duramato:
Shazbat for the time beeing dont have and same goes for RSS feeds. For icon sounds good but needs to be subtle :p

PR #739, opened 2016-01-13
Fix conditional check

fernandog:
@miigotu

https://github.com/SickRage/sickrage-issues/issues/148

also here are all indexers category results for that user's issue:
https://gist.github.com/XxUnkn0wnxX/060ec28dde8af715554f

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -23,9 +23,10 @@
 
 import os
 import re
-import urllib
-import datetime
 import time
+import datetime
+import posixpath  # Must use posixpath
+from urllib import urlencode
 from bs4 import BeautifulSoup
 
 import sickbeard
@@ -39,9 +40,8 @@ from sickbeard.common import Quality
 from sickrage.helper.encoding import ek, ss
 from sickrage.show.Show import Show
 from sickrage.helper.common import try_int, convert_size
-# from sickbeard.common import USER_AGENT
 from sickrage.providers.nzb.NZBProvider import NZBProvider
-from sickbeard.common  import cpu_presets
+from sickbeard.common import cpu_presets
 
 
 class NewznabProvider(NZBProvider):
@@ -54,8 +54,6 @@ class NewznabProvider(NZBProvider):
                  search_fallback=False, enable_daily=True, enable_backlog=False):
 
         NZBProvider.__init__(self, name)
-
-        # self.headers.update({'User-Agent': USER_AGENT})
 
         self.urls = {'base_url': url}
         self.url = self.urls['base_url']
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

sickbeard/providers/newznab.py
```diff
@@ -156,7 +156,7 @@ def get_newznab_categories(self):
             return False, return_categories, error_string
 
         data = BeautifulSoup(data, 'html5lib')
-        if not self._checkAuthFromData(data) and data.caps and data.caps.categories:
+        if not self._checkAuthFromData(data) and not data.caps and not data.caps.categories:
```
MGaetan89: If `not data.caps`, then `not data.caps.categories` is always true, no? So it could simply be `if not self._checkAuthFromData(data) and not data.caps:`.

fernandog: not sure why @miigotu added both checks

fernandog:
@miigotu

https://gist.github.com/XxUnkn0wnxX/5eba2cb8be370a0f44f1

```
  File "/Users/XxUnkn0wnxX/Apps/SickRage/sickbeard/providers/newznab.py", line 159, in get_newznab_categories
    if not self._checkAuthFromData(data) and not data.caps and not data.caps.categories:
AttributeError: 'NoneType' object has no attribute 'categories'
 [e31a2e2]
```

sickbeard/providers/newznab.py
miigotu: `if not (self._checkAuthFromData(data) and data.caps and data.caps.categories):`

medariox:
@fernandog 
If you really need to check the existence of an attribute of an attribute, you need to do it like this:

```
if not self._checkAuthFromData(data) and not data.caps:
    if not data.caps.categories:
....
```

Assuming data always exists.

But as @MGaetan89 already said, with a negation you don't need to do this. if data.caps doesn't exist, data.caps.categories can't exist either. Without the negation it would make sense tho.

fernandog:
http://puu.sh/muexU.png
"Missing parameter" with https://api.nzb.su/api?t=caps&apikey=***

fernandog:
Strange that in logs `2016-01-14 02:19:18 INFO     Thread-32 :: Api Disabled`

are we using the correct URL ??
https://www.reddit.com/r/usenet/comments/24rhj3/dognzb_api_disabled/

api.nzb.su instead of nzb.su ?

http://pastebin.com/dWxUKp1F

duramato:
The problem with nzb.su i think they are only responding to hattps requests afaik

sickbeard/providers/newznab.py
medariox: @miigotu 
Would return an AttributeError if data.caps doesn't exist

miigotu: no, it wouldnt. when bs4 subelement doesnt exist it is None. The attribute error is when caps is None and you try to access a subelement of that, because the caps doesnt exist.

```
>>> d = soup('<error>blah</error>', 'html5lib')
>>> d.caps
>>> d.caps.categories
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'categories'
```

miigotu:
You cant use os.path.join for urls, may have been the problem here. On windows this could have joined it with `\`

PR #765, opened 2016-01-17
Fix "Size" Bug on main/home Show List display

Amadeus-:
There is currently a bug now in which the the "Size" column on the main/home Show List display includes disk file usage for any episodes that are marked as 'archived', 'downloaded',etc..., regardless of whether they actually exist on the disk or not. 

(In other words, if you delete the files manually, the "Size" column would not reflect those changes.)

The method I used here is the only one that I could come up with that would fix this bug without needing to make structural changes to how information on episodes are stored in the database.   I kept things as simple as possible, by iterating through the episodes for the particular show and checking if the file_size is > 0 BEFORE checking if the file actually exists.

Perhaps someone with hundreds/thousands of episodes could test to ensure that this methodology doesn't create lag; however, I think it should be OK.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -721,8 +721,7 @@ class Home(WebRoot):
         sql_statement += ' OR (status IN ' + status_quality + ') OR (status IN ' + status_download + '))) AS ep_total, '
 
         sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate >= ' + today + ' AND (status = ' + str(UNAIRED) + ' OR status = ' + str(WANTED) + ') ORDER BY airdate ASC LIMIT 1) AS ep_airs_next, '
-        sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate > 1 AND status <> ' + str(UNAIRED) + ' ORDER BY airdate DESC LIMIT 1) AS ep_airs_prev, '
-        sql_statement += ' (SELECT SUM(file_size) FROM tv_episodes WHERE showid=tv_eps.showid) AS show_size'
+        sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate > 1 AND status <> ' + str(UNAIRED) + ' ORDER BY airdate DESC LIMIT 1) AS ep_airs_prev '
         sql_statement += ' FROM tv_episodes tv_eps GROUP BY showid'
 
         sql_result = myDB.select(sql_statement)
@@ -730,9 +729,16 @@ class Home(WebRoot):
         show_stat = {}
         max_download_count = 1000
         for cur_result in sql_result:
-            show_stat[cur_result['showid']] = cur_result
+            show_stat[cur_result['showid']] = dict(cur_result)
+            show_stat[cur_result['showid']]['show_size'] = 0
             if cur_result['ep_total'] > max_download_count:
                 max_download_count = cur_result['ep_total']
+                
+            sql_statement = 'SELECT location, file_size FROM tv_episodes WHERE showid='+str(cur_result['showid'])+' AND file_size > 0 AND season > 0 AND episode > 0 AND airdate > 1 AND status IN ' + status_download
+            sql_size_results = myDB.select(sql_statement)
+            for sql_size_result in sql_size_results:
+                if os.path.isfile(sql_size_result['location']):
+                    show_stat[cur_result['showid']]['show_size'] += sql_size_result['file_size']
 
         max_download_count *= 100
 


SickRage2:

Thanks!

Amadeus-:
One thing that I didn't change, but would probably make everything much better (and would be very easy to add), is code to "zero out" the file_size field of an episode in the tv_episodes table IF the file (given in the 'location' field) doesn't exist anymore.   Know what I'm saying?

I can't imagine why this wouldn't be desirable; however, I wanted to make sure before including it in the PR.   But, if it were added, it would make the whole routine MUCH more efficient for those users with significant number of episodes that were manually deleted at some point in the past.

sickbeard/webserve.py
```diff
@@ -721,8 +721,7 @@ def show_statistics():
         sql_statement += ' OR (status IN ' + status_quality + ') OR (status IN ' + status_download + '))) AS ep_total, '
 
         sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate >= ' + today + ' AND (status = ' + str(UNAIRED) + ' OR status = ' + str(WANTED) + ') ORDER BY airdate ASC LIMIT 1) AS ep_airs_next, '
-        sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate > 1 AND status <> ' + str(UNAIRED) + ' ORDER BY airdate DESC LIMIT 1) AS ep_airs_prev, '
-        sql_statement += ' (SELECT SUM(file_size) FROM tv_episodes WHERE showid=tv_eps.showid) AS show_size'
+        sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate > 1 AND status <> ' + str(UNAIRED) + ' ORDER BY airdate DESC LIMIT 1) AS ep_airs_prev '
```
Amadeus-: Note:  Only change here is a removal of that final comma (for syntax)

fernandog:
@Amadeus- archived status can still have file. you shouldn't ignore it. what you should do is check if location is empty. If so, not count size. Dont use status.

Also when you refresh show, SR erase the "size" value in DB when it detects file was removed.

Amadeus-:
@fernandog   this PR does not ignore archived files.   The diff doesn't show it, but earlier in the function:

```
status_download = '(' + ','.join([str(x) for x in Quality.DOWNLOADED + Quality.ARCHIVED]) + ')'
```

So, my routine does catch all 'downloaded' and 'archived' files.  (I have archived files myself ..both actual and deleted ...so was able to test)

The second thing you said is great, that makes my suggestion for more changes unnecessary.   

You mentioned that the 'size' value is erased when the show is refreshed; however, it's not happening as often as it should because I'm seeing incorrect file sizes a lot (I typically delete files as soon as I'm done watching them.)

So, if what you said is good (i.e., that the file_size field is zeroed-out at some point when SR detects a file is missing) then it sounds like this PR should fix the bug without any additional work.

Anyway, I think if you test this PR it should be obvious pretty quickly if it's good/bad/etc.   It works great for me!

sickbeard/webserve.py
```diff
@@ -721,8 +721,7 @@ def show_statistics():
         sql_statement += ' OR (status IN ' + status_quality + ') OR (status IN ' + status_download + '))) AS ep_total, '
 
         sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate >= ' + today + ' AND (status = ' + str(UNAIRED) + ' OR status = ' + str(WANTED) + ') ORDER BY airdate ASC LIMIT 1) AS ep_airs_next, '
-        sql_statement += ' (SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate > 1 AND status <> ' + str(UNAIRED) + ' ORDER BY airdate DESC LIMIT 1) AS ep_airs_prev, '
-        sql_statement += ' (SELECT SUM(file_size) FROM tv_episodes WHERE showid=tv_eps.showid) AS show_size'
```
miigotu, line 725: `' (SELECT SUM(file_size) FROM tv_episodes WHERE showid=tv_eps.showid AND location != '') AS show_size'`

Amadeus-, line 725: Well, assuming that what @fernandog said is correct, then this alteration wouldn't do anything but make the already existing routine more efficient.   (I'm assuming that the location is set to '' and the file_size are set to 0 at the same time.)

In other words, the 'location' string still exists for episodes well after files have been manually deleted.   I didn't look through the code to see how often shows are "updated"; however, it's certainly not so often that I didn't notice this bug several times.

I suppose the question becomes, "are you OK with having the information on the page be incorrect until a show is refreshed?"   Perhaps it's a design flaw that's not worth addressing; however, I just hated having people see data that was incorrect when it was possible to fix it.

fernandog, line 725: Updates/refreshes everyday at 3am

miigotu, line 725: If you are deleting an episode manually/outside of SR knowldge, you need to refresh the show manually. A second SQL query is unacceptable.

Amadeus-, line 725: It's your call.  But, I imagine a lot of people do it and have no idea that the design of the program is such that requires a manual "re-scan files" (it's not intuitive.)     And, I think that 're-scan files' is the only way to mark episodes as deleted...as far as I can tell.    Is there a way to refresh all shows?

But I don't mind that much either way.   I just figured you'd want it fixed.   I still think it's a bug caused by a design limitation (an end-user would expect modern software to keep track of what's on the disk without needing a manual 're-scan') . ..but, I don't feel strongly enough about it to have a major debate ...lol.    So, yea, if you want to leave it as it is, feel free to go and and close this PR. :)

miigotu, line 725: How can you delete files and not tell SR that you deleted something and expect it to randomly know instantly?

OmgImAlexis, line 725: @miigotu doesn't SickRage use a watch handler like [watchdog](https://pypi.python.org/pypi/watchdog) to check for updated on directories?

miigotu, line 725: No, because drives wouldnt sleep then,

fernandog:
@Amadeus- here is the code: 
https://github.com/SickRage/SickRage/blob/develop/sickbeard/tv.py#L1094

I see it's not erasing size column BUT checking my DB, all files without location have size = 0

do you have `sickbeard.SKIP_REMOVED_FILES` enabled or disabled?
do you keep your SR running 24/7 ?

PR #806, opened 2016-01-22
Improved scene number correction

p0psicles:
Scene numbers could be corrected through the setEpisodeSceneNumbering() ajax request.
First this you need to change the value in the Scene column. But you need to use the format [season]x[episode].

I've added a regex that assumes you intent to set season 1 when only one number is entered. For ex. 5 will result in 1x5.


SickRage2:

Thanks!

p0psicles:
see https://github.com/SickRage/sickrage-issues/issues/117

p0psicles:
The only caveat here is that when you enter '5' for episode 4 in season 5. Is that it then will configure 1x5. I noticed that later with help of  @fernandog. We discussed this, and agreed that it's better to not allow entering a single digit. But give users some feedback in the form of colors (green,red) for entering valid scene episode numbers. 0x0 wil result to default. Maybe even something like a tooltip?

I don't think this PR can do much harm. But we must not forget this.

miigotu:
Just PR a fix for that ASAP, I cant merge to master until this regression is resolved

PR #815, opened 2016-01-22
Quick'n'dirty, return true rather than false when PP finds `exists sa…

miigotu:
…me size` so it is added to the DB as processed.

Fixes https://github.com/SickRage/sickrage-issues/issues/483

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -516,7 +516,7 @@ class PostProcessor(object):
             to_return = (show, season, [], quality, version)
 
             qual_str = common.Quality.qualityStrings[quality] if quality is not None else quality
-            self._log("Found result in history for %s - Season: %s - Quality: %s - Version: %s" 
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

OmgImAlexis:
@miigotu if you run manual PP this will still process, correct?

miigotu:
Yes, this doesnt change anything other than removing the error. This used to return true way back and is the only reason you see this error now. It happens when the history gets trimmed on SR startup, it can forget that it has processed this file before and try it again. When it returns true it just adds it back to the cache. In this case, I think thats sort of db thrashing that we trim the db and then add the same data back on PP. Thoughts @labrys?

labrys:
@miigotu I think the log line is terrible :grinning: but I think your logic is sound.

miigotu:
@labrys it can be changed, I was just being a smart ass because I think its a bad solution =P. This is how it originally was though and we didnt have issue reports the lol.

PR #840, opened 2016-01-25
Added try-except for subtitles with non-supported chars and guessit e…

medariox:
…rror

This avoids crashing with these two errors:
- https://github.com/SickRage/sickrage-issues/issues/630
- https://github.com/SickRage/sickrage-issues/issues/607#issuecomment-174157750

First error is also reported upstream here: https://github.com/guessit-io/guessit/issues/260

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -144,8 +144,14 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
                     episode_num(subtitles_info['season'], subtitles_info['episode'], numbering='absolute')), logger.DEBUG)
         return existing_subtitles, None
 
-    subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
-    video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
+    try:
+        subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
+        video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
+    except UnicodeEncodeError as error:
+        logger.log(u'An error occurred while encoding \'{}\' with your current locale. '
+                   'Rename the file or try a different locale. Error: {}'.format
+                   (subtitles_info['location'], ex(error)), logger.WARNING)
+        return existing_subtitles, None
     user_score = 132 if sickbeard.SUBTITLES_PERFECT_MATCH else 111
 
     video = get_video(video_path, subtitles_path=subtitles_path)
@@ -174,7 +180,12 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
             return existing_subtitles, None
 
         for subtitle in subtitles_list:
-            matches = subtitle.get_matches(video, hearing_impaired=sickbeard.SUBTITLES_HEARING_IMPAIRED)
+            try:
+                matches = subtitle.get_matches(video, hearing_impaired=sickbeard.SUBTITLES_HEARING_IMPAIRED)
+            except ValueError as error:
+                logger.log(u'An error occurred while getting a subtitle match for: {}. Error: {}'.format
+                           (video.name, ex(error)), logger.WARNING)
+                continue
             score = subliminal.subtitle.compute_score(matches, video)
             logger.log(u'[{}] Subtitle score for {} is: {} (min={})'.format
                        (subtitle.provider_name, subtitle.id, score, user_score), logger.DEBUG)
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

medariox:
Not sure about the log messages here. Better ideas anyone?

duramato:
They sound good to me

sickbeard/subtitles.py
```diff
@@ -144,8 +144,14 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
                     episode_num(subtitles_info['season'], subtitles_info['episode'], numbering='absolute')), logger.DEBUG)
         return existing_subtitles, None
 
-    subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
-    video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
+    try:
+        subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
```
miigotu, line 148: Should this have been:
`subtitles_path = get_subtitles_path(subtitles_info['location'].encode(sickbeard.SYS_ENCODING))`
instead of:
`subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)`

? Just curious if encoding the result instead of the param may have been the issue.

medariox, line 148: Hmm, don't think so. The original error was at:

```
video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
```

```
2016-01-23 22:06:57 DEBUG    FINDSUBTITLES :: Checking subtitle candidates for Heavy Object S01E13 (Heavy Object - S01E13 - In a Calvary Battle, You Must Break Their Foothold ～Amazon City All-Out War I～ - 1080p HDTV.mkv)
2016-01-23 22:06:57 ERROR    FINDSUBTITLES :: Unable to find subtitles for Heavy Object S01E13. Error: charmap : E:\Animes\Heavy Object\Season 01\Heavy Object - S01E13 - In a Calvary Battle, You Must Break Their Foothold ï½žAmazon City All-Out War Iï½ž - 1080p HDTV.mkv : error 108 : error 109 : character maps to <undefined> [f67f6e9]
  File "J:\SickRage\sickbeard\subtitles.py", line 490, in run
    new_subtitles = episode_object.download_subtitles()
  File "J:\SickRage\sickbeard\tv.py", line 1433, in download_subtitles
    self.subtitles, new_subtitles = subtitles.download_subtitles(subtitles_info)
  File "J:\SickRage\sickbeard\subtitles.py", line 148, in download_subtitles
    video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
  File "C:\Python27\lib\encodings\cp1252.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_table)
UnicodeEncodeError: 'charmap' codec can't encode character u'\uff5e' in position 108: character maps to <undefined>
```

I think that it is caused by REALLY weird characters like `～` (!= ~), that can't be even encoded to UTF-8. Should be very rare (think we only got one report so far), but we might still want to catch that kind of exception.

duramato, line 148: @medariox ç cant be encoded as utf-8 afaik only on latin-1

medariox, line 148: @duramato 
ç shouldn't be a problem for UTF-8, actually u'\uff5e' `～` shouldn't be either (just tested). So most probably the user had the locale set as `cp1252` and that doesn't have `～`. Still think that it might be a helpful warning message anyway.

medariox:
Example of what I mean:

```
 python -c "print(u'\uff5e'.encode('cp1252'))"
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/encodings/cp1252.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_table)
UnicodeEncodeError: 'charmap' codec can't encode character u'\uff5e' in position 0: character maps to <undefined>
```

```
python -c "print(u'\uff5e'.encode('utf-8'))"
～
```

User has locale set to `cp1252` and that's the reason why `u'\uff5e'` can't be encoded. Warning message tells user to use different locale e.g. UTF-8.

We could force encoding to UTF-8 of course instead of sickbeard.SYS_ENCODING, but do we want this?

labrys:
@medariox screwing around with encoding/decoding is a dangerous prospect.  In fact we shouldn't be using `setdefaultencoding` either but we do.  `setdefaultencoding` and `ek` are probably the largest contributors to our mojibake issues and codec errors.

medariox:
@labrys 
No kidding! :confused: We all know encoding is a mess in SR right now with ek() and ss(). It was planned to get rid of them since ages. But what has this to do with the PR? The question here is if we should use the user-set locale to do the encoding (and thus catch the exceptions that can happen, see my example) or rather force to UTF-8 encoding and possibly(?) avoid this kind of issue, but this might have other unknown repercussions.

labrys:
@medariox Forcing `utf-8` can lead to unexpected errors and can't be assumed to be the correct answer.  For example, the reason your example code works is the object being encoded to a `utf-8` bytestring is already unicode which can be converted to `utf-8`.  One problem is what if the string being encoded is not already unicode?  You would need to decode it to unicode first and what codec should be used to decode it?  Also what if the target that you are passing the data to (e.g. console or whatever) can't handle `utf-8` encoding? e.g. the Windows console defaults to cp1252, trying to write `utf-8` to it is problematic.  In addition trying to force windows to the correct code page (cp65001 being the closest) doesn't even work on Windows (it's a python bug).

medariox:
@labrys 
Okay, as I thought there is no way around this try-except for now. Until we have a proper encoding implementation we will have to catch these errors. I think the PR can be merged like it is if there are no suggestions about the warning messages.

sickbeard/subtitles.py
miigotu, line 148: Maybe it should be .encode('utf-8') since its what we expect internally anyways.

miigotu:
I wil merge this, but I still think we should encode to utf8 instead of sys_encoding. the location is already unicode, and with the hackjob of setDefaultEncoding we expect utf8 almost everywhere else.

PR #854, opened 2016-01-28
Fixed TPB RSS search, added logging to some providers

medariox:


SickRage2:

Thanks!

sickbeard/providers/thepiratebay.py
```diff
@@ -47,7 +47,7 @@ def __init__(self):
         self.url = 'https://thepiratebay.se/'
         self.urls = {
             'search': self.url + 's/',
-            'rss': self.url + 'tv/latest'
+            'rss': self.url + 'browse/205'
```
miigotu, line 50: This needs to be 200, this limits it to SD TVShows, 208 is HD TVShows, 200 is both.

medariox, line 50: But 200 includes "HD - Movies"?

medariox, line 50: I still think 200 is a better solution then no HD. So if there is no better solution I would go for browse/200.

miigotu, line 50: 200 is all of the video category, but without it you dont get anything hd.
/tv/latest is just bugged right now, Im sure it will be fixed eventually

medariox, line 50: Ok, I would say let's use 200 for now and we can switch back to /tv/latest when it's fixed, as TPB is useless otherwise. Not really convinced they will fix it anytime soon, as they have kind of a troubled situation going on right now.

miigotu, line 50: sorta feels like they broke it on purpose, but yeah xD

PR #855, opened 2016-01-28
Build in same checks for when an anidb aid can't be mapped to a tvdbid

p0psicles:
Possibly due to github giving an error page when an anime-list.xml file downloaded.

### DIFF
--- a/lib/adba/aniDBfileInfo.py
+++ b/lib/adba/aniDBfileInfo.py
@@ -87,6 +87,9 @@ def download_file(url, filename):
     except Exception:
         _remove_file_failed(filename)
         return False
+    except requests.exceptions.RequestException, e:
+        _remove_file_failed(filename)
+        return False
 
     return True
 
--- a/lib/anidbhttp/model.py
+++ b/lib/anidbhttp/model.py
@@ -4,9 +4,11 @@ from sickrage.helper.encoding import ek
 import sickbeard
 from sickbeard import helpers
 import requests
+import logging
 
 __all__ = ["Anime", "Category", "Title", "Episode", "Tag"]
-
+log = logging.getLogger(__name__)
+log.addHandler(logging.NullHandler())
 
 class Entity(object):
     def __init__(self, id):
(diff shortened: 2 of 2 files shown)


SickRage2:

Thanks!

lib/adba/aniDBfileInfo.py
```diff
@@ -87,6 +87,9 @@ def download_file(url, filename):
     except Exception:
         _remove_file_failed(filename)
         return False
+    except requests.exceptions.RequestException, e:
```
miigotu, line 90: this is unnecessary, its already catching the more general Exception above.

lib/anidbhttp/model.py
```diff
@@ -184,7 +186,11 @@ def set_tvdbid(self):
         No params required.
         """
         from adba.aniDBtvDBmaper import TvDBMap
-        self._tvdbid = TvDBMap().get_tvdb_for_anidb(self.id) if self.id else None
+        try:
+            self._tvdbid = TvDBMap().get_tvdb_for_anidb(self.id) if self.id else None
+        except:
```
miigotu, line 191: need to specify and exception class

lib/adba/aniDBfileInfo.py
labrys, line 90: And the majority of those exceptions covered by that are already caught explicitly above.  So either generalize them to the `requests.exceptions.RequestException` or just add the missing `TooManyRedirects` if you feel its necessary to catch them all individually. (hint: it shouldn't be)

labrys, line 90: Also, wrong format for exceptions.  should be:

``` python
try:
    # code to try
except Exception as e:
    # exception handling
```

and if you aren't using the message then there's no reason to capture it as `e` either.  And try to avoid catchall exceptions like `Exception`

p0psicles, line 90: @miigotu if it's already catching the more general exceptions, why didn't it catch the exception last night? It shouldn't have saved the result in anime-lists.xml then?!

PR #860, opened 2016-01-28
Removed sickbeard imports from lib/anidbhttp/model.py

p0psicles:
As that does not belong there.
But for this I had to rewrite the "add from list" / recommendations.
- Removed the import sickbeard from the anidbhttp client and all code that was using that.
- Created a new dir /sickrage/show/recommendations/.
- Moved over imdbPopular.py and renamed it to imdb.py
- Created new class recommended, as a base class for recommendation shows. This will make sure we can move towards one mako template in future.
- Created new module anidb.py, with the code to retrieve the anime shows and map it to the RecommendedShow() class.
- Adjusted the addShows_animePopular.mako to make use of the RecommendedShow() objects.
- Regression tested imdb watchlist. But as this is not using the RecommendedShow class (yet), nothing much has changed.
- Took into account posixpath.join and removed the import from lib...

The other commit is for the issue when github is down, and responses with an 400 page, and that one is downloaded instead of the anime-list.xml.


SickRage2:

Thanks!

PR #862, opened 2016-01-28
Replacing help & info icons with colored versions.

neoatomic:


SickRage2:

Thanks!

neoatomic:
Did add some more icons to the page.

Examples. :

gui/slick/views/config.mako
```diff
@@ -14,9 +14,9 @@
 % endif
 
 <div id="config-content">
-<table class="infoTable" cellspacing="1" border="0" cellpadding="0" width="100%">
+<table class="infoTable" cellspacing="1" border="0" cellpadding="0" width="145%">
```
OmgImAlexis, line 17: This shouldn't be changed to more than `100%` instead the class should have the width changed or the parent div should have it's width changed.

miigotu, line 17: Damn I missed that lol

labrys:
Also the arguments line should wrap.

OmgImAlexis:
@labrys shouldn't the argument line just use `<pre>` tags?

miigotu:
That goes beyond the scope of this PR imho, the whole intention was just to add the different icons.

MGaetan89:
Why not one sprite for all icons?
Maybe vertical center text with icons too? But that's debatable.

neoatomic:
@OmgImAlexis How can i change the width correctly so the text stays on one line.? Could you post an example that i can test/play around with.?

OmgImAlexis:
@neoatomic that main problem you'll have is screen width. It shouldn't matter if it's on one or multiple lines since depending on the amount of text and the screen width it could either be off the screen or be visible.

labrys:
@OmgImAlexis Argument line would be fine to use `<pre>` tags, however then the css would have to handle the wrapping.  The argument line can potentially be really long, so wrapping should be enabled.  Also as @miigotu mentioned that's not really a part of this PR though, it was about the colored icons.

OmgImAlexis:
@labrys just thought I'd mention it since the width got changed on here.

PR #877, opened 2016-01-29
Added fallback to UTF-8 for UnicodeEncodeError, removed ValueError try-except (fixed with guessit2)

medariox:
@miigotu @labrys 

What do you think about the fallback solution?

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -265,17 +265,7 @@ def refresh_subtitles(episode_info, existing_subtitles):
 
 def get_video(video_path, subtitles_path=None):
     if not subtitles_path:
-        try:
-            subtitles_path = get_subtitles_path(video_path).encode(sickbeard.SYS_ENCODING)
-        except UnicodeEncodeError:
-            # Fallback to UTF-8
-            try:
-                subtitles_path = get_subtitles_path(video_path).encode('utf-8')
-            except UnicodeEncodeError as error:
-                logger.log(u'An error occurred while encoding \'{}\' with your current locale. '
-                           'Rename the file or try a different locale. Error: {}'.format
-                           (video_path, ex(error)), logger.WARNING)
-                return None
+        subtitles_path = get_subtitles_path(video_path)
 
     try:
         if not sickbeard.EMBEDDED_SUBTITLES_ALL and video_path.endswith('.mkv'):
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -147,11 +147,16 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
     try:
         subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
         video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
-    except UnicodeEncodeError as error:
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

sickbeard/subtitles.py
```diff
@@ -147,11 +147,16 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
     try:
         subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
         video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
-    except UnicodeEncodeError as error:
-        logger.log(u'An error occurred while encoding \'{}\' with your current locale. '
-                   'Rename the file or try a different locale. Error: {}'.format
-                   (subtitles_info['location'], ex(error)), logger.WARNING)
-        return existing_subtitles, None
```
fernandog: @medariox keep this `return existing_subtitles, None`

sickbeard/subtitles.py
```diff
@@ -147,11 +147,16 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
     try:
         subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
```
miigotu, line 148: I think the try/except is good, but I still think you are encoding the wrong string.
In several places you are encoding the result that subliminal returns and I think you should be encoding the parameter/path you are passing INTO subliminal. AKA, encode `subtitles_info['location']` not the result of `get_subtitles_path`
`subtitles_path = get_subtitles_path(subtitles_info['location'].encode(sickbeard.SYS_ENCODING))`
This follows the sandwitch format, when you encode when a string leaves your application and decode when it comes in. (entering a library is leaving your application, same as accessing filesystem, sending over network, etc.)
If that causes issues then it should be encoded somewhere else also.

@labrys ?

medariox, line 148: `get_subtitles_path` is not part of subliminal, it's just there to determine whether the user is using a relative, absolute or empty subtitle path. But I get what you mean and I'm not entirely sure why I encode it that way, but I do vaguely remember that it had a reason (e.g. encoding the result caused less issues). We can surely try to encode the subtitle location before passing it to get_subtitles_path and see how that goes. I'm also totally open to other suggestions. I think it would be great to talk this out to achieve a solid solution.

labrys, line 148: @miigotu Generally that is correct, you should in most cases decode to unicode (i know that sounds funny) when it enters your program and encode to a target encoding when it leaves.  

Generally for nearly all web applications `utf-8` is a safe encoding option since it is the default for most if not all internet communication.  However, problems arise when you decode/encode the wrong coding (e.g. trying to decode `utf-8` as `ascii`) or if the target does not support the unicode character (e.g. the output needs to be encoded as `ascii` but the data contains a unicode character).  Problems can also occur if you use the wrong form of the encoding.   For example `UTF-16BE` vs `UTF-16LE`.  Windows by default uses `UTF-16` with Little-Endian byte order, so if you used `UTF-16BE` instead you would have issues.

medariox:
Will be working on getting rid of ek() here: https://github.com/SickRage/SickRage/tree/subs-no-ek

PR #884, opened 2016-01-29
Update UI statistics labels

labrys:
- Rename Low Quality => Allowed
- Rename Downloaded => Preferred
- Fix backlog mako timezone error


SickRage2:

Thanks!

fernandog:
@labrys 

confusing! Its better the other way. Whats wrong with "Low quality" ?

image

labrys:
@fernandog Problem with `Low Quality` is since we no longer download progressively higher qualities it is no longer an accurate name.  As for your screenshot, there are still more that needs to be tweaked to avoid discrepancies like that, however I think this is a step in the right direction.  Ive tried out different wordings and `Allowed` / `Preferred` was the most concise while still being fairly representative.  I also considered `Interim` / `Final` but figured I'd try this out first since its similar to settings that already exist so should cause less confusion once the other details are ironed out.

OmgImAlexis:
Preferred and Allowed make sense if you look at it this way.
I'll **allow** SickRage to download any of these but I'd **prefer** it replace with one from the preferred list.

`Interim / Final` would work if everyone using it knew what they meant and I doubt a lot of people would understand off the bat so it's best to just keep it as @labrys said.

PR #891, opened 2016-01-30
Use unicode everywhere in subtitles

medariox:
This removes the use of ek() for subtitles.

The only parts where non-unicode is actually present, are the user inputted strings. In subtitles sickbeard.TV_DOWNLOAD_DIR and sickbeard.SUBTITLES_DIR, so these are the only two places where a conversion to unicode will be necessary. sickbeard.SYS_ENCODING will be tried first and utf-8 as fallback. Needs good testing and feedback is appreciated, as always.

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -144,14 +144,8 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
                     episode_num(subtitles_info['season'], subtitles_info['episode'], numbering='absolute')), logger.DEBUG)
         return existing_subtitles, None
 
-    try:
-        subtitles_path = get_subtitles_path(subtitles_info['location']).encode(sickbeard.SYS_ENCODING)
-        video_path = subtitles_info['location'].encode(sickbeard.SYS_ENCODING)
-    except UnicodeEncodeError as error:
-        logger.log(u'An error occurred while encoding \'{}\' with your current locale. '
-                   'Rename the file or try a different locale. Error: {}'.format
-                   (subtitles_info['location'], ex(error)), logger.WARNING)
-        return existing_subtitles, None
+    subtitles_path = get_subtitles_path(subtitles_info['location'])
+    video_path = subtitles_info['location']
     user_score = 132 if sickbeard.SUBTITLES_PERFECT_MATCH else 111
 
     video = get_video(video_path, subtitles_path=subtitles_path)
@@ -211,7 +205,7 @@ def download_subtitles(subtitles_info):  # pylint: disable=too-many-locals, too-
                                                               None if not sickbeard.SUBTITLES_MULTI else
                                                               subtitle.language)
         if subtitles_path is not None:
-            subtitle_path = ek(os.path.join, subtitles_path, ek(os.path.split, subtitle_path)[1])
+            subtitle_path = os.path.join(subtitles_path, os.path.split(subtitle_path)[1])
 
         sickbeard.helpers.chmodAsParent(subtitle_path)
         sickbeard.helpers.fixSetGroupID(subtitle_path)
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

medariox:
@fernandog @duramato 
You had encode problems in the past? Please test!

miigotu:
I see some issues here, please wait to merge until we can talk about it later (I have to run out the door right now)

sickbeard/subtitles.py
```diff
@@ -283,17 +263,22 @@ def get_video(video_path, subtitles_path=None):
...
+        try:
+            # sickbeard.SUBTITLES_DIR is not unicode, try to decode for now
+            new_subtitles_path = sickbeard.SUBTITLES_DIR.decode(sickbeard.SYS_ENCODING)
+        except UnicodeDecodeError:
+            # Fallback to UTF-8. This will almost never happen, but just in case
+            new_subtitles_path = sickbeard.SUBTITLES_DIR.decode('utf-8')
     elif sickbeard.SUBTITLES_DIR:
-        new_subtitles_path = ek(os.path.join, ek(os.path.dirname, video_path), sickbeard.SUBTITLES_DIR)
+        new_subtitles_path = os.path.join(os.path.dirname(video_path), sickbeard.SUBTITLES_DIR)
         dir_exists = sickbeard.helpers.makeDir(new_subtitles_path)
```
miigotu, line 275: sickbeard.helpers.makeDir is accessing the filesystem, this is where new_subtitles_path SHOULD be bytes and not unicode, otherwise an implicit encode happens.

sickbeard/subtitles.py
```diff
@@ -342,8 +327,14 @@ def subtitles_download_in_pp():  # pylint: disable=too-many-locals, too-many-bra
 
         run_post_process = False
         # Check if PP folder is set
-        if sickbeard.TV_DOWNLOAD_DIR and ek(os.path.isdir, sickbeard.TV_DOWNLOAD_DIR):
-            for root, _, files in ek(os.walk, sickbeard.TV_DOWNLOAD_DIR, topdown=False):
+        if sickbeard.TV_DOWNLOAD_DIR and os.path.isdir(sickbeard.TV_DOWNLOAD_DIR):
```
miigotu, line 330: same with isdir and most os.oath.\* calls

miigotu:
I think you have the idea backwards. When it is a param passed to subliminal or as a path, it needs to be utf-8.  Removing ek is what caused all of the issues before SRTV split =P.

medariox:
@miigotu 
Why? Python documentation states that "[Applications are encouraged to represent file names as (Unicode) character strings.](https://docs.python.org/3.1/library/os.path.html)". Subliminal also seems to handle this perfectly fine. I get what you mean with the implicit encoding and that is totally fine as long as we stay consistent and the user sets the correct locale. E.g. if you don't have your system locale to latin-1, you shouldn't use latin-1 chars! What we currently do is encoding everything to utf-8, giving the user the feeling that "everything is fine", even with a completely screwed up locale for example (this will just result in more and more issues!). You want Japanese chars in your file names? You will have to set utf-8 as locale or live with funky chars instead of the Japanese chars. As long as this doesn't break the application, this should be the expected behavior imo.

miigotu:
Passing unicode to the filesystem always causes an implicit conversion. Locale should be irrelevant. Thats why we explicitly encode to utf-8, to guarantee the file system can understand it. Otherwise it generally is auto-encoded/decoded with charmap or c (ascii)

medariox:
That's exactly why we shouldn't force it to use UTF-8, but let python handle the implicit conversion instead. With locale I meant system locale, not python locale. E.g. if a user has a screwed up system locale, he shouldn't expect to be able to use paths or store files with unsupported chars.

miigotu:
Thats the point though, in python2 the locale is always 'C', regardless of system locale. We currently use a hack to force it to utf-8 because we arent encoding it to utf-8 in places that it needs to be. This PR just makes that problem worse.

unicode on the inside is correct, but when it leaves the program to access the filesystem or network  or enter a library you have to explicitly encode it to bytes (such as utf-8). Leaving it as unicode is what causes the implicit encode to ascii.

labrys:
@miigotu The exception there, of course, is if the library is designed to work with unicode, then passing unicode to the library is considered safe (well as safe as can be considered the pita that encodes/decodes prove to be).

PR #910, opened 2016-02-01
Dont show traceback if error sending torrent

fernandog:
https://gist.github.com/GoodOmens83/b268b2143a837703589b


miigotu:
I just realize you aren't importing ex after I merged this.

PR #911, opened 2016-02-01
Dont return false if we don't have code description

fernandog:
@miigotu is there a reason to return false if we don't have code description!? I don't get it. We don't have all status code mapped in common

If we return false we get the old "undefined return code" and we don't get the log message about request:

``` python
logger.log(self.name + u': Response to ' + method.upper() + ' request is ' + self.response.text, logger.DEBUG)
```

### DIFF
--- a/sickbeard/clients/generic.py
+++ b/sickbeard/clients/generic.py
@@ -72,8 +72,9 @@ class GenericClient(object):
         code_description = http_code_description(self.response.status_code)
 
         if code_description is not None:
-            logger.log(self.name + u': ' + code_description, logger.INFO)
-            return False
+            logger.log(self.name + u': Code description is' + code_description, logger.DEBUG)
+        else
+            logger.log(self.name + u': Status code is ' + self.response.status_code, logger.DEBUG)
 
         logger.log(self.name + u': Response to ' + method.upper() + ' request is ' + self.response.text, logger.DEBUG)
 


labrys:
@fernandog which codes aren't mapped?  Particularly, which codes aren't mapped that would be returned for valid data?

fernandog:
the correct why is handle all responses in the clients code

we should check if result is True for response: {"id": 2, "result": true, "error": null} and return True/False
like deluge: https://github.com/SickRage/SickRage/blob/develop/sickbeard/clients/deluge_client.py#L108
like Transmission:
https://github.com/SickRage/SickRage/blob/master/sickbeard/clients/transmission_client.py#L77

but when we return False, we should return also "error" value

others clients like qbittorrent return `_request` response directly
https://github.com/SickRage/SickRage/blob/develop/sickbeard/clients/qbittorrent_client.py#L70

we should check if resposnse is JSON and check is there is any similar field like "result"

as we return False when there's no code_description, we assume we couldn't add torrent. That's wrong.
We are returning False only because we couldn't convert code to description.

fernandog:
**Transmission:**
{"arguments":{},"result":"success"}
{"arguments": {}, "result": "invalid or corrupt torrent file"}
{"arguments":{"torrent-added":{"hashString":"_**","id":__,"name":"_**.1080p.HDTV.X264-DIMENSION"}},"result":"success"}

**Deluge:**
{ “id”: 1, “error”: “Some error description.”, “result”: False} - ?

@SickRage/collaborators would help map the result from all clients?
need to get the log message for : "CLIENT NAME Response to POST request is"
my case: Transmission: Response to POST request is {"arguments":{},"result":"success"}

miigotu:
This returns false if the code_description is NOT None. If it is a successful call, the description is None. If the response from the client is a failure it also sends back a different http status code. Otherwise we need to get the possible responses from all clients and that is tedious.

if you look at the issue reports, the error was happening when the client was returning 'error': null

fernandog:
On this case return code is 200:
{"arguments": {}, "result": "invalid or corrupt torrent file"}

Even 200 code we can have failure
Like this too { “id”: 1, “error”: “Some error description.”, “result”: False}

Not all clients works the same way about http error codes

fernandog:
we only have to check "result" key: transmission and deluge already does. we only have to return the error message. return a dict with {result: True/False, error_msg: string}

miigotu:
they all dont return json, they return with a status code. it works this way.

PR #963, opened 2016-02-08
Fix Unicode decoding errors in newznab logging and ex

labrys:
Just going with a simple `decode(errors='replace')` since it's just for logging.

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -274,10 +274,10 @@ class NewznabProvider(NZBProvider):  # pylint: disable=too-many-instance-attribu
             logger.log(u"Search Mode: %s" % mode, logger.DEBUG)
             for search_string in search_strings[mode]:
                 if mode != 'RSS':
-                    logger.log(u"Search string: %s " % search_string, logger.DEBUG)
+                    logger.log(u"Search string: {}".format(search_string.decode(errors='replace')), logger.DEBUG)
 
                 search_url = posixpath.join(self.url, 'api?') + urlencode(search_params)
-                logger.log(u"Search URL: %s" % search_url, logger.DEBUG)
+                logger.log(u"Search URL: {}".format(search_url.decode(errors='replace')), logger.DEBUG)
 
                 time.sleep(cpu_presets[sickbeard.CPU_PRESET])
                 data = self.get_url(search_url)
--- a/sickrage/helper/exceptions.py
+++ b/sickrage/helper/exceptions.py
@@ -45,7 +45,12 @@ def ex(e):
                 if not message:
                     message = fixed_arg
                 else:
-                    message = '%s : %s' % (message, fixed_arg)
+                    try:
+                        message = u'{} : {}'.format(message, fixed_arg)
+                    except UnicodeError:
+                        message = u'{} : {}'.format(
+                            unicode(message, errors='replace'),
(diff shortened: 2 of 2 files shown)


SickRage2:

Thanks!

sickbeard/providers/newznab.py
```diff
@@ -274,10 +274,10 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
             logger.log(u"Search Mode: %s" % mode, logger.DEBUG)
             for search_string in search_strings[mode]:
                 if mode != 'RSS':
-                    logger.log(u"Search string: %s " % search_string, logger.DEBUG)
+                    logger.log(u"Search string: {}".format(search_string.decode(errors='replace')), logger.DEBUG)
```
miigotu, line 277: search_string is already unicode there.

sickbeard/providers/newznab.py
```diff
@@ -274,10 +274,10 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
...
 
                 search_url = posixpath.join(self.url, 'api?') + urlencode(search_params)
-                logger.log(u"Search URL: %s" % search_url, logger.DEBUG)
+                logger.log(u"Search URL: {}".format(search_url.decode(errors='replace')), logger.DEBUG)
```
miigotu, line 280: urls dont support non-ascii, so urlencode has already converted them.

sickbeard/providers/newznab.py
labrys, line 277: yes it is, but the point is on non-conforming locales (e.g. windows cp1252)  some of those characters are unprintable in the log and raise tracebacks.  after the ex fix this line was raising unicode errors.

sickbeard/providers/newznab.py
labrys, line 280: yeah i figured this one was unneccessary, but just wanted to push it without wasting any more time testing it.

sickbeard/providers/newznab.py
miigotu, line 277: then it would be .encode not decode =P

labrys, line 277: No, and this kind of confusing mistake is why people screw up unicode so often.  The log message is unicode (e.g. u'').  You have to decode the string permissively with replace into the unicode log message, otherwise when the logger tries to perform the logging (thus re-encoding) to a console which doesn't support the characters it goes stupid.  Now if we had control over the logger's method directly and not the kludge that our logger is, we would encode on the output of the log to the console by passing the encoding param to the logger which I'm not sure our configuration supports.  As a side note this may very well lead to mojibake but at least it won't kill the thread and its for logging not processing so we could really care less.

labrys, line 277: Also as another side note, this could also be caused by double-mis-conversion of a unicode character.  Once again, using the decode replace it likely mangles but does so safely.

miigotu, line 277: You don't decode when the string is already unicode ^

I think just switching it to use .format was enough

miigotu, line 277: Actually I just checked the code and they are encoded to utf-8 in genericProvider to avoid having to encode in each provider.
`search_string['Episode'].append(episode_string.encode('utf-8').strip())`
`search_string['Season'].append(episode_string.encode('utf-8').strip())`

PR #1042, opened 2016-02-17
Create gists as secret instead of public

fernandog:
``` python
    def create_gist(self, public, files, description=github.GithubObject.NotSet):
        """
        :calls: `POST /gists <http://developer.github.com/v3/gists>`_
        :param public: bool
        :param files: dict of string to :class:`github.InputFileContent.InputFileContent`
        :param description: string
        :rtype: :class:`github.Gist.Gist`
        """
```

https://github.com/SickRage/SickRage/blob/develop/lib/github/AuthenticatedUser.py#L453

### DIFF
--- a/sickbeard/logger.py
+++ b/sickbeard/logger.py
@@ -290,7 +290,7 @@ class Logger(object):  # pylint: disable=too-many-instance-attributes
                         if LOGGING_LEVELS[level] == ERROR:
                             paste_data = ''.join(log_data[i:i + 50])
                             if paste_data:
-                                gist = git.get_user().create_gist(True, {'sickrage.log': InputFileContent(paste_data)})
+                                gist = git.get_user().create_gist(False, {'sickrage.log': InputFileContent(paste_data)})
                             break
                     else:
                         gist = 'No ERROR found'


SickRage2:

Thanks!

OmgImAlexis:
What's the point in this considering the link is going to be posted publicly anyway?

fernandog:
@OmgImAlexis you can open only if you have a link

if some ones is trying to use gist search to find passwords or whatever, if is 'public' your gist will appear there

OmgImAlexis:
It might as well be the same though, people can just use google to search since it'll be index as it's posted publicly. Just seems pointless.

fernandog:
yes. but new gists wont be possible to do that

Arcanemagus:
@fernandog The point is that the link will still be publicly posted here, so making it a "secret" gist wouldn't really change anything.

duramato:
It reduces the chaces of it beeing crawled, by beeing public on gist it can be crawled from gist and from the issue its posted . By beeing private it can only be crawled from the issue its on

miigotu:
I agree with all of you in this case. Yes it doesnt make it absolutely private, but it does make it slightly less likely to be found.

PPL wont be able to abuse github gist search to find passwords, but google can. Though google will give it a lower ranking because of less linkbacks.

PR #1061, opened 2016-02-20
Plex: Always return True when getting auth token if username/password…

disrani:
… aren't configured

This was causing failures to update Plex server when no username and password
were specified since we were failing to get the auth token

### DIFF
--- a/sickbeard/notifiers/plex.py
+++ b/sickbeard/notifiers/plex.py
@@ -130,6 +130,7 @@ class Notifier(object):
             return False
 
         if not self.get_token(username, password, plex_server_token):
+            logger.log(u'PLEX: Error getting auth token for Plex Media Server, check your settings', logger.WARNING)
             return False
 
         file_location = '' if not ep_obj else ep_obj.location
@@ -220,7 +221,7 @@ class Notifier(object):
             return True
 
         if not (username and password):
-            return sickbeard.PLEX_SERVER_NO_AUTH
+            return True
 
         logger.log(u'PLEX: fetching plex.tv credentials for user: ' + username, logger.DEBUG)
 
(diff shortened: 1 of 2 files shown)


SickRage2:

Thanks!

sickbeard/notifiers/plex.py
```diff
@@ -130,6 +130,7 @@ def update_library(self, ep_obj=None, host=None,  # pylint: disable=too-many-arg
             return False
 
         if not self.get_token(username, password, plex_server_token):
+            logger.log(u'PLEX: Error getting auth token for Plex Media Server, check your settings', logger.ERROR)
```
fernandog: Warning. Errors can be reported

sickbeard/notifiers/plex.py
```diff
@@ -220,7 +221,7 @@ def get_token(self, username=None, password=None, plex_server_token=None):
             return True
 
         if not (username and password):
-            return sickbeard.PLEX_SERVER_NO_AUTH
```
fernandog, line 223: Did you enabled this new setting?

miigotu:
I was planning to do this too, but I wanted some verification with the hidden setting I had added first.

The ini setting and code should be removed for that setting with this change

sickbeard/notifiers/plex.py
miigotu: Yeah this should be warning or not log at all

disrani:
I have changed that to a WARNING, but what is the rule of thumb here for what constitutes an error? I would think not being able to perform a notification should be raised as an error. Without this message a user would have no idea that the notification was not sent.

Also removed the NO_AUTH setting.

fernandog:
User action: warning.
Error: issue, error or else that devs need to change code

PR #1077, opened 2016-02-27
prefer tlsv1

gondalez:
Fixes get requests failing due failed SSL handshake
manifests in sickrage log as 

```
2016-02-27 09:28:10 WARNING  Thread-37 :: Error getting xml for [https://api.example.com/api?apikey=**********&t=caps]
```

debugging in python

before

```
magneto:~# python
Python 2.7.10 (default, Aug 17 2015, 22:55:51)
[GCC 4.3.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import urllib2; urllib2.urlopen("https://api.example.com/api?...").read()
  File "<stdin>", line 1, in <module>
  ... snip ...
urllib2.URLError: <urlopen error [SSL: TLSV1_UNRECOGNIZED_NAME] unknown error (_ssl.c:590)>
```

after

```
>>> import ssl; ssl._create_default_https_context = lambda: ssl._create_unverified_context(ssl.PROTOCOL_TLSv1)
>>> import urllib2; urllib2.urlopen("https://api.example.com/api?...").read()
<caps><server appversion="1.0.0" version="0.1" ... snip ...
```

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -71,7 +71,7 @@ if sys.version_info < (2, 7):
 # https://mail.python.org/pipermail/python-dev/2014-September/136300.html
 if sys.version_info >= (2, 7, 9):
     import ssl
-    ssl._create_default_https_context = ssl._create_unverified_context  # pylint: disable=protected-access
+    ssl._create_default_https_context = lambda: ssl._create_unverified_context(ssl.PROTOCOL_TLSv1)  # pylint: disable=protected-access
 
 import shutil_custom  # pylint: disable=import-error
 shutil.copyfile = shutil_custom.copyfile_custom


SickRage2:

Thanks!

miigotu:
The way to fix this is to use an updated libssl on the system. Forcing the context to tls1 wont work because many sites have disabled tls1 and require tls1.1 or tls1.2

gondalez:
No probs.. makes sense, thanks @miigotu!
I'm running SR on a (discontinued) netgear readyNAS. After your feedback I found that it's running an outdated version of openssl (0.9.8).

I wonder if it's worth applying this hack or logging a warning only when the openssl version is < 1.0.0?
Using `ssl.OPENSSL_VERSION` perhaps.

It would help people like me running on old hardware. Upgrading openssl on the NAS is not trivial since apt-get runs against a limited source etc.

miigotu:
Ill have to think about this a bit. I find it strange that your distro has python 2.7.9+ but not libssl 1.0+

miigotu:
email me the url you are working with, since I cannot reproduce your problem here with 0.9.8

miigotu~gmail

gondalez:
I've emailed that @miigotu. 

I suppose my scenario is pretty unusual. 
The reason for the mismatch is because I've updated python on the NAS but cannot easily update libssl.

So perhaps it is better to close this PR if you can't think of a clean solution.

OmgImAlexis:
Since you're running netgear readyNAS it should be running Debian so you can use [this](https://community.netgear.com/t5/Using-your-ReadyNAS/ROS-6-OpenSSL-and-package-updates/td-p/898919) to update as it's just uses dpkg like all other Debian installs.

miigotu:
RN4 uses a really old debian =P

However, we cant force a lower encryption.

PR #1550, opened 2016-04-18
Add putio client and update search providers templates

miigotu:
Still needs changes, just rebased on develop and fixed conflicts before it gets lost in the updates.

Replaces #1468 
From code it looks like it never logs in, needs tested.

PR #1564, opened 2016-04-19
[Feature] Put.io

devclinton:
Fixes # feature http://feathub.com/SickRage/SickRage/+27

- Add support for put.io client
- Updated code to user DEVELOP as base branch
- Changed URLs to Sickrage.io for client
- Change authentication to access token only and added a Get Token Link. There probably should be an update to https://sickrage.github.io to check for access_token in URL and show a div. 
- Fixed bug in Config Search form on verbiage update on Username/password fields


SickRage2:

Thanks!

gui/slick/views/config_search.mako
```diff
@@ -586,17 +586,19 @@
 
                                 <div class="field-pair" id="torrent_username_option">
                                     <label>
-                                        <span class="component-title" id="username_title">Client username</span>
+                                        <span class="component-title" id="torrent_username_title">Client username</span>
                                         <span class="component-desc">
                                             <input type="text" name="torrent_username" id="torrent_username" value="${sickbeard.TORRENT_USERNAME}" class="form-control input-sm input200" autocapitalize="off" autocomplete="no" />
-                                            <p>(blank for none)</p>
+                                            <p id="blank_username">(blank for none)</p>
+                                            <p id="get_token"><a href="https://api.put.io/v2/oauth2/authenticate?client_id=2392&response_type=token&redirect_uri=https://sickrage.github.io">Get Token</a>
```
OmgImAlexis, line 593: This should have `&` encoded as `&amp;`.

OmgImAlexis:
Does this take into account a directory so people can use `/tvshows` or does it just dump all files into the root directory in put.io?

devclinton:
Since put.io is an online torrent service, you have to use an external
downloader. I have one on my github that can be used,
http://oudenmallon.github.io/putioDownloader/. I can enable the directory
option to be used with that if that is desired. Should I add a tutorial on
how to set that up somewhere?

OmgImAlexis:
I know that, I'm talking about on put.io. You can add sub folders meaning if someone also uses this for CouchPotato as an example they can have that set to use /movies and this to use /tvshows meaning when the downloading software they use downloads/syncs the files they'll end up separated.

If this just adds them to the root put.io directory then they'll be all together with any other torrents they add.

@miigotu how do you feel about adding in put.io downloading like CouchPotato has?

miigotu:
I think we need to worry about getting this part right first.  I kind of want to use the PR I made for it and maybe improve that, because I still see a few bugs in this one.

devclinton:
I have no problems with that. I think it is an great idea. I should be able
to implement that last night. Should I also add an option for local folder
if the user has something to sync their put.io folder with their local
machine. I actually use my putioDownloader script in playlist mode so it
downloads a copy of all the files as a playlist instead of full files. This
allows me to have them indexes by Kodi.

miigotu:
For the download sync I was thinking a resumable FTP client, since put.io has FTP. That would prevent broken downloads from restarting SR and it would also work for future planned FTP providers.

devclinton:
FTP is a good idea. Would an FTP provider give options to manage the put.io
files there - IE delete files that have been already downloaded since you
have a finite space per account on put.io? Would it be possible to support
other download clients for it if it synced to a specific folder? I am
guessing the FTP sync would be part of Sickrage base so that would be a
better option overall but it would be nice to support others if possible.

OmgImAlexis:
Put.io only provides a read only FTP server so deleting wouldn't work at all.

miigotu:
You can use the API to delete though

OmgImAlexis:
By the way for the downloading you can use aria2 like what CouchPotato does.

miigotu:
#1550 @oudenmallon please work on it from that base.

devclinton:
Will do. It may be Monday before I can work on it. I am leaving right after work today to the beach for the weekend. If I get a chance, I will try to test Saturday/Sunday morning. Worst case I will test after work on Monday. Thanks for the work @miigotu

cvockrodt:
I personally can't get it working, not sure what client url to use in the options... 
I've tried (with and without ssl): 
https://api.put.io 
https://put.io 
https://sickrage.github.io
https://sickrage.io

but none of those seem to be working.

After some change, not sure what I did, there was a torrents option in the Manage menu, i.e. manage torrents, which brought up put.io in a frame inside sickrage. Thought that was weird.

Hope we can get some instruction on this

miigotu:
It hasn't even bee n tested yet, as the tag log notes

cvockrodt:
figured it'd still be in develop if it wasn't done, i'm seeing it in master.

miigotu:
If you see how many fixes went out, it was important to release them. The put.io client causes no problems for the overall operation of the app if it doesnt work, so putting it out to get tested doesnt hurt, seeing as it requires an account just to even test it.

PR #1717, opened 2016-05-08
Added optional episode separator to name_parser 'bare' regex

a1studmuffin:
Checklist:
- [x] PR description
- [x] Based on develop branch (latest version)
- [x] Single change per PR

I noticed the post-processor was failing to rename episodes with the format:

show.name.01e05.blah.avi

I extended the name_parser 'bare' regex to add support for an optional episode separator, and also added a name_parser test case to verify the change has not broken any existing tests and also passes a new test for this case.

This is my first contribution ever on GitHub so apologies if I've done something dumb. :)

### DIFF
--- a/sickbeard/name_parser/regexes.py
+++ b/sickbeard/name_parser/regexes.py
@@ -175,6 +175,7 @@ normal_regexes = [
      r'''
      ^(?P<series_name>.+?)[. _-]+                # Show_Name and separator
      (?P<season_num>\d{1,2})                     # 1
+     (e?)                                        # Optional episode separator
      (?P<ep_num>\d{2})                           # 02 and separator
      ([. _-]+(?P<extra_info>(?!\d{3}[. _-]+)[^-]+) # Source_Quality_Etc-
      (-(?P<release_group>[^ -]+([. _-]\[.*\])?))?)?$                # Group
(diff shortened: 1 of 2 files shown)


SickRage2:

Thanks!

duramato:
Bare isn't intended for that naming,  it's for the `lol` releases like `Show.103.hdtv-lol` if I remember correctly. 
Wouldn't it make more sense to edit another one?

a1studmuffin:
Ah I see, I picked 'bare' because - from a pure programming perspective - it seemed like the best fit for the regex modification I added. Which one would you suggest I modify instead? Or is there somewhere I can learn more about the purpose of each regex? Cheers!

miigotu:
Its fine to use that one as long as it doesnt break anything, can you check if it also works with `show.name.s01e05.blah.avi` (add a test maybe?)

a1studmuffin:
Thanks miigotu. Regarding your additional test case request, it appears the 'bare' regex isn't designed to catch cases where a season/episode marker begins with an 's'. The example you provided `show.name.s01e05.blah.avi` is already caught by the 'standard' regex. While investigating this, I also discovered that the 'bare' test cases are currently disabled under BasicFailedTests, due to the existing test `show.name.2010.special.hdtv-blah` failing.

So I think this could go two ways - personally I'm leaning towards 2) as the less risky change:

1) I make this change in the 'standard' regex instead, by making the prefix 's' in `S01E02`, `S01.E02` etc. optional, and adding a new 'standard' test case for it. However, I'm a little nervous that this might open up the 'standard' regex to too many positive results, as this regex accepts any number of digits as the season number (d+), so it may confuse years with the season number in some weird cases, eg. `show.name.2010.e02.blah.avi` would be detected as season 2010 episode 2. By contrast, the 'bare' regex only matches against 1-2 digits for the season number (d{1,2}), so that seems like a safer change, however the test cases are disabled. So the other option would be:

2) I leave the PR as is (I've verified the 'bare' test case does work, and the test case you asked for already exists as a 'standard' regex test case), and at some point in the future we fix the 'bare' test cases.

miigotu:
The BasicFailed tests are intended to fail.  Standard should catch ._s01e01._

a1studmuffin:
Awesome, thanks. Excellent work on the project btw, looking forward to contributing more regularly!

PR #1745, opened 2016-05-11
newpct rss ayer

snamds:
Newpct provider. Added 'ayer' (yesterday) to rss search. Avoid it could lead to entry
misses

### DIFF
--- a/sickbeard/providers/newpct.py
+++ b/sickbeard/providers/newpct.py
@@ -115,7 +115,10 @@ class newpctProvider(TorrentProvider):
                                 # Provider does not provide seeders/leechers
                                 seeders = 1
                                 leechers = 0
-                                torrent_size = cells[labels.index('Tamaño')].get_text(strip=True)
+                                tamanoIdx = 2
+                                if 'Tamaño' in labels:
+                                    tamanoIdx = labels.index('Tamaño')
+                                torrent_size = cells[tamanoIdx].get_text(strip=True)
 
                                 size = convert_size(torrent_size) or -1
                                 item = {'title': title, 'link': download_url, 'size': size, 'seeders': seeders, 'leechers': leechers, 'hash': ''}
--- a/sickbeard/providers/newpct.py
+++ b/sickbeard/providers/newpct.py
@@ -50,7 +50,7 @@ class newpctProvider(TorrentProvider):
         q => Show name
         category_ = Category 'Shows' (767)
         idioma_ = Language Spanish (1)
-        bus_de_ = Date from (All, ayer, hoy)
+        bus_de_ = Date from (All, mes, semana, ayer, hoy)
         """
         results = []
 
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

sickbeard/providers/newpct.py
```diff
@@ -75,57 +73,60 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
...
+                                torrent_row = row.find('a')
+                                title = self._processTitle(torrent_row.get('title', ''))
+                                download_url = torrent_row.get('href', '')
+                                if not all([title, download_url]):
+                                    continue
+
+                                # Provider does not provide seeders/leechers
+                                seeders = 1
+                                leechers = 0
+                                torrent_size = cells[labels.index('Tamaño')].get_text(strip=True)
```
duramato: I think this it's better if it's changed to the absolute  numbering instead of label,  the `ñ` can cause issues in Windows and sys that aren't in unicode

snamds: I've only added lines from 76 to 80. The other changes shown are because of indentation.
The rest of the code is already working ok. I use it on windows.
Columns of this parsed form has changed sometimes so I think it's better to keep the 'tamaño' indexing of the title.

duramato: It depends on the windows language afaik. It might work for you because your locale can encode the character, but I've had to "fix"  a similar issue just yesterday,  with parsing of non ASCII not working for some windows installs,  that's why said that

snamds: Don't think this has to do with my locale. The page retrieved has an explicit 'windows-1252' charset enconding and newpct.py has 'utf-8'. IMHO, that should work anywhere

duramato: Issue isn't server  side,  if it wasn't utf_8 you wouldn't  see it correctly  in the page in  the first place :P
It is today rated to locale,  if your locale doesn't know of the existence  of the char how can it encode/decode it? 
Trt going to a python shell and encode/decode with different encodings you will see what I mean.  Or you can try in a Linux with a locale of and I think should produce the same result. 
TBN that Ubuntu 14.04 has a bug with locale where it most times doesn't work.

sickbeard/providers/newpct.py
```diff
@@ -115,7 +115,10 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
                                 # Provider does not provide seeders/leechers
                                 seeders = 1
                                 leechers = 0
-                                torrent_size = cells[labels.index('Tamaño')].get_text(strip=True)
+                                tamanoIdx = 2
+                                if 'Tamaño' in labels:
+                                    tamanoIdx = labels.index('Tamaño')
+                                torrent_size = cells[tamanoIdx].get_text(strip=True)
```
snamds: Do you like it this way? :P

sickbeard/providers/newpct.py
```diff
@@ -75,57 +73,63 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
...
+                                    continue
+
+                                # Provider does not provide seeders/leechers
+                                seeders = 1
+                                leechers = 0
+                                tamanoIdx = 2
+                                if 'Tamaño' in labels:
+                                    tamanoIdx = labels.index('Tamaño')
+                                torrent_size = cells[tamanoIdx].get_text(strip=True)
+
```
snamds: Do you prefer it that way?

duramato: works, but just "torrent_size = cells[2].get_text(strip=True)" would  be cleaner and safe for all systems :P

snamds: 'ñ' is a spanish character and 'newpct' is a spanish provider with spanish content. I'm pretty sure 99.99999% of people using this provider are spanish and will have a local encoding that will let them exececute "tamanoIdx = labels.index('Tamaño')" with no problems
Anyway "if 'Tamaño' in labels" will avoid problems for the other 0.00001% :P

Just want to be more robust to potential column changes...

duramato: trust me when i say that issues occur, on Windows and in Ubuntu 14.04 regardless of its locale since its broken on it regardless of what you try to do to fix it.

sickbeard/providers/newpct.py
```diff
@@ -75,57 +73,61 @@ def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-man
...
+                                download_url = torrent_row.get('href', '')
+                                if not all([title, download_url]):
+                                    continue
+
+                                # Provider does not provide seeders/leechers
+                                seeders = 1
+                                leechers = 0
+                                #cells[2] is the 'Tamaño' column
+                                torrent_size = cells[2].get_text(strip=True)
+
```
snamds: You win! :S :D

miigotu:
Avoiding yesterday doesnt make it miss anything on RSS, RS is search every 40 minutes. This is just spamming the provider unnecessarily (doubling the hit per client for daily search).

snamds:
Newpct doesn't have a standard rss feed. When I said rss I meant the dailysearcher looking for new episodes on newpct. Sorry for the misleading.
This newpct search was done with a query specifying 'hoy' (today). If a new episode is added at 23.50 and you do the query at 00.30 (40 minutes after), no result is found.
I have also found episodes that have been added with the date of the day before.

Because of this I've added 'ayer' (yesterday) to the results.

miigotu:
In that case you should just use `All` instead of doing 2 requests every daily search update (which runs every 40 minutes.)

snamds:
The options are:
'hoy' (today). Retrieves from 00.00h to current
'ayer' (yesterday). Retrieves from 00.00h to 23.59h of the day before
'semana' (week). Retrieves the last seven days
'All'. Retrieves full database

I've been looking at the history of this provider implementation. 'Week' was done at the begining. It was changed to 'today' but has some inconvenients.
Due to series air hour, is frequent to have new releases near 00.00h. And those can be missed if you only use 'today' in the every 40 min run.
I've done two searches 'yesterday' first and 'today' after. These two searches are the minimum needed to ensure no misses.

miigotu:
Just use Week, 2 searches per daily search will not be merged into the repo. We have too much trouble with providers blocking SR due to requests spamming.

snamds:
Ok. I've reverted to the original code, changed 'today' to 'week' and removed the 'tamaño' string.

PR #2321, opened 2016-09-10
Split core.js

beschoenen:


SickRage2:

Thanks!

beschoenen:
@miigotu I'd like your feedback on this.

miigotu:
@beschoenen It actually used to be in separate files, and it was difficult to maintain that way. We are/were actually in the process of moving everything into core.js as part of a full js rewrite. This allowed for one minified  request from the server to be cached as well.

OmgImAlexis:
This won't work and it'll add around 0.5s of load for every single file you have that hooks into a document load event. That's the reason it was all added to a single file. If you're going to steal the idea of splitting it like I did for Pymedusa you should at least see how I did it so you don't end up with a slower load time.

beschoenen:
@OmgImAlexis I didn't steal anything, we just think alike, I guess ;). I'll have a look at your repo.

miigotu:
@beschoenen no angular.js please, thanks ha

beschoenen:
@miigotu wasn't planning on anything that drastic :P, I just think that there should be a better way than 1 huge-ass file.

OmgImAlexis:
Loading a single file is actually not that bad since it's quicker to load a single file's e-tag, etc. and then load it from cache than it is to load multiple files and their respective e-tags, etc.

If you're going to split it then you need to use the current loading system and drop all of the document loads otherwise you'll end up with race conditions that load the main section before the common code.

Take a look at what we did in pymedusa and you'll see what I'm on about.

PR #2382, opened 2016-09-26
Fix and enhance content of email notification

CoRpO:
- Display correct show name, episode number and quality where available.
- Fix footer html code


miigotu:
What happens when show name or episode name includes a `-` ?

CoRpO:
@miigotu well, it will fail to cut correctly the datas, as with the actual code. Still better than to get an array in the email in place of the show's name. To get the correct datas implies to change the whole notification chain (which is doable btw). But it seems to involve a lot of code change

miigotu:
Actually `_parse_ep` needs to use the `re.search('.+ - (.+?-.+) -.+', ep_name)` to properly split it. This regex was added about a year ago for shows with `-` in the name. Of course that regex would need adjusted to be a full replacement but its the best way

CoRpO:
Better rework the notification chain than to try to create the perfect fit-all-cases regex :/
Correct datas are available. Problem is that I cannot check every notification type.

I'll have a look at both cases (the .sort() in the actual _parseEp is so ugly ...)

PR #2431, opened 2016-10-04
Tagger resolution improvements

snamds:
- Allow tags lists in tags.py
- Added "resolution" regex item that captures "hd 720" (without p)
- Modified newpct provider to treat "hd 720" as "hd 720p x264"


SickRage2:

Thanks!

sickrage/recompiled/tags.py
```diff
@@ -3,7 +3,10 @@
 
 
 # Resolutions
-resolution = re.compile(r'(?P<vres>4320|2160|1080|720|480|360)(?P<scan>[pi])', re.I)
+resolution = [
+    re.compile(r'(?P<vres>4320|2160|1080|720|480|360)(?P<scan>[pi])', re.I),
+    re.compile(r'([sph]d)(.?tv)?.?( )*(?P<vres>4320|2160|1080|720|480|360)(?P<scan>[pi])?', re.I)
```
miigotu: Should not capture extra groups if it can be avoided, also it should make sure the vres is followed by a non digit in cases where scan is missing to prevent false matches as much as possible

``` python
re.compile(r'(:?[sph]d)(:?.?tv)?[ .]*(?P<vres>4320|2160|1080|720|480|360)(?P<scan>[pi])?[^\d]', re.I)
```

sickrage/tagger/episode.py
```diff
@@ -75,7 +80,7 @@ def scan(self):
         """
         attr = 'res'
         match = self._get_match_obj(attr)
-        return '' if not match else match.group('scan').lower()
+        return '' if not match or not match.group('scan') else match.group('scan').lower()
```
miigotu:
``` python
return ('', match.group('scan').lower())[bool(match and match.group('scan'))]
```

miigotu (changes requested):
We have to be extra careful changing the tagger, my suggestions should help keep it reliable.

sickrage/tagger/episode.py
snamds: This wouldn't work. The tuple can't be created because "match.group('scan')" is None. ".lower()" of Nonetype raises an exception

miigotu: doh, ur right. I just dont like the double `not`...

``` python
return match.group('scan').lower() if match and match.group('scan') else ''
```

PR #2530, opened 2016-10-23
Better support for show names with apostrophes

MaximusCub:
- Better support for show names with apostrophes - eg. Also search for "RuPaul's Drag Race" as "RuPauls Drag Race".

### DIFF
--- a/sickbeard/show_name_helpers.py
+++ b/sickbeard/show_name_helpers.py
@@ -135,7 +135,10 @@ def allPossibleShowNames(show, season=-1):
         showNames = get_scene_exceptions(show.indexerid, season=season)
 
     showNames.append(show.name)
-
+    # Show names with an apostrophe may be uploaded with it omitted.
+    if "'" in show.name:
+        showNames.append(show.name.replace("'", ""))
+    
     if not show.is_anime:
         newShowNames = []
         country_list = common.countryList


SickRage2:

Thanks!

beschoenen:
Isn't replacing with `\'` better? @miigotu

MaximusCub:
The problem isn't that we are sending unescaped apostrophes, it is that people actually upload shows to indexers with the apostrophe dropped. 

Searching with the apostrophe only, means that you will not find matches where uploaders decided to omit it. 

This change will make it so that we try to search both with and without. In my tests this helped find shows in the indexers where we previously reported no matches/hits.

OmgImAlexis:
If you're doing this you might as well also add a check for things like `mr.` vs `mr`.

MaximusCub:
I can happily change this so that we have a defined list of characters that we do this for. Would that be preferable ?
If so, it raises the question of how to handle show names with multiple types of such characters. Do we just have one intact search and one fully stripped of these characters? Or do we have every combination in between?

miigotu:
These changes are not needed, since on the search provider end they generally ignore the punctuation in the search and return items with and without the punctuation regardless. If they dont you simply add a scene exception for the show. We cant add an extra search call for every episode/show automatically when most of the time it isnt needed. We have enough trouble with providers not liking how hard we hit them.

MaximusCub:
I tried using the scene exception feature, but could not get it to work. Haven't had a chance to dig in a debug.

miigotu:
you add the scene exception just as it appears on the provider. If your show is `Marvel's Agents of S.H.I.E.L.D` and the indexer has `Marvels Agents of S.H.I.E.L.D.S01E01.RLS` you add a scene exception `Marvels Agents of S.H.I.E.L.D`. It isn't rocket science.

MaximusCub:
The problem isn't that we are sending unescaped apostrophes, it is that
people actually upload shows to indexers with the apostrophe dropped.

Searching with the apostrophe only means that you will not find matches
where uploaders decided to omit it.

This change will make it so that we try to search both with and without. In
my tests this helped find shows in the indexers where we previously
reported no matches/hits.

duramato:
And a Scene exception is just for that, it ads to the search strings.

miigotu:
https://github.com/SickRage/SickRage/pull/2530#issuecomment-255940004

My comment has nothing to do with unescaped apostrophes. Scene exceptions are to add additional names to try with the search.

PR #2683, opened 2016-12-09
added option to disable autorefresh for shows paused and ended

ghost:
Fixes #

- I use sickrage on a old device and i have a large library and when it does the daily SHOWQUEUE-REFRESH it take a couple of hours, making difficult to use the device. In my library i have a lot of show with status "ended" and if you think about it is not necessary to refresh the show every day if the show is ended years ago. So i have added an option in the general configuration, and if you enable it the show will not be automatically refresh. You still be able to force a manual refresh from the show page or the mass update page.


SickRage2:

- [ ] Make sure your PR is based on the DEVELOP branch

Thanks!

miigotu:
Must make this PR to the develop branch.

sickbeard/show_queue.py
```diff
@@ -525,6 +525,11 @@ def __init__(self, show=None, force=False):
         self.force = force
 
     def run(self):
+
+        if self.force == False and sickbeard.SHOWUPDATE_DISABLE_PAUSEDENDED and self.show.status == "Ended" and self.show.paused:
```
miigotu: `if (self.show.status == "Ended" or self.show.paused) and sickbeard.SHOWUPDATE_DISABLE_PAUSEDENDED and  not self.force:`

This should be skipped if either paused OR ended IMHO, thoughts?

sickbeard/__init__.py
```diff
@@ -958,6 +959,9 @@ def path_leaf(path):
         elif SHOWUPDATE_HOUR < 0:
             SHOWUPDATE_HOUR = 0
 
+
+        SHOWUPDATE_DISABLE_PAUSEDENDED = bool(check_setting_int(CFG, 'General', 'showupdate_disable_pausedended', 0))
```
miigotu: I think this should default to 1 (True)

sickbeard/show_queue.py
```diff
@@ -525,6 +525,11 @@ def __init__(self, show=None, force=False):
...
+            logger.log('Skipping show [{0}] because is paused and ended.'.format(self.show.name), logger.DEBUG)
```
miigotu: `paused or ended`

sickbeard/webserve.py
```diff
@@ -3864,7 +3864,7 @@ def saveAddShowDefaults(defaultStatus, anyQualities, bestQualities, defaultFlatt
     def saveGeneral(  # pylint: disable=unused-argument
             self, log_dir=None, log_nr=5, log_size=1, web_port=None, notify_on_login=None, web_log=None, encryption_version=None, web_ipv6=None,
             trash_remove_show=None, trash_rotate_logs=None, update_frequency=None, skip_removed_files=None,
-            indexerDefaultLang='en', ep_default_deleted_status=None, launch_browser=None, showupdate_hour=3, web_username=None,
+            indexerDefaultLang='en', ep_default_deleted_status=None, launch_browser=None, showupdate_hour=3, showupdate_disable_pausedended=False, web_username=None,
```
miigotu: showupdate_disable_pausedended=True (this should be enabled by default I think.)

sickbeard/__init__.py
ghost: the only reason i choose false instead of true is because i didn't want to change the current behaviour, so the user have to activate it.

sickbeard/webserve.py
ghost: like above, the only reason i choose false instead of true is because i didn't want to change the current behaviour, so the user have to activate.

miigotu: I understand that, but it has been requested that this feature be added for a long time and I have just been busy with work. Ended/Paused shows shouldn't be updating unless the user forces it imho (definitely not when paused).  Enabling this by default will earn us some good will with xem and tvdb by our collective userbase not milking their api's so hard.

miigotu: You could just adjust this PR and completely remove the option and just always skip paused shows and I think you get the behavior you want without an added option, and the issue of shows updating when paused is fixed for other people who have complained.  "Ended" is irrelevant then. We always try to avoid adding options and I think this is a case where we can alter the behavior for the better without an option, its a win/win.

sickbeard/show_queue.py
ghost: i choose to use "pause and ended" because there are continuing show that i a want to pause it temporary but i want to keep up to date the list of the episode. An alternative could be to force a show refresh when the user is going to unpause it.

miigotu: The issue people complained about was that when a show is paused it should have absolutely no changes/thread action, for a continuing show you want updates for and also dont want downloads for you should set the default ep status to skip/ignore rather than pause it, and pause should prevent updates and downloads for that show.

sickbeard/webserve.py
ghost: ok for me. I will remove the option and skip refresh if is paused OR ended.

miigotu:
@SickRage/collaborators @SickRage/moderators This should be PR'd to develop, but the conversation needs some 3rd party input. Anyone have thoughts?

sickbeard/webserve.py
miigotu: Just if paused. Ended shows that are not paused already have an algorithm to avoid updates after a certain period of time, that increases depending on how long it has been since it ended.

miigotu: And even with the algorithm they also wont update if they havent changed on tvdb^

ghost: i don't think is working because i have show ended years ago and every day they get updated. In my log i see "INFO SHOWQUEUE-REFRESH :: Performing refresh on XXXXX"

duramato:
This option should be only for paused shows, ended can, and are very likely, to update. ie Prison Break, X-Files, etc. 
But the rest lgtm

sickbeard/webserve.py
miigotu: Thats just a log line that says its starting the thread that does the refresh. Refresh != update.   When the show updater runs is when it checks tvdb, refresh just checks the disk for new episodes or changed files or missing art.

ghost:
i have removed the option and stop for paused shows.

miigotu:
Ok now, move the check out of the QueueItemRefresh to line https://github.com/SickRage/SickRage/pull/2683/files#diff-8e48fe44d6097ae0b4a7ec42fc45cda2R118
Otherwise a queue item will be added for each show that is not run or finished and removed from the queue, and will result in a memory leak.

Lastly, PR this to the develop branch instead of the master branch.

PR #2867, opened 2017-01-09
Fix some queue task initialization issues, add new param to the PP ta…

miigotu:
…sk to force return result of the task, currently disabled

Signed-off-by: miigotu <miigotu@gmail.com>

### DIFF
--- a/sickbeard/post_processing_queue.py
+++ b/sickbeard/post_processing_queue.py
@@ -64,7 +64,7 @@ class ProcessingQueue(generic_queue.GenericQueue):
                     length['manual'] += 1
         return length
 
-    def add_item(self, directory, filename=None, method=None, force=False, is_priority=None, delete=False, failed=False, mode="auto"):
+    def add_item(self, directory, filename=None, method=None, force=False, is_priority=None, delete=False, failed=False, mode="auto", run_now=False):
 
         replacements = dict(mode=mode.title(), directory=directory)
         if not directory:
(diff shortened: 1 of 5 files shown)


SickRage2:

Thanks!

sickbeard/post_processing_queue.py
```diff
@@ -100,13 +100,21 @@ def add_item(self, directory, filename=None, method=None, force=False, is_priori
...
-            logger.log(message.split('\n')[0], logger.INFO)
+            if run_now:
+                with self.lock:
+                    item.run()  # Non threaded, but with queue lock
+                    message = item.last_result
+            else:
+                super(ProcessingQueue, self).add_item(item)
+                logger.log(message.split('\n')[0], logger.INFO)
+
             return message
```
miigotu, line 111: @clinton-hall I could do it this way, but I would really rather move to a non-blocking way, maybe we can figure something else out?

miigotu: If I have to I will move the normal manual pp to a new web uri and leave the old one with run_now for nzbtomedia to use only

miigotu: I also could just add a param to the uri for instant=True/False maybe

clinton-hall: @miigotu 
for most processing this really isn't an issue. The problem comes (using nzbToMedia) with Torrents since I pause the seeding at the start, do the linking/moving, send to SickRage and then resume seeding or delete the torrent.
Obviously deleting the torrent before SickRage has renamed/moved the file is bad, but also resuming seeding can create file locks.

I am wondering if an easier way to look at this, instead of moving the manual pp etc might be to switch nzbToMedia over to a new web uri or even move to api?

Basically, I can make this work with a number of options:
1. API call that returns confirmation when files have been renamed, or returns an error code /message if there was a problem.
2. run_now/instant flag that places this into the top of the queue and waits for the process to complete before returning the status.
3. send to queue but return a queue event id that can be monitored via uri/api to determine when this is complete, and the status.

obviously 1 & 2 both serve to bypass the queue, 3 would however most probably require a lot more coding at your end to add this process monitoring interface?

miigotu: Im adding a param `force_next` to the processEpisode uri, it will return the data back to nzbtomedia is this param is truish.

This will allow it to work the easiest for now until the rest of the api can be rewritten, which might be a long time.

PR #2596, opened 2016-11-06
Add home split options

beschoenen:
Add split options to the home screen.

image

image

Current options are: None, Anime, Sports, Anime & Sports, Root dir.

PR #2904, opened 2017-01-14
Created tabbed show list on home page

### LINKED ISSUE #2902: Tabbed show list
I split my shows in normal shows and anime shows.
But I find it a bit annoying that I always have to scroll down to see my anime list.
Perhaps you could consider having a tabbed show list?
So when the flag 'split show lists' is enabled, 2 tabs are shown instead of splitting then on the same screen.
Just an idea. What do you think?

h3llrais3r:
Fixes #2902

- Use tabs when when splitting shows lists (anime and tv shows) on home screen

Screenshots:

With split option:

image

image

image

image

Without split:
image

Light style:

image


SickRage2:

Thanks!

miigotu:
This is going to conflict with #2596 so we need to get together and decide which way we all think is the best way to go and/or how to incorporate all 3 of our ideas into it.
@beschoenen  thoughts?

My idea is that a dropdown of checkboxes would be the cleanest and most flexible, as it would let you show/hide anime, documentaries, different root dirs, etc.  Like the `Select Columns` on the mass update page.  I also thought about checkboxes in pills like on the show's pages where you can turn on and off episodes with certain statuses, but that would limit the space we have.

EDIT: Your build is failing because of jshint:
```
   ../gui/slick/js/core.js
   1940 |                activate: function( event, ui ) {
                                                    ^ 'ui' is defined but never used.
   1940 |                activate: function( event, ui ) {
                                             ^ 'event' is defined but never used.
```

Unrelated:
I also have an idea I would like to try to rotate the table column headers on mass update for the columns with only checkboxes: http://jimmybonney.com/articles/column_header_rotation_css/

miigotu:
Also could you add a screenshot of how it looks?  We just have several ways to go with this and visual aid might help decide whats best.

h3llrais3r:
Screenshots added.

h3llrais3r:
I also had a quick look at #2596, but isn't that just a split on the root dirs?
You can already filter on them by using the dropdown, so why another dropdown to filter on it?
Although it would be nicer to have a dropdown of the root with the paths...

Secondly you can't filter on anime with #2596 when the anime series are in the same root dir as all the other shows.

Finally I like the tabs better than a dropdown because we only need to click once to switch.
And also with the tabs, you can still filter on root dirs.

I think we have the most flexibility with my PR, but it's still open for improvements. 😉

miigotu:
I agree that a single click to switch is nice, but the visual change is not so good. For example, custom backgrounds can be used and this blocks the background, and the show-container no longer looks like its floating.
Also, tabs show even when you have no anime and when SPLIT_HOME is false.

image
image

The problem is that there is no one-size-fits-all solution here, so I just want to discuss it thoroughly.

I also see some similar issues with @beschoenen PR, there is added busy-ness to the ui and the design is focused around the simple layout. His takes 2 clicks to change view.

I think we need a compromise between the two, and possibly a bit added on top.

If we sacrifice the 1 click convenience and add another click (most ppl wouldnt be changing views that often):

A drop down similar to `Select Columns`, maybe called `View Options` with checkboxes for:

subheading `Type`: `Anime`, `Sports`, `Normal`, where sports and anime options would be hidden if you had no sports or anime shows, the entire subgroup hidden if there is only an option for one
type (with the single option selected in the background)

subheading `Split` with options `Anime`, `Sports`, `Root`, again not showing sports or anime if there are none of those shows in the show list, not showing `Root` if there is only one root directory, with the entire heading hidden if none of the split options would be valid

subheading `Paths` with each root directory listed that has a show on it, and the `Paths` subheading not showing if there is only one root.

Possibly a subheading `Layout` that incorporates the `Layout` dropdown into this combo dropdown, same for Sort_by and Direction

With it dynamic this way it would be as simple as possible for people with simple setups, yet they have no less function than anyone else, and still everyone has total control over what and how shows are displayed. This way it would be 2 clicks to get what you are looking for, but it is also extremely flexible and allows you to have a very nice custom selection because of all of the different combinations you can choose from. It also doesn't change the UI's overall look/feel/layout much since it would just take the place of the current `Root Directory` dropdown (and possibly also reducing the clutter if incorporating the layout option also).

I understand it would be easier to explain this with an example but I simply don't have the time to work out all that js unless we all think its a good idea lol, js is no fun for me.

EDIT: I understand too this doesnt solve your need to scroll to see anime, so either you would have to click the dropdown and ncheck `Normal` or we could also work in one other subheading for the order the split is displayed, anime-normal-sports, normal-anime-sports, w/e... Either way we need to be mindful that anime is a tertiary feature, and have to try and not effect the vast majority of people this doesnt make sense for.

h3llrais3r:
Can you give an example of the problem with the custom background and show me what's not nice with the visual change?

I think that the tab when nothing is splitted can easily be removed. I'll check for that.

And I don't actually want to filter the home screen. This can already be done with the existing root dropdown. The reason why I added the tabs it that I can quickly go to my anime shows.
It's a lot faster than clicking a dropdown and then filtering on anime. 
So basicly: the homescreen should load everything and with the tabs, you can easily go from normal to anime by click the tab. This is a lot better than scrolling down to the anime list (in current version) and filtering (@beschoenen PR).

Just something to think about. :)
Perhaps we can include @beschoenen his PR here or this one in his?

miigotu:
Perhaps we could do all I said above, incorporating all of the dropdowns on home into one and adding those new options, AND add subheading `Split Type` with options like `Tabbed`, `List`, `Root`

miigotu:
If you guys think thats a good idea I will add the combobox and the js needed to set the options in webserve and trigger reload of the showlist onclick, @h3llrais3r you could add the tabbed interface, and @beschoenen could adjust the layout for split type root

miigotu:
Hiding background example (if it were more shows in my test install it would look much worse):

image
image

h3llrais3r:
How do you get those backgrounds on home screen? I never saw that...
But I even think that you should not show backgrounds on home screen. 😄 
You already have those on the show details page.

h3llrais3r:
@miigotu @beschoenen What will we do know?
Perhaps first start with the split option of @beschoenen and then see if we can introduce the tabs?
Another possibility is to have pills instead of tabs. Then we don't have too much impact on the global background image... 😄

h3llrais3r:
Tried something else to solve the issue with the tabs with the background.
We can also use pills 😉 

image

image

image

image

Just a trial, we can still play with colors and positioning of the pills.

miigotu:
What if people want to see both anime and normal at the same time?  Thats why I think it should be toggles

h3llrais3r:
If the split is not enabled in the config, we don't show them. 😉

h3llrais3r:
@miigotu any feedback yet from @beschoenen?
Perhaps I have another proposition, which has much less impact.
If we take the PR of @beschoenen with the split options, we just add some quick links at the top (in form of button or something else) to navigate quickly to the part you want. 
We add a quick link to each header of the table we want to go to. By doing this, everyone can choose what they want. Either splitting or no splitting and having a way to quickly go to the right subset of your library on the page.
So if we choose for example split on 'anime', a quick link 'anime' will appear at the top to quickly go to the anime shows. So we don't need to scroll down a lot when we have a big library. (This is the main reason why I started this PR 😄 )

However, adding a 'tabbed home' option in the split options is the probably the best way to go. 😉

beschoenen:
Hey guys,
Sorry for not replying earlier, I've been very busy up until now.

I must say, I like the tabs and the custom background very much (the pills not so much).

Would it still be possible to split on root directory, though? Cause that's what I like about my PR.

h3llrais3r:
Hi @beschoenen.
Yes, we would add different split options: split on root directory, split on anime, ...
And perhaps add a split toggle saying: split in tabs. So when enabled, the split is still done based on your dropdown, but the way to visualize it is done based on the toggle? So basicly: keep your implementation and the possibility to show the split in the existing way or the tab way.
What do you think?

miigotu:
I still like the idea of a button->popover div minicking a dropdown. It __removes 3-4 dropdowns from the ui__ and adds in one, while expanding the existing combination of capabilities.  Nothing is lost and the ui looks better

o = radio select within the option group
[ ] = checkbox select
```
 _________________
(Display Options v)
 ------------------
  |--Display--------------|
  |    o Poster           |
  |    o Small Poster     |
  |    o Banner           |
  |    o Simple           |
  |--Sort Order-----------|  <---- Only show this group in poster display
  |    o Name             |
  |    o Date             |
  |    o Network          |
  |    o Progress         |
  |--Sort Direction-------|  <---- Only show this group in poster display
  |    o Ascending        |
  |    o Descending       |
  |--Split Type-----------|
  |    o Show Type        |
  |    o Root Directory   |
  |--Split Method---------|  <---- Only show this group if there are sports or anime (if split type is Show Type) or more than one root dir (if plsit type is root)
  |    o Tabbed           |
  |    o Grouped          |
  |    o Pills            |
  |--Show Types-----------|  <---- Only show this group if there are sports or anime
  |    [ ] Normal         |
  |    [ ] Sports         |
  |    [ ] Anime          |
  |--Root Directories-----|  <---- If there is only one root dont show this group (the options will be custom root labels user can configure)
  |    [ ] Blah           |
  |    [ ] AnotherRoot    |
  |--Columns--------------|  <---- Only show this group when __not__ in poster display (or maybe keep it as a seperate dropdown?)
  |    [ ] Next Episode   |
  |    [ ] Prev Episode   |
  |    [ ] Show           |
  |    [ ] Network        |
  |    [ ] Quality        |
  |    [ ] Downloads      |
  |    [ ] Size           |
  |    [ ] Active         | 
  |    [ ] Status         |
  ------------------------
```
And use js to attach actions to clicks in the different groups.

@beschoenen @OmgImAlexis @h3llrais3r is there some reason this isn't feasible/desirable?

h3llrais3r:
I personally think it's too crowded when mixing different things together.
And I don't see anywhere my 'tabbed feature' which is an easy way to toggle between different types of shows. So for me this has no added value... it's only another way of displaying the current options (which I never use).
However, if we also add a toggle to switch the view from a normal split to a tabbed split, it becomes more interesting. 😉

miigotu:
@h3llrais3r Added Split mode to the diagram, Show tabs/pills when those are selected. The point I'm making with the dropdown method is that more options can be retained while still being able to cater to everyones needs/use cases, and it also ADDS more flexibility while removing 4 dropdowns from the main ui (that most people only use very rarely)

You can show or hide specific root directories, specific types of shows, split them how you want, sort them how you want, display them in the mode you want, sort them how you want, etc all from ONE place
Also note that most of the groups in my diagram are not visible unless certain conditions are met, so it is actually MUCH less busy than it looks

beschoenen:
@miigotu I agree with you that we have to do something about the mess surrounding the various buttons and drop-down boxes when switching view modes, it can be really confusing.

h3llrais3r:
Ok, seems better than.
What are we going to do about the slider for the size of the posters?
We keep it and show it when poster is chosen?

miigotu:
Yes, on poster view you would have the slider, filter show name, and the display options dropdown. Other views would be display options and clear filters button

miigotu:
Another idea, all of those options in my diagram could be shown in a modal instead of a popover, with confirm/cancel buttons. That might look better and make more sense.

PR #2980, opened 2017-01-25
Fix tvdb indexer search with special characters

h3llrais3r:
Searching for tv/anime shows with special characters sometimes does not give results. (For example: chaos;child)
By applying unidecode on the parsed series names we can still have a match.


SickRage2:

Thanks!

miigotu:
No unidecode please. The result is utf-8 so just decode it.

sickbeard/classes.py
```diff
@@ -158,7 +160,7 @@ def selectSeries(self, allSeries):
                     series_names.extend(curShow['aliasnames'].split('|'))
 
                 for name in series_names:
-                    if re.sub(r'[^\w\s]', '', search_term.lower()) in re.sub(r'[^\w\s]', '', name.lower()):
+                    if re.sub(r'[^\w\s]', '', search_term.lower()) in re.sub(r'[^\w\s]', '', unidecode(name.lower())):
```
h3llrais3r, line 163: @miigotu The unidecode is the only way to get a match because it replaces the 'ä' in the tvdb search result by 'a' so we get a match with the regex.
Just using the utf-8 encode gives back 'chäos;head' as search result but due to the regex the 'ä' is replaced by '' and this doesn't match our search string 'chaos;head'.
Searching on 'chäos;head' does not even return search values from tvdb.
Any other ideas?

miigotu: Ill just merge it, I dont mind that much anymore lol

h3llrais3r: I agree that we should use the good encoding/decoding normally, but I think in this case it's the only way to get a match on your search.

miigotu: Doesnt really matter much to me either way for two reasons, 1) I dont watch foreign shows (lol), and 2) The tvdb lib needs replaced with the new one so it all has to change soon anyways

Issue #2996, opened 2017-01-27
[BUG] Re-scan Files with Scene Numbering enabled


heymoe:
Post this bug a year ago pretty much to the day and it might have been forgotten / in limbo in the "old-sickrage-issues" section so I'm re-submitting it. It's still broken with the latest release.

Link to the original issue report: https://github.com/SickRage/old-sickrage-issues/issues/673
as well as a cut-n-paste of the initial report.

----

Branch: Master
Commit hash: 3f0efd1338faf8796fd355bada4f546ec1cb5ed6
Your operating system and python version: CentOS 5.11 / Python 2.7

American Dad! is one of those shows where the episode numbering on theTVDB and what the scene uses are all out of sync.

Example: theTVDB: Season 8 Episode 1 = Scene: Season 7 Episode 1

XEM Link for episode mappings: http://thexem.de/xem/show/2913

So using the example above, s08e01 (tvdb) is wanted. SR set to Scene Numbering looks up in XEM and finds that s08e01 (tvdb) = s07e01 (scene) so it searches and finds that available. SR triggers the download of s07e01 (scene), processes the download and places the new file in the "Season 08" folder as "American Dad! - s08e01.mp4" (tvdb). Everything is working as expected so far. This is the expected behavior.

The Problem: So, SR has now successfully found and downloaded the correct file (s07e01 - scene) for s08e01 (tvdb) that was wanted and placed it in the correct directory and named it properly on the filesystem.

If you now click "Re-Scan Files", SR sees the file "American Dad! - s08e01.mp4" on the file system in the "Season 08" folder and starts processing that file again as part of the "Re-Scan". Refreshing the "American Dad!" page in SR once the Re-Scan is done, SR will now show that s09e01 has been found and if you hover over the episode # to see what the path is to the file for the episode (s09e01) it shows "Season 08/American Dad! - s08e01.mp4". s08e01 (scene) = s09e01 (tvdb)

Basically SR seems to be applying the XEM lookup on the files on the file system as if those files were scene episodes and not already processed and sorted tvdb episodes which is causing episodes that haven't actually been downloaded to be marked as downloaded.

What I expected is that SR would see the already downloaded and processed file on the file system as being the episode it says it is (tvdb) and process it as such instead of thinking it found a new file that it hasn't seen before and process it like it was a scene numbered episode which would translate that file to the wrong episode.

I hope this all makes sense. If not let me know and I'll try and explain the problem better.. Thanks!

[...]

PR #3004, opened 2017-01-28
Skip scene detection when rescanning files

h3llrais3r:
Seems to fix https://github.com/SickRage/SickRage/issues/2996 but impact on other things is not known/tested.
But as far as I can see, the touched code in the tv.py is only called on refreshDir, so it should be ok.

### DIFF
--- a/sickbeard/name_parser/parser.py
+++ b/sickbeard/name_parser/parser.py
@@ -98,7 +98,7 @@ class NameParser(object):
                 else:
                     self.compiled_regexes.append((cur_pattern_num, cur_pattern_name, cur_regex))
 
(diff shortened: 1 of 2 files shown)


SickRage2:

Thanks!

miigotu:
I think this could be done much easier, but I might be wrong.

Can you test without these changes and then just changing the lines in _parse_string to:
`if bestResult.show.is_scene and not self.showObj:` ?

My idea is that is self.showObj then it must be a refresh/update/rescan since we initialized it with a show object.

miigotu:
Merging this for now, because its fairly obvious that this should work from what I see. If we come up with a better idea later we can change it then.

PR #3064, opened 2017-02-01
Fix mass edit season folders

sharkykh:
Fixes Mass Edit inverting/setting wrong values into Season Folders/flatten_folders

- Invert flatten_folders value in massEditSubmit [webserve.py]
- Fix massEdit setting the wrong 'Season Folders' value [manage_massEdit.mako]

See commits descriptions for explanations.

### DIFF
--- a/gui/slick/views/manage_massEdit.mako
+++ b/gui/slick/views/manage_massEdit.mako
@@ -149,8 +149,8 @@
                     <div class="col-lg-9 col-md-8 col-sm-7 col-xs-12 component-desc">
                         <select id="flatten_folders" name="flatten_folders" class="form-control form-control-inline input-sm">
                             <option value="keep" ${('', 'selected="selected"')[flatten_folders_value is None]}>&lt; ${_('Keep')} &gt;</option>
-                            <option value="enable" ${('', 'selected="selected"')[flatten_folders_value == 0]}>${_('Yes')}</option>
-                            <option value="disable" ${('', 'selected="selected"')[flatten_folders_value == 1]}>${_('No')}</option>
+                            <option value="disable" ${('', 'selected="selected"')[flatten_folders_value == 0]}>${_('Yes')}</option>
+                            <option value="enable" ${('', 'selected="selected"')[flatten_folders_value == 1]}>${_('No')}</option>
                         </select>
                         <label for="flatten_folders">${_('Group episodes by season folder (set to "No" to store in a single folder).')}</label>
                     </div>
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -3457,7 +3457,7 @@ class Manage(Home, WebRoot):
                 new_air_by_date = show_obj.air_by_date
             else:
                 new_air_by_date = True if air_by_date == 'enable' else False
-            new_air_by_date = 'on' if new_air_by_date else 'off'
+            new_air_by_date = 'off' if new_air_by_date else 'on'
 
             if flatten_folders == 'keep':
                 new_flatten_folders = show_obj.flatten_folders
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -3457,13 +3457,13 @@ class Manage(Home, WebRoot):
                 new_air_by_date = show_obj.air_by_date
             else:
                 new_air_by_date = True if air_by_date == 'enable' else False
-            new_air_by_date = 'off' if new_air_by_date else 'on'
+            new_air_by_date = 'on' if new_air_by_date else 'off'
 
             if flatten_folders == 'keep':
                 new_flatten_folders = show_obj.flatten_folders
             else:
                 new_flatten_folders = True if flatten_folders == 'enable' else False
-            new_flatten_folders = 'on' if new_flatten_folders else 'off'
+            new_flatten_folders = 'off' if new_flatten_folders else 'on'
 
             if subtitles == 'keep':
                 new_subtitles = show_obj.subtitles


SickRage2:

Thanks!

miigotu:
Those values werent wrong. Season Folders is the inverse of flatten_folders.  So (Season Folders = 1) == (flatten_folders = 0)

sharkykh:
Before I start, the reason I made the change to `manage_massEdit.mako` like it is, is because of the [illogicality of the massEdit form](https://github.com/SickRage/SickRage/blob/develop/gui/slick/views/manage_massEdit.mako#L152-L153).
```mako
<select id="flatten_folders" name="flatten_folders" class="form-control form-control-inline input-sm">
    <option value="keep" ${('', 'selected="selected"')[flatten_folders_value is None]}>&lt; ${_('Keep')} &gt;</option>
    <option value="enable" ${('', 'selected="selected"')[flatten_folders_value == 0]}>${_('Yes')}</option>
    <option value="disable" ${('', 'selected="selected"')[flatten_folders_value == 1]}>${_('No')}</option>
</select>
```
This talks about the select[id=flatten_folders]:
When it's value='enable', the user selected Season Folders = Yes
When it's value='disable', the user selected Season Folders = No
This is a stray from your logic of `(Season Folders = 1) == (flatten_folders = 0)`
According to this `(Season Folders = 1) == (flatten_folders = 1)`

This is my reason for commiting https://github.com/SickRage/SickRage/pull/3064/commits/68f388a67e8d744ca9357a6de035ea6c16726354.

## Now,
>Those values werent wrong. Season Folders is the inverse of flatten_folders. So (Season Folders = 1) == (flatten_folders = 0)
>@miigotu

That's true, but because [`sickbeard/webserve.py: editShow`](https://github.com/SickRage/SickRage/blob/develop/sickbeard/webserve.py#L1493) inverts the value of flatten_folders
```python
flatten_folders = not config.checkbox_to_value(flatten_folders)  # UI inverts this value
```
Stuff gets messed up.

Let's say you submit massEdit form to [massEditSubmit](https://github.com/SickRage/SickRage/blob/develop/sickbeard/webserve.py#L3462-L3466):
```python
if flatten_folders == 'keep':
    new_flatten_folders = show_obj.flatten_folders
else:
    new_flatten_folders = True if flatten_folders == 'enable' else False
new_flatten_folders = 'on' if new_flatten_folders else 'off'
```
with:
- flatten_folders='enable': (when user selects 'Yes')
  - new_flatten_folders = 'on'
    - editShow inverts it to 'off'
       - Season Folders = OFF
**this is why it's working this way right now**
- flatten_folders='disable': (when user selects 'No')
  - new_flatten_folders = 'off'
    - editShow inverts it to 'on'
       - Season Folders = ON
**again, this is why it's working this way right now**

## However:
- flatten_folders='keep': (when user selects '< Keep >')
  - new_flatten_folders = **current show's flatten_folders status**,
     let's say - 1 / 'on'
    - editShow inverts it to 0 / 'off'
       - Season Folders = OFF

### **do you see the issue?**

You **_can_** just invert `new_flatten_folders` when `flatten_folders='keep'` and it will fix it.

But my point is that the logic in the Season Folders section on the massEdit form
is against the widespread logic of SickRage that is, like you said
`(Season Folders = 1) == (flatten_folders = 0)`
so it needs to be corrected as well.
I believe this is also why it's a bit confusing to explain...

I tried my best to explain my point, hope that's clearer.

I rest my case :sweat_smile:

P.S. [I made a change in webserve.py (I edited the wrong line by mistake, that's fixed now)](https://github.com/SickRage/SickRage/pull/3064/commits/a56028f3ec32eb7ddca7e396f7ed19ce15616ea6)

miigotu:
I'd rather refactor and rename FLATTEN_FOLDERS I think.

sharkykh:
@miigotu 
> I'd rather refactor and rename FLATTEN_FOLDERS I think.

I just finished going over the files and _(slowly but surely)_
changing `flatten_folders` in all its forms to `season_folders`. Texts & titles included.

The only issue I have left is how to update the current database scheme for this change
(i.e. for when a user updates SickRage after this change has been committed)

I have two ideas but I'm not sure if any of them are the best solution,
so I'd like your opinion before I continue:

1. Running a check of the database on start up,
and if there is a `flatten_folders` column in the `tv_shows` table scheme, we do as follows:
**In config.ini:** change `flatten_folders_default = X` to `season_folders_default = not X`
**In sickbeard.db:** for every show in `tv_shows` we will update `flatten_folders` to its inverted value, and then change the table scheme from `flatten_folders` to `season_folders`.

2. We could run a check with **git** just before an update or a checkout to see if the target branch and commit have **this flatten_folders => season_folders** commit merged to them.
We can do that with [`git branch --contains <commit>`](http://stackoverflow.com/questions/1419623/how-to-list-branches-that-contain-a-given-commit/1419637#1419637) and check output for target branch name.
Then, of course, the same changes as in the 1st solution.
We could also turn back the database if the branch doesn't contain that commit if we wanted to...


**Solution (1)** is good, _**however**_ if the user decides to checkout another branch or commit that that change isn't included in, season folders are gonna be a mess for him/her.

**Solution (2)** is ... well, weird (?) and very specific, and now that I think about it, we can't possibly know the commit SHA before committing it.
We could commit and than commit again with the SHA of the first commit embedded to that git branch check, but that seems a bit too much for just small change.

Either way, my solutions don't offer a "one time change".
They all add one more check to start-up and that seems ... not the best.

**While writing this I realized that there are database versions, so we could just go one version up.**
**How does that help me?**

miigotu:
You could just leave the db column as flatten_folders and invert just before saving it to db, and when loading from db. Then when we finally get a big db update planned out we can fix it then along with the other major db changes.

PR #3271, opened 2017-03-01
New unpack 'Treat as video' option to treat RARs as video files

miigotu:
Quick n dirty config only option for allowing rar files in the show folder without unpacking them, pp will just move them if enabled and unpack is off

Fixes #

-
-
-


SickRage2:

Thanks!

miigotu:
@sharkykh mind adding html and code for changing this value from the GUI? If DONT_UNPACK is enabled then UNPACK should be automatically disabled when changing the setting
I really need a break lol

sharkykh:
1. Wait, why are you adding another setting if we already have UNPACK?
2. Doesn't make sense, Markup-language-wise, to do checkboxes that disable each other... You could just do that with radio buttons... Something like
   - 0 = Unpacking disabled (ignore rar files)
   - 1 = Unpacking enabled (extract rar files and process the media files they contain)
   - 2 = Move rars to show folder (do not extract)

I would only do that tomorrow anyway, I think

miigotu:
It was really just a quick PoC test for someone on irc, It can be adjusted however necessary.

sharkykh:
Wait, did I get this right?
This option would allow the user to move/copy/symlink/hardlink rar files, as-is, from TV Download dir to be _**in the show's folder**_? like, next to the video after post-processing? :confused: 

I don't think this looks good...

image

miigotu:
Yes, it renames and moves the rar to the show folder, just as if it were a video file itself. Many modern media players can play videos in rars directly, so it saves the user drive space

sharkykh:
Still working on the GUI, didn't touch any python just yet.
What do you think? Better?
image

miigotu:
Sure, I also need to find a better way to auto-find unrar, but thats a different story.

sharkykh:
I'm gonna change `is_media_file` back because I think it'll cause a chain of issues if someone actually decides to use this option. This function gets used in many modules...
Just wanted to let you know. I know what you said...
>  It can be adjusted however necessary.

miigotu:
You dont need to change it back. If the option is enabled then it IS a media file. Thats the point.

miigotu:
I mean, ofc it CAN be done another way, but this is the simplest way and everything will work just fine with how it is too.

sharkykh:
OK, I you're positive it's gonna work then I'll keep it as it is.

sharkykh:
erm... [What about that?](https://github.com/SickRage/SickRage/blob/b0cdd5788e894c90ecc31089947eaf225b447037/sickrage/helper/media_info.py#L119)

miigotu:
It returns (None, None), thats the only method `qualityFromFileMeta` that actually reads the media file, and it returns gracefully.

sharkykh:
Uh, can't push to the branch

miigotu:
Make a PR to this branch ^

miigotu:
Promoted you to a collaborator so you can push to this remote.
Be careful not to push tags, and don;t push direct to develop/master

sharkykh:
```
remote: Permission to SickRage/SickRage.git denied to sharkykh.
fatal: unable to access 'https://github.com/SickRage/SickRage.git/': The requested URL returned error: 403
Pushing to https://github.com/SickRage/SickRage.git
```
Still denied (guess it hasn't updated yet)
I'll open a PR now (had some local issues with my commits till now..)

sharkykh:
I thought it would copy the commit notes too, sorry.
> - Change UNPACK from bool to int,
Implement the 'Treat as video' option for archives
> - Fix add_site_message to support translations

gui/slick/views/config_postProcessing.mako
```diff
@@ -272,22 +272,36 @@
...
+                                        ${_('What to do with archived releases found in your <i>TV Download Dir</i>?')}
+                                    </label>
+                                </div>
+
+                                <div class="col-lg-9 col-md-9 col-sm-8 col-xs-12 pull-right component-desc">
                                     <div class="row">
                                         <div class="col-md-12">
-                                            <input id="unpack" class="enabler" type="checkbox" name="unpack"
-                                                ${('', 'checked="checked"')[bool(sickbeard.UNPACK)]} />
-                                            <label for="unpack">${_('unpack any TV releases in your <i>TV Download Dir</i>?')}</label>
```
miigotu, line 279: Need to have a label attached to the input. If it needs to be hidden use class `sr-only` on the label

miigotu (commented):
Also, need to add a config migration, so that user's existing setting is not lost in their config just by updating.

gui/slick/views/config_postProcessing.mako
sharkykh: [This is the label ](https://github.com/SickRage/SickRage/pull/3271/files/4e03c68c5f7cb72ee3981a8241587b5eff995fd7#diff-b7e36dabf5283a72bc76bbb6d7c258c5R275). I'll just add a for="unpack" to it.

sharkykh:
About migrating the config there's no need.

Original UNPACK | Config read | New UNPACK
--------- | ------ | ---
False | check_setting_int(0) | 0
True | check_setting_int(1) | 1

It worked for me, like it should because of:
[check_setting_int](https://github.com/SickRage/SickRage/blob/b0cdd5788e894c90ecc31089947eaf225b447037/sickbeard/config.py#L577-L580)

gui/slick/views/config_postProcessing.mako
miigotu: Yep, just has to be linked with `for=` =P

miigotu:
Maybe that one should also add a site_message if not sickbeard.DEVELOPER and no providers are enabled for the mode

sharkykh:
I'll try and add that

sharkykh:
This will show up only one time: Startup. I didn't think putting it in every search cycle is smart...
It changes according to whatever needs attention.
Link is to /config/providers ofc
I'd love your feedback on my code.

image

miigotu:
"Adjust enabled providers backlog/daily check site_message code. When indexing a 2 item tuple the index value should only be 0, 1,(min 0, max len(tuple) - 1), True, or False. Anything else will result in an IndexError. Also, break out of the loop once we know we have both, to save a little time starting up. Auto-(add|remove) the message when saveProviders"

PR #3321, opened 2017-03-05
Adjust unrar location code to try and autofind unrar, and possibly do…

miigotu:
…wnload unrarw32.exe and use that on windows.

PR #3354, opened 2017-03-07
Auto download UnRAR.exe for Windows users (revisited)

sharkykh:
* Reintegrate #3321 (disabled in https://github.com/SickRage/SickRage/commit/3cf9d37986f04186cec555c4d4e30f4ff79d2f97) - **Automate UnRAR.exe downloading for Windows**
Original unrar.exe (+licence.txt) from RARLab.com will be hosted at sickrage.github.io as a compressed zip file (reduces size in half...)
* Change the method for finding a WinRAR installation (I needed it to report whether a tool was found)
### The following PR needs to merge before this PR goes live: https://github.com/SickRage/sickrage.github.io/pull/230

### DIFF
--- a/sickbeard/config.py
+++ b/sickbeard/config.py
@@ -132,11 +132,12 @@ def change_unrar_tool(unrar_tool, alt_unrar_tool):
             # Look for WinRAR installations
             found = False
             winrar_path = 'WinRAR\\UnRAR.exe'
-            for check in (
-                os.path.join(os.environ["ProgramFiles(x86)"], winrar_path),
-                os.path.join(os.environ["ProgramW6432"], winrar_path),
-
-            ):
+            # Make a set of unique paths to check from existing environment variables
+            for check in {
+                os.path.join(location, winrar_path) for location in (
+                    os.environ.get("ProgramFiles(x86)"), os.environ.get("ProgramW6432"), os.environ.get("ProgramFiles")
+                ) if location
+            }:
                 if ek(os.path.isfile, check):
                     # Can use it?
                     try:
--- a/sickbeard/config.py
+++ b/sickbeard/config.py
@@ -133,11 +133,14 @@ def change_unrar_tool(unrar_tool, alt_unrar_tool):
             found = False
             winrar_path = 'WinRAR\\UnRAR.exe'
             # Make a set of unique paths to check from existing environment variables
-            for check in {
+            check_locations = {
                 os.path.join(location, winrar_path) for location in (
                     os.environ.get("ProgramFiles(x86)"), os.environ.get("ProgramW6432"), os.environ.get("ProgramFiles")
                 ) if location
-            }:
+            }
+            check_locations.add(os.path.join(sickbeard.PROG_DIR, 'unrar\\unrar.exe'))
+
+            for check in check_locations:
                 if ek(os.path.isfile, check):
                     # Can use it?
                     try:
--- a/sickbeard/config.py
+++ b/sickbeard/config.py
@@ -131,20 +131,17 @@ def change_unrar_tool(unrar_tool, alt_unrar_tool):
         if platform.system() == 'Windows':
             # Look for WinRAR installations
             found = False
-            install_paths = set()
+            winrar_path = 'WinRAR\\UnRAR.exe'
             # Make a set of unique paths to check from existing environment variables
-            if "ProgramW6432" in os.environ:  # it's simpler in Vista or newer
-                install_paths.update([os.environ["ProgramW6432"],  # Program Files
-                                      os.environ["ProgramFiles(x86)"]]),  # Program Files (x86)
-            else:
-                install_paths.add(os.environ["ProgramFiles"])  # Program Files / Program Files (x86)
-                if "ProgramFiles(x86)" in os.environ:  # is Windows 64 bit
-                    install_paths.update([os.environ["ProgramFiles(x86)"],  # Program Files (x86)
-                                          re.sub(r'\s?\(x86\)', '', os.environ["ProgramFiles(x86)"])])  # Program Files
-            install_paths = { os.path.join(location, 'WinRAR\\UnRAR.exe') for location in install_paths }
-            install_paths.add(os.path.join(sickbeard.PROG_DIR, 'unrar\\unrar.exe'))
+            check_locations = {
+                os.path.join(location, winrar_path) for location in (
+                    os.environ.get("ProgramW6432"), os.environ.get("ProgramFiles(x86)"),
+                    os.environ.get("ProgramFiles"), re.sub(r'\s?\(x86\)', '', os.environ["ProgramFiles"])
+                ) if location
+            }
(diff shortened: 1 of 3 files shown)


SickRage2:

Thanks!

sickbeard/config.py
```diff
@@ -116,26 +116,50 @@ def change_unrar_tool(unrar_tool, alt_unrar_tool):
         rarfile.ORIG_UNRAR_TOOL = unrar_tool
     except (rarfile.RarCannotExec, rarfile.RarExecError, OSError, IOError):
         if platform.system() == 'Windows':
-            for unrar_tool in ('C:\\Program Files\WinRAR\\UnRAR.exe', 'C:\\Program Files (x86)\WinRAR\\UnRAR.exe'):
+            # Look for WinRAR installations
+            unrar_tool_found = False
+            for unrar_tool in ('C:\\Program Files\\WinRAR\\UnRAR.exe', 'C:\\Program Files (x86)\\WinRAR\\UnRAR.exe'):
```
miigotu: We should probably use %programfiles(x86)% and %ProgramW6432%

```python

winrar_path = 'WinRAR\\UnRAR.exe'
for unrar_tool in (os.path.join(os.environ["ProgramFiles(x86)"], winrar_path), os.path.join(os.environ["ProgramW6432"], winrar_path)):
```

miigotu: This should also check for `.\\unrar\\unrar.exe`, for when the user updates or deletes config.ini and the exe has already been downloaded in their restored Data folder.

miigotu (dismissed):

sharkykh:
I only looked through my phone but looks good.

You said we also needed to check if the file has already been downloaded to `sickbeard.PROG_DIR/unrar/unrar.exe`
We could add that to when we're checking for WinRAR installations

I'll do that soon (if you won't beat me to it haha)

Also might be a small issue with the environment vars - WinXP doesn't support either. 
https://en.m.wikipedia.org/wiki/Environment_variable#cite_ref-Schulz_2014_Ordner_9-18

sharkykh:
actually `os.environ["ProgramW6432"]` will possibly throw a KeyError exception on Windows versions prior to Vista.
`os.environ["ProgramFiles"]` will return the (x86) directory if python is running in 32bit mode on 64bit Windows, or the regular Program Files in all other cases.
We can use this. Locale is changing, yes, but the (x86) stays.
So how do you feel about this?
```
os.path.join(os.environ["ProgramFiles"], winrar_path),
os.path.join(re.sub(r'\s?\(x86\)', '', os.environ["ProgramFiles"]), winrar_path),

re.sub(r'\s?\(x86\)', '', 'C:\\Program Files (x86)')
'C:\\Program Files'
re.sub(r'\s?\(x86\)', '', 'C:\\Program Files')
'C:\\Program Files'
re.sub(r'\s?\(x86\)', '', 'C:\\Program Files(x86)')
'C:\\Program Files'
```

Edit: after I posted I realized that I didn't account for "C:\Program Files" being the first value

miigotu:
Eh that doesnt feel right for some reason, I'd rather make them a set and check multiple variables
Gimme a sec ill update that part

miigotu:
Better?

sharkykh:
I spent the better part of an hour now coming up with a way to please every version of windows:
```python
install_paths = set([os.path.join(sickbeard.PROG_DIR, 'unrar\\unrar.exe')])

if "ProgramW6432" in os.environ:  # it's simpler in Vista or newer
    install_paths.add(os.path.join(os.environ["ProgramW6432"], winrar_path),  # Program Files
                      os.path.join(os.environ["ProgramFiles(x86)"], winrar_path))  # Program Files (x86)
else:
    install_paths.add(os.path.join(os.environ["ProgramFiles"], winrar_path))  # Program Files / Program Files (x86)
    if "ProgramFiles(x86)" in os.environ:  # is Windows 64 bit
        os.path.join(os.environ["ProgramFiles(x86)"], winrar_path),  # Program Files (x86)
                     os.path.join(re.sub(r'\s?\(x86\)', '', os.environ["ProgramFiles(x86)"]), winrar_path))  # Program Files
for check in install_paths:
.........
```

Haven't tested yet.

miigotu:
Just checking all 3 variables wont make it work? Check my last commit

sharkykh:
`ProgramFiles(x86)` and `ProgramW6432` won't be available in any 32bit,
`ProgramW6432` won't be available in versions of Windows older than XP,
and finally, if python is running as 32bit in 64bit Windows - `Program Files` will return the (x86) folder.

That's why it's complex...

I will however implement your way to iterate through and apply `os.path.join` to them

sharkykh:
I tested it through cmd and through a python console,
but because I'm using Win7, I don't test the other vars.
So I need to adjust locally and test them. Doing so now.

**edit: looks good.**

sharkykh:
Implementing tests with mocking.. Don't merge yet.

miigotu:
I think you could have left the way I did it and added in 
```python
check_locations.add(re.sub(r'\s?\(x86\)', '', os.environ["ProgramFiles"]))
```
That way if it is 64 bit windows it always have a version with and without (x86)
32bit windows will always have without (x86)
on all windows versions, even in WOW64 mode

The way you have it is harder to follow.

sharkykh:
But wouldn't os.environ.get() throw a KeyError whenever 'ProgramFiles(x86)' or 'ProgramW6432' aren't available?

miigotu:
No, .get wont throw an error, it will return None if it isnt a key, thats why in the end of the set comprehension I check if location

sharkykh:
OK, then I agree with you.
I'll make the changes

miigotu:
```python
>>> os.environ.get('alala')
>>> os.environ['lala']
  File "<input>", line 1, in <module>
  File "/home/dusted/Envs/sickrage/lib/python2.7/UserDict.py", line 40, in __getitem__
    raise KeyError(key)
KeyError: 'lala'
```
=P

sharkykh:
I wish I understood mock better. I'm sure there are easy ways to mock the rarfile functions to fit into our crazy scenarios, I just don't know how.

miigotu:
Testing for windows is very limited in our test environment. Need to use AppVeyor or similar for windows specific tests like this

tests/config_tests.py
```diff
@@ -271,16 +272,31 @@ def test_change_unrar_tool(self):
...
+            self.assertTrue(config.change_unrar_tool('unrar', 'bsdtar'))
+
         my_environ = mock.patch.dict(os.environ,
-                                     { 'ProgramFiles': 'C:\\Program Files (x86)\\',
-                                       'ProgramFiles(x86)': 'C:\\Program Files (x86)\\',
-                                       'ProgramW6432': 'C:\\Program Files\\' }, clear=True)
+                                     {'ProgramFiles': 'C:\\Program Files (x86)\\'}, clear=True)
         with my_environ:
-            self.assertFalse(config.change_unrar_tool('unrar', 'bsdtar'))
+             self.assertFalse(config.change_unrar_tool('unrar', 'bsdtar'))
```
miigotu: indentation ^ =P

PR #3355, opened 2017-03-08
Fix for RAR contents not being deleted

sharkykh:
RAR contents wasn't being deleted
**when the following criteria were met:** 

Setting | Value
------ | ---------
**Processing Mode** | is `Auto`
**Process Method** | _**not**_ `Move`
**Delete RAR Contents** | is `True` [selected]

It should be fine, not supposed to break anything.
All I did was add a check for the process_method.

[ #3345 ]

Issue #3230, opened 2017-02-26
[APP SUBMITTED]: [Errno 2] No such file or directory: '/volume5/Downloads/103 NZB/Complete/101 Episodes/Primeval S05E02 WS PDTV XviD-BiA/Primeval.S05E02.WS.PDTV.XviD-BiA.nzb'


HPMSGEN8:
### INFO
Python Version: **2.7.13 (default, Dec 19 2016, 02:32:04) [GCC 4.7.3 20130102 (prerelease)]**
Operating System: **Linux-3.10.77-x86_64-with-glibc2.2.5**
Locale: UTF-8
Branch: **master**
Commit: SickRage/SickRage@8965a3145b69c292e7c139ae065d5f8fa7c57dda
No Log available with ERRORS:
### ERROR
```
MANUAL :: [8965a31] Unable to move file /volume5/Downloads/103 NZB/Complete/101 Episodes/Primeval S05E02 WS PDTV XviD-BiA/Primeval.S05E02.WS.PDTV.XviD-BiA.nzb to /volume3/Episodes 3/Primeval/Season 05/Primeval S05E02 Episode 2.nzb: error 2 : No such file or directory
```
---
_STAFF NOTIFIED_: @SickRage/owners @SickRage/moderators

SickRage2:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Include basic information: Branch/Commit, OS, What you did, What happened, What you expected
- [ ] Enable debug logging (be sure to disable after the bug is fixed)
- [ ] Post debug logs, either inline (for smaller logs) or using gist

Please make sure you also read [how to create an issue](https://github.com/SickRage/sickrage-issues/blob/master/README.md) and followed all of the steps. 

The title should describe your issue. Having "SR not working" or "I get this bug" for 100 issues, isn't really helpful. We will close issues if there isn't enough information.

Sometimes the devs may seem like grunts and respond with short answers. This isn't (always) because the dev hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let us know, and this bot may get updated to automatically answer you.

Thanks!

sharkykh:
Hey @HPMSGEN8, please test develop branch.
We need your feedback, thanks!

stef64:
Ho....funny !
Just happened to me, I have that behaviour sometimes but it does save it properly.
The problem in my case is my drive is in standby and need few second to spin-up.
[...]

PR #3358, opened 2017-03-08
Post-Processing modifications

miigotu:
Replaces #3355 

#### TODO:
- [x] Make `unrar()` handle one rar at a time, and then loop through rars in `process_dir()`,
and process each one in `process_dir()` just after extraction so we know whether to delete the rar or not when `method == 'move'`
- [x] `sickbeard.NO_DELETE` is not used - needs clarification on it's purpose
_"Don't delete empty folders - leave empty folders when Post Processing?
note: can be overridden using manual Post Processing"_
**Maybe supposed to be [here](https://github.com/SickRage/SickRage/blob/a633004336168a9a9fa194904adcfdedbb7bc6dc/sickbeard/processTV.py#L215-L215) with delete_on**

Test results [@sharkykh]
=========

**Release variations:**
---
- [x] Scene-packed release (rar + `r00`-`rNNN` files)
  :warning:  with move method - extracts and processes okay, _**but:**_
  - [x] ~contains files rNN and nfo files.~
  - [ ] doesn't delete the downloaded folder.
```
06:33:52 DEBUG::POSTPROCESSOR-AUTO :: Validating folder C:\Github\SickRage-Dev\Tests\Completed_Downloads\Lethal.Weapon.S01E15.720p.HDTV.X264-DIMENSION
06:33:52 DEBUG::POSTPROCESSOR-AUTO :: C:\Github\SickRage-Dev\Tests\Completed_Downloads\Lethal.Weapon.S01E15.720p.HDTV.X264-DIMENSION : No processable items found in folder
```
- [x] Downloaded folder with associated .mkv, .srt, .nfo + RARBG .mp4 + *.txt + Screens folder with .png
- [x] Single rar file in TV Download dir
- [x] NZB Release (like from nzbToMedia) and https://github.com/SickRage/SickRage/issues/3230#issuecomment-285331397

**Process methods + mode:**
---
* Auto
  - [x] move
  - [x] copy (=all others)
* Manual + Manual with release_name (NZB-like)
  - [x] move 
  - [x] copy (=all others)

**Various Options/Settings/Config:**
---
- [x] delete rar contents [ On: :white_check_mark: ; Off: :white_check_mark:  ]
   - [x] ***Fixed*** - ~X **Failed:** RAR Contents was deleted when DELRARCONTENTS was `False` [LOGS](http://pastebin.com/YQZd0hZQ)~
- [ ] Associated files
- [ ] No delete
- [ ] todo


SickRage2:

- [ ] Make sure your PR is based on the DEVELOP branch

Thanks!

miigotu:
@sharkykh can you test this six ways from sunday? This is for sure a proper change, but it might need more fixes also.

sharkykh:
I'm probably too tired to answer but I have to say a couple of things
1. You are correct - that delete rar contents after looping through extracted directories was a workaround.
When I added this there were like 4 or 5 issues about post processing and I didn't want to break anything else at the time.
2. Keep in mind that the reason it doesn't delete the extracted folders right now is because it doesn't allow for the delete_on override for when process method is not move, lets say copy. It just continues the process_dir loop without deleting (line 216)

We should try to simplify this bit of code in either this branch or my branch..
This needs more work and ofc like you said - testing

miigotu:
`if not(process_method == "move" and result.result) or (mode == "manual" and not delete_on):`
With my change delete_on is True for rar contents if sickbeard.DELRARCONTENTS or delete_on

Last commit I made all rar contents to `move` (the old behavior before refactor) and no release_name (which is only for nzbtomedia)

sharkykh:
Yes, that makes a bit more sense to me right now :+1:
However, the `delete_on=sickbeard.DELRARCONTENTS or delete_on` does not. What value does it yield?
Btw, [this is where we set `delete_on` for auto post processing](https://github.com/SickRage/SickRage/blob/master/sickbeard/post_processing_queue.py#L130-L131) and we have to take it into consideration

miigotu:
I dont think those 2 lines are necessary either. 99% of the problems were due to a wrong var used in the pp for delete, and list_to_allow problem. So we have a bunch of patches added after that that were trying to fix the underlying problem but they just added more complexity.

miigotu:
delete_on is only relevant to manual post processing, and when we fake it for rar contents.

Now, the last problem I see is in list_associated_files where it deletes files even when mode != 'move'
This prevents users from being able to seed.

I think the file_path_list_to_delete and DELETE_NON_ASSOCIATED_FILES code needs removed from there, those files are deleted in proc_dir

miigotu:
That also removes the need for the rename change lol

sharkykh:
This is not the first time that I look at `list_associated_files`'s name and  think about how misleading it is.
This function really isn't supposed to do anything more than to **_list_** them.

sharkykh:
Just a thought - we might want to take the chance to properly document process_dir (and some other key methods and functions).
I feel like readabillity is the no.1 issue with the post processing method(s)

miigotu:
Which is why I was slowly refactoring them. Its so complicated and unreadable.

miigotu:
Sorry @sharkykh @h3llrais3r for removing so much of the code you guys added there, but this is the best route for now.
We need to be simplifying it rather than complicating it =P

miigotu:
I dont have time to do testing for this for now:
renaming a show in the show folder with subtitles and not MOVE_ASSOCIATED
unpacking a rar with DELRARCONTENTS and without
calling the api (manual pp but with a rar name in the params as nzbName)
etc

sharkykh:
Merging this **_will have to wait_**. 
We should make some kind of checklist to make sure we don't miss on some edge cases,
We MUST implement process_dir into pp_tests.py... maybe with fake video files (1 second videos)

miigotu:
they can be a file with random bytes

miigotu:
I added another fix. This was probably the #1 cause of the pp issues.
Also added a todo for a way to extract and process each rar in sequence, rather than extracting all rars during pp and then processing them after, so we know when to delete the rar or not.

sharkykh:
Top down makes it go from the deepest folder (furthest down the tree) up, right?
What does it fix?
It looks kinda weird in the logs:
```
08:53:18 INFO::POSTPROCESSOR :: Auto post processing task for C:\Github\SickRage-Dev\Tests\Completed_Downloads was added to the queue
08:53:19 INFO::POSTPROCESSOR-AUTO :: Beginning auto post processing task: C:\Github\SickRage-Dev\Tests\Completed_Downloads
08:53:19 DEBUG::POSTPROCESSOR-AUTO :: Processing in folder C:\Github\SickRage-Dev\Tests\Completed_Downloads
08:53:19 INFO::POSTPROCESSOR-AUTO :: Processing C:\Github\SickRage-Dev\Tests\Completed_Downloads
08:54:44 DEBUG::POSTPROCESSOR-AUTO :: Processing folder C:\Github\SickRage-Dev\Tests\Completed_Downloads\Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv]\Screens
08:55:56 DEBUG::POSTPROCESSOR-AUTO :: C:\Github\SickRage-Dev\Tests\Completed_Downloads\Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv]\Screens : No processable items found in folder
```

Also, if we do topdown=false here, then we should remove [os.walk from validate_dir altogether.](https://github.com/SickRage/SickRage/blob/a633004336168a9a9fa194904adcfdedbb7bc6dc/sickbeard/processTV.py#L290-L311)
What's the point of it going through the same folders we just checked one loop ago (the generator_to_use loop in process_dir I mean)

- .\Completed_Downloads
  process_dir -> os_walk
  - .\Completed_Downloads\Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv]\Screens
    - validate_dir -> os_walk -> No processable items found in folder
  - .\Completed_Downloads\Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv]
    - validate_dir -> os_walk
      - .\Completed_Downloads\Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv]**\Screens**
        - no found files
      - .\Completed_Downloads\Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv]
        - found files, run parser...etc

   ... etc

miigotu:
Im not too sure on it yet, still working it out in my mindscape =P

sharkykh:
Is process_dir supposed to delete folders even if they aren't empty?
I can't figure out when to check if folder is empty and when not to.

(I'm debugging through the code with test downloads)

miigotu:
Until I adjust how extractions work, idk. Right now it handles it correctly but it leaves the rar behind.

sharkykh:
No no, not rars,
Nvm, I will change it to delete everything for now just so I could continue testing.
ofc according to `if not(process_method == "move" and result.result) or (mode == "manual" and not delete_on):`

miigotu:
I think thats supposed to be sickbeard.NO_DELETE, but I think NO_DELETE isnt supposed to be about `empty folders` its just whether it should check if a folder is empty or not before deleting it.

miigotu:
And that topdown will probably come out, im not sure yet

miigotu:
I think its really close, neeeeeds tests

miigotu:
How is this working?

sharkykh:
Test results 
=========
Moved to first comment

sharkykh:
I'll post comments as I go through.

While testing processing a rar with method=move
```
06:15:04 INFO::POSTPROCESSOR-AUTO :: Processing succeeded for C:\Github\SickRage-Dev\Tests\Completed_Downloads\Lethal.Weapon.S01E15.720p.HDTV.X264-DIMENSION\lethal.weapon.115.720p-dimension\Lethal.Weapon.S01E15.720p.HDTV.X264-DIMENSION.mkv
06:15:56 DEBUG::POSTPROCESSOR-AUTO :: Found unwanted files: [u'Lethal.Weapon.S01E15.720p.HDTV.X264-DIMENSION.mkv']
```
`process_media` was okay,
but the unwanted files list is outdated (the file has already been moved).
https://github.com/SickRage/SickRage/blob/fcde20ca993ef05dd65188f8d9768107a0f988a6/sickbeard/processTV.py#L211-L214

miigotu:
@sharkykh should fix this line =P

```python
unwanted_files = filter(lambda x: x in video_files + rar_files + ['.stfolder'], file_names)
```
to
```python
unwanted_files = filter(lambda x: x not in video_files + rar_files + ['.stfolder'], file_names)
```

sharkykh:
@miigotu On the next test it included associated files .srt + .nfo that were moved (obviously because they're not filtered)
```
Found unwanted files: [u'Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv].eng.srt', u'Lethal.Weapon.S01E15.HDTV.x264-LOL[ettv].nfo', u'RARBG.mp4', u'Torrent-Downloaded-from-ExtraTorrent.cc.txt']
```

sharkykh:
RAR Contents was deleted when method was 'move' and sickbeard.DELRARCONTENTS was `False`
### [LOGS](http://pastebin.com/YQZd0hZQ)

miigotu:
Eh, when DELRARCONTENTS iz false should make the method copy instead of move when processing rar contents
method=("copy', "move")[sickbeard.DELRARCONTENTS]

sharkykh:
Yeah, you removed the method_fallback in commit number two =P

----
I'll write that in
Perfect.  :white_check_mark: Fixed.

miigotu:
I might have to put the generator_to_use back in too.
Need to see if having a release_name if it still parses the whole download dir or just the release like it's supposed to

sharkykh:
With what parameters does nzbToMedia call post-processing?
I don't use nzb, how does a release folder look like?

miigotu:
Its the same as a manual pp, with delete_on, force_next, proc_dir (folder the release is in), nzbName (the rar or video to process)

miigotu:
When nzbName is passed, only that file should process

miigotu:
I think we can make further modification.
The delete files/folders could be done in the _move method, since in no other method should anything be deleted
Also, when looping over rars, after extraction and processing contents the rar should only be deleted when method == 'move' I think.
Thoughts?

sharkykh:
> The delete files/folders could be done in the _move method, since in no other method should anything be deleted

Sounds good, except for the fact that it's like a function inside a function inside a function :stuck_out_tongue: 
> Also, when looping over rars, after extraction and processing contents the rar should only be deleted when method == 'move' I think.

Yep. If the user selects 'move' it means he doesn't care about the downloaded files (not seeding the rar)


I would test some more later.

sickbeard/processTV.py
```diff
@@ -163,34 +164,36 @@ def process_dir(process_path, release_name=None, process_method=None, force=Fals
...
+            if result.result and result.current_extracted_dir:
+                # process method for rar contents is either `move` or 'copy' (depends on sickbeard.DELRARCONTENTS's value)
+                # delete_on is always set to sickbeard.DELRARCONTENTS here because all of these files are extracted files.
+                # we don't need to delete the folder here if the delete_on is set correctly, because it will be deleted in process_dir
+                # TODO: List files associated with the rar and pass them through to process_dir? Inner video filename may not match the rar release name
+                # And maybe only the video was rar'd?
+                result = process_dir(
+                    process_path=result.current_extracted_dir,
+                    release_name=None,
+                    process_method=('copy', 'move')[sickbeard.DELRARCONTENTS],
```
sharkykh: When DELRARCONTENTS is True, and you're running Manual post-processing without selecting delete files,
it moves (and deletes) the RAR contents.

We may want to change this to:
```python
process_method=('copy' if not (mode == 'manual' and delete_on) else 'move') if sickbeard.DELRARCONTENTS else 'copy'
```
or
```python
process_method=('copy', ('move', 'copy')[not (mode == 'manual' and delete_on)])[sickbeard.DELRARCONTENTS]
```

whatever you feel is more readable

miigotu: It should, delete_on has nothing to do with rar contents

miigotu: And, I was thinking of a way to remove some of those options. They are case specific settings from way back for when the PP wasnt moving files that it should, so people wanted to see what was left behind to make sure it moved the files it needed to.

miigotu: But for now, maybe we can make that change you suggested, and see how ppl react.

sharkykh: Okay I finally fixed my code..
I also forgot to say I tested `copy` method

So the ideal is that manual post-processing would work the exact same way the settings are?
Meaning we'll remove delete_on altogether?

> It should, delete_on has nothing to do with rar contents

For now, while Manual Post-Processing has this:
image
We should set ONE logic and follow it in `process_dir()`

**Or we create more problems with all of that "here it has an effect" and "here it doesn't have an effect"...**

sharkykh: >But for now, maybe we can make that change you suggested, and see how ppl react.

👍

miigotu: I think it should always delete rar contents, always delete unwanted files when method is 'move', and always delete empty folders.
manual pp would only differ because of release_name/nzbName (same thing) then
I am sure we will get some complaints from people who obsess over their subtitles, but only a few.

sharkykh: manual pp will also differ with the ability to <kbd>Force already Post Processed Dir/Files</kbd> and <kbd>Mark Dir/Files as priority download</kbd>

sharkykh: I think we should delete the option from manual p-p right now.
It's confusing me even as I'm debugging/testing

miigotu: Just leave the delete_on param in the method signature of webserver.processEpisode or it will break nzbToMedia

sharkykh:
Is `.stfolder` supposed to be an unwanted file or not?
Should it be listed and deleted or not?

papaxris:
.stfolder is used by file synchronization apps.  It generally should be kept; usually because the user placed it in the folder to allow the folder to be scanned for files.

miigotu:
Yeah the pp is supposed to totally ignore .stfolder. eg: It should not list it, move it, delete it, or anything.

sharkykh:
@miigotu I'm losing my confidence in this... 
Listing associated files for the rar files isn't perfect.
It ignores files not in ALLOWED_EXTENSIONS and they stay in the folder containing the rar.
Also after operation on the folder containing the rar is complete, the directory doesn't get deleted.

😩 Oh yeah, and I'm also exhausted from this.

sickbeard/processTV.py
```diff
@@ -176,47 +181,53 @@ def process_dir(process_path, release_name=None, process_method=None, force=Fals
...
+                                 or '.stfolder')]
         if unwanted_files:
             result.output += log_helper("Found unwanted files: {0}".format(unwanted_files), logger.DEBUG)
 
         delete_folder(ek(os.path.join, current_directory, '@eaDir'), False)
         delete_files(current_directory, unwanted_files, result)
 
-        if delete_folder(current_directory, check_empty=(sickbeard.NO_DELETE, delete_on)[sickbeard.NO_DELETE]):
+        if delete_folder(current_directory, check_empty=sickbeard.NO_DELETE):
             result.output += log_helper("Deleted folder: {0}".format(current_directory), logger.DEBUG)
```
miigotu: to here,
I want to remove all of this, and just do
`delete_folder(current_directory, check_empty=True)`

The delete of other files can happen in move method and it makes it much simpler. NO_DELETE and DELRARCONTENTS can be eliminated too I think.

sickbeard/processTV.py
```diff
@@ -176,47 +181,53 @@ def process_dir(process_path, release_name=None, process_method=None, force=Fals
...
 
         if not validate_dir(current_directory, release_name, failed, result):
             continue
 
         video_files = filter(helpers.is_media_file, file_names)
         if video_files:
-            process_media(current_directory, video_files, release_name, process_method, force, is_priority, result)
+            process_media(current_directory, video_files, release_name, process_method, force, is_priority, assoc_files, result)
         else:
             result.result = False
```
miigotu, line 225: I want to remove all of the code from here....

sickbeard/processTV.py
```diff
@@ -176,47 +181,53 @@ def process_dir(process_path, release_name=None, process_method=None, force=Fals
...
                     process_path=result.current_extracted_dir,
                     release_name=None,
-                    process_method=('copy', 'move')[sickbeard.DELRARCONTENTS],
+                    process_method=('copy', 'move')[sickbeard.DELRARCONTENTS or process_method == 'move'],
                     force=force,
                     is_priority=is_priority,
-                    delete_on=sickbeard.DELRARCONTENTS,
                     failed=failed,
                     mode=mode,
+                    assoc_files=PostProcessor(rar_full_path, rar_file).list_associated_files(rar_full_path),
```
miigotu: Im still not sure if we need to do this, technically associated files should be in the rar.

sharkykh: > technically associated files should be in the rar.

I'd beg to differ. [This is from sceneaccess.eu](https://cloud.githubusercontent.com/assets/10238474/23809194/62d4d0f2-05d5-11e7-90f8-afac5dc3e895.png) (probably the only private tracker account I have)
NFO is (essentially) an associated file and as you can see, it's outside the RAR.
That's usually the case with private trackers, from what I've seen in the last... 5 years or so. They haven't changed a thing. Only public trackers discourage archiving TV/Movie releases.


And on top of all that, with this code it's _**still not getting deleted**_ if it's not in ALLOWED_EXTENTIONS

sickbeard/processTV.py
sharkykh: I don't know...
What good will that do?

miigotu: Because files should only ever be deleted if they are rar contents, or method is move. Removing all of this extra code and doing the delete in the move method makes it much clearer that it is only supposed to delete when method is move and the it's get a lot simpler.

sharkykh: If we do that,
it should be _**as close as possible**_ to the associated files code (listing and deleting),
then it would be even more simple.

miigotu: Technically in the move method it would just delete the entire folder 
```python
if result.result and not validate_dir:
    delete_folder(path, False)
```

And where the other code was removed `delete_folder(current_dir, True)`

sharkykh: > And where the other code was removed `delete_folder(current_dir, True)`

That with have to be with some form of `if not(process_method == "move" and result.result)`

sharkykh: Umm. edited cause I got confused there.
:white_check_mark: In the move command you would delete the folder if it was successfully processed.

But the code snippet you wrote will delete the folders if no media was found in them. That would make some people mad if they want some other stuff to remain in their folder.
I know what you're going to say - they shouldn't put anything other than downloads for/of SickRage. So we'll just have to put a "warning" note, or a note that sticks out, in the config - saying folders in there that do not contain processable video files will be deleted (**only on move?**)

sickbeard/processTV.py
miigotu: OK so we can pass associated files in, NBD... The other ideas still fit though

sharkykh: Yeah but it is a hacky fix. We might be better off doing a small method just for processing extracted folders, and get rid of all that recursiveness?

miigotu: It's only 1 level of recursion so I think it's OK, I like seeing the exact flow in one place there instead of jumping around in the file 50 places

sickbeard/processTV.py
miigotu: No, the delete_folders in the process_dir would only delete empty folders. (check_empty=True always)

In the move method it would only delete a folder if a video was processed successfully out if it, there are no more processable items, and it wasn't the TV_DOWNLOAD_DIR... So nothing that doesn't belong to SR would be deleted

sharkykh: Ok so the code is for the move command, that's where I got confused.
You're going to call `validate_dir` again? (I've become an `os.walk` hater :laughing: )

miigotu: I'm not sure about that part yet, but moving the deletes for sure

miigotu:
Need to finish this one up soooon =P

HPMSGEN8:
Hi @miigotu and @sharkykh just to let you know that .vob is also an file that is not getting removed. Example - source-mighty.morphin.power.rangers.s01e39.dvdrip.xvid-deimos.vob  

The .avi & .mkv is getting processed 100%

sharkykh:
IIRC vob files are video files too, are they not processed at all?

HPMSGEN8:
They are Sample Files so no these should get deleted on the ones i got. under 20MB

papaxris:
I have seen instances where Anime scene releases are VOB files.  Not necessarily are VOB files limited to sample files.  Just FYI.

HPMSGEN8:
@papaxris thanks i did not know that :)

miigotu:
@sharkykh what do you think about biting the bullet and merging this? lol its scary

sharkykh:
@miigotu Indeed it is... 😨
There are still a few small kinks I want to sort out. I'll jump into this.
But it's not like it's not working right now.
**In retrospect,** I should've changed my "live" SickRage (as opposed to the development one) to this branch in the meantime, to see if there are any unforeseen errors and such..

miigotu:
yeah "it's working...but the code sucks...but it's working...but the code suck...but it's working" lol

sharkykh:
I don't think listing the associated files for the inner process_dir (for rar files) was the right call..
It's really hard to bring myself to edit this code..

toeggi:
Hi! As this was set to Stale, does that mean the fix was never committed to master? I have issues with different files not being deleted, and as such the folders of already pp videos are left behind. These files can be .txt .exe and .png. Just a thought from my non-dev brain; could a fix be to add functionality to define what files should be moved along the video (already there with the ass. files) and to define what files should be deleted/cleaned up after pp?

papaxris:
Please re-open due to new duplicate posts on the same issue.

PR #3367, opened 2017-03-09
Extend locale support to include javascript files

sharkykh:
To-do checklist
---
- [x] Enable text extraction from `*.js` files
**(no `*.min.js`, and no sub-dirs of `gui/**/js/`)**
- [ ] Import a `gettext` for javascript.
Found some... need to check if they are compatible:
  - [alexanderwallin/node-gettext](https://github.com/alexanderwallin/node-gettext)
**can parse PO/MO**, similar to Python's gettext (based on GNU gettext)
  - [i18next](https://github.com/i18next/i18next/) + [i18next-gettext-converter](https://github.com/i18next/i18next-gettext-converter)
JSON Based
  - [adrienrn/jsgettext](https://github.com/adrienrn/jsgettext)
  - ~[frascata/gettext-js](https://github.com/frascata/gettext-js)~

  ~**_OR_**~
  ~Write a method to get strings through Python in JSON format~
  ~and use and function `_(message)` to replace text.~
- [ ] Mark strings in `*.js` files for translation,
try to re-use strings as much as possible.
- [x] Extract messages
- [x] Commit source `messages.pot` and merge PR


SickRage2:

Thanks!

miigotu:
I think https://github.com/adrienrn/jsgettext is the best choice.

miigotu:
Merging this now, its large and it doesnt break anything right now so we can continue in another PR

sharkykh:
There's a bower package named `underscore` so we won't be able to use <kbd>_</kbd>.
`underscore` is used for two [`_.debounce`](http://underscorejs.org/#debounce) calls:
```
Search "_." (2 hits in 1 file)
  C:\Github\SickRage-Dev\SickRage\gui\slick\js\core.js (2 hits)
	Line 2126:  $('#filterShowName').on('input', _.debounce(function() {
	Line 3348:  $('#min_level,#log_filter,#log_search').on('keyup change', _.debounce(function () {
```
and a TODO about using it more [(Line 1587)](https://github.com/SickRage/SickRage/blob/f7e67c0f54789ef8fcbd26b38c897ba3c37b1bf2/gui/slick/js/core.js#L1587-L1588)
```
// @TODO We might be able to change these from typewatch to _ debounce like we've done on the log page
//       The main reason for doing this would be to use only open source stuff that's still being maintained
```

Still, I don't see a reason why we would use `underscore` for only **_2-6_** calls.
But if we're not going to stop using `underscore`,
we can use <kbd>__</kbd> for gettext.

sharkykh:
@miigotu 
`underscore` wrapper might cause an issue with the strings you wrapped in <kbd>\_\(\*\)</kbd> in core.js.
This is on Google Chrome DevTools console
```javascript
> _
function (a){return a instanceof s?a:this instanceof s?void(this._wrapped=a):new s(a)}

> _('Hello World')
s {_wrapped: "Hello World"}
```

Just had another idea,
For the time being we can do this:
```javascript
__ = _ // Move underscore to double-underscore
_ = function(str) { return str; } // Placeholder to allow babel to extract strings for now,
                                  // this will ultimately wrap str with gettext.
```

PR #3376, opened 2017-03-10
Fix todo in config.py

diondehoog:
Fixes TODOs in sickbeard/config.py

- All the frequency functions in config.py now return True or False when the function succeeds or fails
- All the change functions in config.py now return True or False when the function succeeds or fails
I first made a branch based on development of the sickrage/sickrage git repository, but I'm not allowed to push on this one (I am not a smart man). So I a new branch on the forked version hence the reason there are so few commits.


SickRage2:

Thanks!

diondehoog:
Okay, so I tried to pull request to merge it into master instead of development... it is now a request to pull it into development, I think it's time for me to take a break hehe

sickbeard/config.py
```diff
@@ -392,146 +419,196 @@ def change_showupdate_hour(freq):
...
+            return True
+
+        sickbeard.VERSION_NOTIFY = version_notify
+        if sickbeard.VERSION_NOTIFY:
+            if not sickbeard.versionCheckScheduler.enable:
+                logger.log("Starting VERSIONCHECK thread", logger.INFO)
+                sickbeard.versionCheckScheduler.silent = False
+                sickbeard.versionCheckScheduler.enable = True
+                sickbeard.versionCheckScheduler.forceRun()
+                return True
```
sharkykh: I think it should also return `True` even if the version check thread **is already enabled**.
***This applies to most of the functions.***

sickbeard/config.py
```diff
@@ -331,16 +331,22 @@ def change_unpack_dir(unpack_dir):
 def change_postprocessor_frequency(freq):
     """
     Change frequency of automatic postprocessing thread
-    TODO: Make all thread frequency changers in config.py return True/False status
 
     :param freq: New frequency
     """
-    sickbeard.AUTOPOSTPROCESSOR_FREQUENCY = try_int(freq, sickbeard.DEFAULT_AUTOPOSTPROCESSOR_FREQUENCY)
+    try:
+        sickbeard.AUTOPOSTPROCESSOR_FREQUENCY = try_int(freq, sickbeard.DEFAULT_AUTOPOSTPROCESSOR_FREQUENCY)
```
sharkykh: This can be outside the `try`.
(applies to many functions)

sickbeard/config.py
```diff
@@ -331,16 +331,22 @@ def change_unpack_dir(unpack_dir):
...
 
-    if sickbeard.AUTOPOSTPROCESSOR_FREQUENCY < sickbeard.MIN_AUTOPOSTPROCESSOR_FREQUENCY:
-        sickbeard.AUTOPOSTPROCESSOR_FREQUENCY = sickbeard.MIN_AUTOPOSTPROCESSOR_FREQUENCY
+        if sickbeard.AUTOPOSTPROCESSOR_FREQUENCY < sickbeard.MIN_AUTOPOSTPROCESSOR_FREQUENCY:
+            sickbeard.AUTOPOSTPROCESSOR_FREQUENCY = sickbeard.MIN_AUTOPOSTPROCESSOR_FREQUENCY
```
sharkykh: This `if` statement can also be outside the `try`.
(applies to many functions)

sickbeard/config.py
```diff
@@ -331,16 +331,22 @@ def change_unpack_dir(unpack_dir):
...
 
-    sickbeard.autoPostProcessorScheduler.cycleTime = datetime.timedelta(minutes=sickbeard.AUTOPOSTPROCESSOR_FREQUENCY)
+        sickbeard.autoPostProcessorScheduler.cycleTime = datetime.timedelta(minutes=sickbeard.AUTOPOSTPROCESSOR_FREQUENCY)
+        return True
+    
+    except:
+        e = sys.exc_info()[0]
+        logger.log("Exception Caught: %s".format(e), logger.DEBUG)
```
sharkykh: Replace `%s` with `{0}` or `{}`.
(applies to many functions)

sharkykh (changes requested):
Hi, and thanks for your contribution!

I have a few general comments:
- I see you already tried to fix that on your own :smile:, however the current code wouldn't work,
as the `%s` is supposed to be a `{0}` or `{}` as explained [briefly here](https://www.quantifiedcode.com/app/issue_class/4ACGxFj1).
~As a rule of thumb - it's better to use `string.format()` over string interpolation, as explained [briefly here](https://www.quantifiedcode.com/app/issue_class/4ACGxFj1). Please update your `logger.log()` calls.~
- ***In my opinion***, `try_int` and `checkbox_to_value` don't need to be inside the `try` statement.
- Many of the `if` statements that are straightforward also don't need to be inside the `try` statements.
For example, `if sickbeard.DAILYSEARCH_FREQUENCY < sickbeard.MIN_DAILYSEARCH_FREQUENCY:` won't throw an exception.
- Also see my inline comment about the `return True`.

***All of these apply to many of the functions.***

diondehoog:
Hi, thanks for the feedback! will work on it as soon as I have some time again :) (this evening or tomorrow)

sickbeard/config.py
miigotu: Naked exceptions without explicit exception type is a PEP violation, and afaik nothing here will throw an exception anyways.

sharkykh: I was thinking about that, and hoping you would notice :smile:
(because I didn't know if any of it will throw one)

So remove the `try` altogether?

miigotu:
@sharkykh @h3llrais3r
PR's from delftswa2017 repo are from a group of Computer Science students in a Dutch technical school.  They have chosen SickRage as the target of an assignment to pick an open source project on GitHub to contribute to over 5 weeks. Very cool, yeah?

@diondehoog - If we are a little picky about the work you and your group members do please don't be discouraged. We will point out as much as we can because we hope to help you gain something out of this assignment. Glad to have you all.

diondehoog:
@miigotu Don't worry about it! All feedback is welcome and we see it as a positive thing to improve ourselves :)

sickbeard/config.py
diondehoog: Hey, I'm pretty new to python so I might make some rookie mistakes hehe.
the TODOs said it should return true if it succeeds and false if it fails, but I don't really immediately see a way in which some of these functions can fail. Can you maybe give me some pointers on how to improve this? :)

miigotu: I'm not sure who/when that TODO was added, it might be very old.

If they can't fail/throw just have them return True.
Wrapping everything in try/except is an approach called EAFP,  but we usually only apply that when something can fail and in finicky areas.

When it can't fail it just adds unnecessary instructions in the bytecode and uses more CPU cycles.  For these methods it makes a negligible difference, but if it were a method that could be called thousands of times an hour it would be a major waste of resources =P

sickbeard/config.py
```diff
@@ -433,105 +436,123 @@ def change_version_notify(version_notify):
...
-
+        
     if sickbeard.DOWNLOAD_PROPERS == download_propers:
-        return
+        return True
 
-    sickbeard.DOWNLOAD_PROPERS = download_propers
-    if sickbeard.DOWNLOAD_PROPERS:
+     sickbeard.DOWNLOAD_PROPERS = download_propers
+     if sickbeard.DOWNLOAD_PROPERS:
```
sharkykh: Excess indentation (this also what caused the build to fail)

sickbeard/config.py
```diff
@@ -433,105 +436,123 @@ def change_version_notify(version_notify):
...
             logger.log("Starting TRAKTCHECKER thread", logger.INFO)
             sickbeard.traktCheckerScheduler.silent = False
             sickbeard.traktCheckerScheduler.enable = True
+        return True
+
     else:
         sickbeard.traktCheckerScheduler.enable = False
         sickbeard.traktCheckerScheduler.silent = True
         logger.log("Stopping TRAKTCHECKER thread", logger.INFO)
+        return True
```
sharkykh: We can actually just return True at the end if the function instead of returning inside each if/else,
as this will yield the same result.

sickbeard/config.py
```diff
@@ -433,105 +436,123 @@ def change_version_notify(version_notify):
...
+        else:
+            sickbeard.subtitlesFinderScheduler.enable = False
+            sickbeard.subtitlesFinderScheduler.silent = True
+            logger.log("Stopping SUBTITLESFINDER thread", logger.DEBUG)
+            return True
+
+    except:
+        e = sys.exc_info()[0]
+        logger.log("Exception Caught: {}".format(e), logger.INFO)
+        return False
```
sharkykh: Forgot two try-except statements...
This is one

sickbeard/config.py
```diff
@@ -433,105 +436,123 @@ def change_version_notify(version_notify):
...
+        else:
+            logger.log("Stopping POSTPROCESSOR thread", logger.INFO)
+            sickbeard.autoPostProcessorScheduler.enable = False
+            sickbeard.autoPostProcessorScheduler.silent = True
+            return True
+
+    except:
+        e = sys.exc_info()[0]
+        logger.log("Exception Caught: {}".format(e), logger.DEBUG)
+        return False
```
sharkykh: This is two.

sharkykh (changes requested):
You're almost there! :)
It also seems like you did very much and then deleted, but that wasn't your fault, as those TODOs might have been outdated...

sharkykh (approved):

diondehoog:
Thanks for all the help and feedback guys! Next time everything should go a bit faster haha 👍

sharkykh:
Looking forward to it :)

PR #3402, opened 2017-03-15
Further locale support for Javascript files

sharkykh:
Continues #3367 

The initial commit to this PR includes **[adrienrn/jsgettext](https://github.com/adrienrn/jsgettext)**,
and my way of implementing it into the code.

Notes:
1. Right now it doesn't support `sickbeard.GUI_LANG` being None. I don't think I even tested it.
2. In my first attempt to serve the locale files I tried to create a Tornado "redirect" from (regex)
`/locale/([a-z]{2}_[A-Z]{2})\.([pm]o)`
  to `/locale/\1/LC_MESSAGES/messages.\2`
  So the url would be as clean as possible,
  but I was having a hard time writing in a new handler that does that to work...
3. **Unfortunately,** jsgettext doesn't support `.mo` files, only `.po` and `.json`.
Apparently support for `.mo` **is planned**, however it seems like the repo was last updated on Jun 19, 2016, so it might take a while.
4. [This line](https://github.com/SickRage/SickRage/pull/3402/commits/cba6f805b1b35554fad3dd0105160f790b8dd7ee#diff-7a62d0eb6a8e5fb4f863e112ea1920b7R98) also creates a weird warning in the developer console. AFAIK it doesn't cause any issues, just annoying.


SickRage2:

Thanks!

sharkykh:
Another discussion point is, we should consider using json translations instead of po.
* It may be a more elegant solution
* It's probably faster (and it's also recommended) to parse json rather than to read a po file
* It would probably get rid of that warning (see 4. in OP)

**Converting po to json in gettext format:**
This part is a bit lacking in options...
There is a module for [python (pojson)](https://pypi.python.org/pypi/pojson) but it's optimised for another format that we can't really use (i18next),
and I also saw several npm modules.
~If it's possible to run it as part of the `compile_catalog` command on setup.py,
then we won't need to run extra tasks/commands when updating translations.~

miigotu:
json is the best format to use I think. Its the most efficient in js, and the type is recognized
We can add a grunt task to:

__most of the process is here__
```bash
setup.py extract_messages
setup.py update_catalog
if CROWDIN_API_KEY env var is set:
    crowdin-cli-py upload sources
    crowdin-cli-py download
setup.py compile_catalog
bower_components/jsgettext/bin/po2json on all of the po files
```

I also want to add a grunt task that automates my changelog updates, maybe even handle the entire release process I do
```bash
git checkout develop
git pull
git checkout master
git merge develop
git log --oneline (latest tag)..HEAD | sed 's/^........//' | xclip -selection c
git tag -s (date-num)
paste+save
git push origin master --tags
./genchanges.sh > ../sickrage.github.io/sickrage-news/CHANGES.md
cd ../sickrage.github.io
git commit -asm 'Update changelog'
git push origin master
```

sharkykh:
I agree about the efficiency.
We can also "compile" it like po->mo by running uglify on the json (I think).

The `jsgettext/bin/po2json` script is written in **Perl**. ( 😞 )
I downloaded Perl and tested it.
It worked, kind of.
1. It couldn't read the locale header / plural forms for some reason,
which is not ideal and may cause issues in the long run.
2. I tested it on he_IL (Hebrew), and it looks like it's not using utf-8
even though the file is encoded in utf-8.
```javascript
      "Unknown" : [
         null,
         "×œ× ×™×“×•×¢"
      ],
```

I failed to find a compatible Python script after I realized `pojson` isn't (as I said in the last comment).
But now that I know `grunt` can pretty much run any task we'd like (using [`grunt-shell`](https://github.com/sindresorhus/grunt-shell) for example), so maybe there's a Node.js module that's compatible.
**_I'll probably look into that later today._**

_____

### On an unrelated note,
I noticed most of the time when grunt recreates .min.js files,
they are different from the latest repository version, even if you updated moments ago.

**In other words:**
 * You run grunt and push vender.min.js + core.min.js,
 * I pull the latest commit, run grunt again
 * My newly created minified js files **are different** for no apparent reason
(not a line endings issue, I use LF)

My older brother suggested we'd switch to yarn instead of npm.
It's much faster and also it checks all of the modules' dependencies,
and he thinks that might be what's causing the difference.

The change is we would need to commit a yarn.lock which should be commited every time it changes,
and use some different commands of course. [yarn : migrating from npm](https://yarnpkg.com/en/docs/migrating-from-npm)
(didn't actually read this, my brother taught me the basics of yarn)

What do you think?
_________

### Sorry for the long post, here's a potato! 🥔 haha

sharkykh:
@miigotu  Just wanted to let you know I'm making very good progress with this right now.
After a couple of hours trying to find a better po-2-json script, the solution just hit me!
You'll see when I finish :smile:

sharkykh:
Added the new grunt task: `grunt update_trans`.
It will execute the commands you asked for, and it will also sync with crowdin.com when you have `CROWDIN_API_KEY` variable set.
You should test it with your setup and verify it works.

And now gettext strings are loaded using ajax from messages.json.
Currently it uses `async: true` in order to avoid the deprecated warning.
I don't believe it will cause any major issues or any issues at all...

Anyway, I am not finished yet.

* **I still have to somehow get the user's system locale**
for when 'Use System Language' is selected in config.
I'm thinking of passing it down from python as a meta-var.
* (Done) ~And to create a better FileHandler to serve the
json files on /locale/en_US.json and handle "not found".~

sharkykh:
> I still have to somehow get the user's system locale (I'm thinking of passing it down from python) for when 'System Language' is selected in config.

Looks like system language in the config doesn't really do much - just sets it to a "null" translation so English is loaded.
My code does exactly that, too.
So this is ready to merge, in my mind.

PR #3404, opened 2017-03-15
Fix pid example for init.systemd

Nuky:
Fixes the with-PID example in `runscripts/init.systemd` to avoid rights problem when accessing pid file.

Explanation:

The python script must be able to write to `/var/run/sickrage/sickrage.pid`, so we must ask Systemd to create `/var/run/sickrage` with the correct owner and rights before running sickrage.
This is a classic problem and apparently the preferred method nowadays is to use `RuntimeDirectory` (see [systemd manpage about that option](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=)).

### DIFF
--- a/runscripts/init.systemd
+++ b/runscripts/init.systemd
@@ -23,6 +23,8 @@
 
 ### Example Using SickRage as daemon with pid file
 # Type=forking
+# RuntimeDirectory=sickrage
+# RuntimeDirectoryMode=0750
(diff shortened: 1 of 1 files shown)


SickRage2:

Thanks!

sharkykh:
Thank you again for your contribution!

Issue #3359, opened 2017-03-08
Problem with storage on a NAS under Windows


rperr:
### Before submitting your issue:

Enable debug logging in SickRage settings, reproduce the error (be sure to disable after the bug is fixed)

Branch/Commit: Latest Update
OS:  Windows 10
What you did: After the update, I clicked into one of my shows and saw that it said the folder was missing and size was 0.00 B even though the Show List shows several GB (I have checked several other shows and all show the same behavior). I tried refreshing, but no luck with that...

What happened: SickRage says the folder is missing and no files exist.  The show folder and files are definitely still there but appears that one of the recent commits may have removed the ability to specify the folder of a show using the '\\SEAGATE-D4\Media\TV Shows' format.  I use a NAS and have used the ''\\SEAGATE-D4\Media\TV Shows' as my root since I installed SickRage several months ago.  I also have the NAS mapped as my Z: drive and tried going into the show settings, but it does not show Z: in the list (think it was always like that which is why I used the \\SEAGATE... format).  If I should do it a different way, I'm more than happy to change the way I specify the show folders...  It also won't postprocess any newly downloaded shows.

What you expected: To be able to properly manage my library and postprocess files to move them to my NAS.

Logs:
```
SHOW REFRESH:
2017-03-07 21:46:33 INFO     SHOWQUEUE-FORCE-UPDATE :: The show dir \\SEAGATE-D4\Media\TV Shows\Lucifer is missing, not bothering to change the episode statuses since it'd probably be invalid

POST-PROCESSING:
2017-03-07 22:12:43 INFO     POSTPROCESSOR-AUTO :: Processing E:\Complete\Beavis.and.Butt.Head.S09E06.mkv (None)
2017-03-07 22:12:43 WARNING  POSTPROCESSOR-AUTO :: A path is required for the destination. Check the root dir and show locations are correct for School Test (I got '\\SEAGATE-D4\Media\TV Shows\Beavis and Butt-Head')
2017-03-07 22:12:43 WARNING  POSTPROCESSOR-AUTO :: Not enough space to continue PP, exiting
2017-03-07 22:12:43 WARNING  POSTPROCESSOR-AUTO :: Processing failed for E:\Complete\Beavis.and.Butt.Head.S09E06.mkv
```

SickRage2:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Include basic information: Branch/Commit, OS, What you did, What happened, What you expected
[...]

PR #3459, opened 2017-03-21
Custom CSS, FileBrowser improvements

sharkykh:
Requested in #3457 

Plus changed the FileBrowser so it would be easier to filter other types (extensions) of files in the future.
And changed it so folders would show up before files, like in a regular file browser.


SickRage2:

Thanks!

gui/slick/js/core.js
```diff
@@ -443,7 +443,8 @@ var SICKRAGE = {
         },
         index: function() {
             $('#log_dir').fileBrowser({title: 'Select log file folder location'});
-            $('#sickrage_background_path').fileBrowser({title: 'Select Background Image', key: 'sickrage_background_path', includeFiles: 1, imagesOnly: 1});
+            $('#sickrage_background_path').fileBrowser({title: 'Select Background Image', key: 'sickrage_background_path', includeFiles: 1, fileTypes: ['images']});
+            $('#custom_css_path').fileBrowser({title: 'Select CSS file', key: 'sickrage_background_path', includeFiles: 1, fileTypes: ['css']});
```
miigotu: Wrong key here?

miigotu (commented):
Not sure of the usefulness of the custom css. Will this be just another generally unused feature?  Mostly I try to avoid adding features unless they are expected to be widely appreciated/used, and are sure to not result in any extra issue reports.

In other words, we have lots of broken stuff to fix right now, I'm weary of creating more bugs.

Side note: I think this PR should add a [[shares]] domain to the config, with subdomains we can loop over and add to the file browser like we do in foldersAtPath when adding windows drives.

```config
[shares]
  [[nas]]
    user = miigotu
    password = sickrage
    server = HostNameOrIP
    share = sharedFolderName

  [[pi]]
    user = miigotu
    password = sickrage
    server = piHostNameOrIP
    share = sharedFolderName
```

Then in the foldersAtPath, something like this:

```python
for name, share in six.iteritems(sickbeard.CFG.get('shares')):
    if not (share.get('server') and share.get('share')):
        continue

    loc = ''
    if share.get('user'):
        loc += share.get('user')
        if share.get('password'):
            loc += ':{0}'.format(helpers.decrypt(share.get('password')))
        loc += '@'
    loc += '{0}\\{1}'.format(share.get('server'), share.get('share'))
    loc = '\\\\{0}'.format(loc)
    entries.append({'name': name, 'path': loc})
```

Ofc setting the values from a submitted form uses a simple process:
```python
sickbeard.CFG['shares'][name]['server'] = server
sickbeard.CFG['shares'][name]['share'] = share
sickbeard.CFG['shares'][name]['user'] = user
if password:
    sickbeard.CFG['shares'][name]['password'] = helpers.encrypt(password)
sickbeard.save_config()
```

gui/slick/js/core.js
sharkykh: Yep. Fixing

sickbeard/webserve.py
```diff
@@ -652,24 +652,32 @@ def sickrage_background(self):
                 return content.read()
         return None
 
+    def custom_css(self):
```
miigotu, line 655: btw, a route exactly like this is all you needed for the locale json lol, in the ui class in webserve, so a call to /ui/locale.json would return the file for GUI_LANG =P

sharkykh: I'll probably change it when I won't be so tired haha

sharkykh: english not good so right now
as you can see
:laughing:

sharkykh:
@miigotu  so far, so good.
Just need to adjust some settings like the patch target... maybe we should remove it altogether
`codecov/project = overall coverage change`
`codecov/patch = coverage of changed lines only`

sharkykh:
I'm adding 3 tests for the functions in browser.py so don't merge yet

miigotu:
You didn't answer my question about the point of this or, or input on the feature I'd like to add in this one =p

sharkykh:
Wow I didn't even see that comment. One sec

sharkykh:
> Not sure of the usefulness of the custom css. Will this be just another generally unused feature?  Mostly I try to avoid adding features unless they are expected to be widely appreciated/used, and are sure to not result in any extra issue reports.
> 
> In other words, we have lots of broken stuff to fix right now, I'm weary of creating more bugs.

Well, I'll be honest;
I was following a single user's request, and I just thought that it's a simple addition and I was in the mood, and so **I don't really have a position on the matter.**

> Side note: I think this PR should add a [[shares]] domain to the config, with subdomains we can loop over and add to the file browser like we do in foldersAtPath when adding windows drives.

Sounds very much like you want me to start working on https://github.com/SickRage/SickRage/issues/3359#issuecomment-285451899 😀 
I'll work on that tomorrow. I'm awake since 7am and it's 11:30pm here 😩

miigotu:
I just usually try my best to avoid new/extra config options that most people either won't use/need or understand.
When it comes to features we have a really large list of features, with votes on them to know how much the community wants them on http://feathub.com/SickRage/SickRage

It's not a big deal, we can allow this one because you are right, it's not much to implement it, but in general we should spend time on stuff that affects the highest number of users as possible... in a good way.

The post processing one is annoying but we need to finish that too x.X
As for the windows shares thing, I just think that will help a lot of people.
We could also possibly look through all local network IP's and call this for each one on startup https://github.com/miketeo/pysmb/blob/master/python2/smb/SMBConnection.py#L126 and append those to the list instead...
eg: connect to each ip and check for shares on the network

miigotu:
PS: Don't be bothered by my babble much, I too have barely been sleeping =P

m00n22-zz:
@miigotu I'm the one requested this feature, not knowing it would be picked up so quickly by @sharkykh . As stated in my original post #3457 I'm really enjoying Sickrage but don't like the current color scheme. Asfare as popularity on this feature I cannot comment besides the fact that I'll definitely use it. But this could lead to some nice css skins for SickRage people can share.

sharkykh:
@miigotu I started working on this and realized that it's much simpler to iterate through the settings when you set it up like a sub-section (that's what it's called by ConfigObj):

(`config.ini`)
```ini
[NetworkShares]
    [[NFS]]
        user = miigotu
        password = sickrage
        server = HostNameOrIP
        share = sharedFolderName
``` 
(`__init__`)
```python
        for share in CFG['NetworkShares']:
            NETWORK_SHARES[share] = {}
            for key in CFG['NetworkShares'][share]:
                NETWORK_SHARES[share][key] = check_setting_str(CFG['NetworkShares'], share, key)
```

then sickbeard.NETWORK_SHARES is:
```
NETWORK_SHARES = {dict}
    u'NFS' = {dict}
        u'password' = {unicode} u'sickrage'
        u'server' = {unicode} u'HostNameOrIP'
        u'share' = {unicode} u'sharedFolderName'
        u'user' = {unicode} u'miigotu'
```

**of course it will eventually store the password as encrypted**


Unfortunately , [UNC paths don't have the option to pass credentials like you suggested. which makes it more complicated](https://superuser.com/a/344933)
So I really do need to make the whole drive mapping for windows from the config to achieve this.

gui/slick/views/layouts/main.mako
```diff
@@ -94,11 +94,15 @@
...
-        %if sickbeard.THEME_NAME != "light":
+        % if sickbeard.THEME_NAME != "light":
             <link rel="stylesheet" type="text/css" href="${srRoot}/css/${sickbeard.THEME_NAME}.css?${sbPID}" />
-        %endif
+        % endif
 
         <%block name="css" />
+
+        % if sickbeard.CUSTOM_CSS:
+            <link rel="stylesheet" type="text/css" href="${srRoot}/ui/custom.css" />
```
OmgImAlexis, line 104: Why is the path hardcoded if there's a config option for `CUSTOM_CSS_PATH`?

miigotu: Webserve has a view for that route, which returns the true custom css

sickbeard/webserve.py
miigotu: @OmgImAlexis here

gui/slick/views/layouts/main.mako
OmgImAlexis: It'd make more sense to just serve it as a file so caching works.

miigotu: Did you try it and see if caching works?  If it doesnt then a header just needs fixed in the view.  Serving it as a file would require us to copy it into the gui folder, or somewhere, and add a static file handler for it.  For such a simple feature that is unlikely to be used I dont think putting too much work into it is warranted, unless the community starts making all kinds of css for SR. Then we can put all kinds of time into it. But for now this was just one user requesting this.

sharkykh: For the record, I was following the same concept as sickrage_background.

miigotu: No worries man, the custom.css is THE ONLY css that is caching correctly.
image

miigotu: We do need to add ?${srPID}(or whatever the pid var is) to it though so cache breaks on restart.

Issue #3399, opened 2017-03-15
Weird snatch result


Rouzax:
### SickRage Info
Branch: master
Commit: 826b084110bdef33f9a519686286a5d6f02e1bcc
Version: v2017.03.13-1
Database Version: 44.0
Python Version: 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
SSL Version: OpenSSL 1.0.2j 26 Sep 2016
OS: Windows-10-10.0.14393
Locale: nl_NL.cp1252

Just did a new search for one episode of The Americans, after having added an additional ignore word `LiGaS` as it had downloaded that one already. 
These are the 1080p results from the indexer for this episode.
![image](https://cloud.githubusercontent.com/assets/4103090/23939510/466f52fe-0961-11e7-8ef2-0c76b30c94ec.png)

It was my guess that SR would snatch the NTb episode because I have x265 as a general ignore word.
![image](https://cloud.githubusercontent.com/assets/4103090/23939564/9302df82-0961-11e7-8d8c-a2c9b248cdc4.png)

As you can guess SR did snatched `The Americans S05E01 1080p WEB-DL DD5 1 HEVC x265-RMTeam`

### Logs
http://pastebin.com/zxDgM0ME

SickRage2:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Include basic information: Branch/Commit, OS, What you did, What happened, What you expected
- [ ] Enable debug logging (be sure to disable after the bug is fixed)
- [x] Post debug logs, either inline (for smaller logs) or using gist

Please make sure you also read [how to create an issue](https://github.com/SickRage/sickrage-issues/blob/master/README.md) and followed all of the steps. 

The title should describe your issue. Having "SR not working" or "I get this bug" for 100 issues, isn't really helpful. We will close issues if there isn't enough information.

Sometimes the devs may seem like grunts and respond with short answers. This isn't (always) because the dev hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let us know, and this bot may get updated to automatically answer you.

Thanks!

sharkykh:
The reason this is not working for you is probably because you have set show-specific ignored words.

The way it works is - it checks if you have show specific ignored words and uses them - and only them - to ignore releases.
If you didn't set show specific ignored words, only then it removes releases based on globally ignored words.

[...]

PR #3507, opened 2017-03-28
Locale work + ignored/required words clarification

sharkykh:
- Simplify serving messages.json files ([as suggested in here](https://github.com/SickRage/SickRage/pull/3459/files/c9ff6a61ac91c2bb6e2cfa0fffcf33e71cc9a3de#r107497918))
- Extract some strings from javascript files
- Add a note to editShow clarifying that setting a show's ignored/required words overrides the globally ignored/required words. ( #3399 )

@miigotu I suggest you push a translation update too before merging.


SickRage2:

Thanks!

gui/slick/js/testRename.js
```diff
@@ -39,7 +39,7 @@ $(document).ready(function(){
         if(url.length < 2083) {
```
miigotu: we should change this to do a post someday soon

gui/slick/js/core.js
```diff
@@ -478,7 +476,7 @@ var SICKRAGE = {
                 growl.host = $.trim($('#growl_host').val());
                 growl.password = $.trim($('#growl_password').val());
                 if (!growl.host) {
-                    $('#testGrowl-result').html(_('Please fill out the necessary fields above.'));
+                    $('#testGrowl-result').html( _('Please fill out the necessary fields above.') );
```
miigotu, line 479: Why a space after ( and before ) on all of these?

miigotu (approved):

gui/slick/js/core.js
sharkykh: Umm felt to me like it's more readable

gui/slick/js/testRename.js
sharkykh: What do you mean?

miigotu: For rename, instead of calling it with a GET request with a huge list ?toRename=show1|show2|etc it should pass them in a post request, so we dont need to check the url length

sharkykh:
I've got one more small thing I'm working on

gui/slick/js/core.js
```diff
@@ -442,9 +440,9 @@ var SICKRAGE = {
             });
         },
         index: function() {
-            $('#log_dir').fileBrowser({title: 'Select log file folder location'});
+            $('#log_dir').fileBrowser({title: _('Select log file folder location')});
             $('#sickrage_background_path').fileBrowser({title: 'Select Background Image', key: 'sickrage_background_path', includeFiles: 1, fileTypes: ['images']});
```
MGaetan89: Why not translate this one?

sharkykh: Missed it, thanks!

gui/slick/js/testRename.js
miigotu: Im adding the rename form now for post

gui/slick/views/editShow.mako
```diff
@@ -290,6 +290,7 @@
                                         <div class="row">
                                             <div class="col-md-12">
                                                 <label>${_('search results with one or more words from this list will be ignored.')}</label>
+                                                <label><b>${_('note')}:</b> ${_('this option overrides the globally ignored words!')}</label>
```
miigotu, line 293: It actually works more like a bitflip, but thats hard to describe to users.

If a word is in global required but show ignored its now ignored, not required
The other global required/ignored are still required/ignored

If a word is in global ignored but show required its now required and is not ignored
The other global required/ignored are still required/ignored

sharkykh: I can't even wrap my head around that.
But yeah now I see [that part of the code](https://github.com/SickRage/SickRage/blob/3877fab198fa54a9773860251f8894a875f6cb98/sickbeard/show_name_helpers.py#L91-L100)...

So I could be wrong about my answer in #3399 ?

miigotu: Tbh I think I need to go back over that logic because it seems it doesnt work as intended sometimes.

Should be like (pseudo)
final_required = frozen_set(show_required) + (frozen_set(global_required).difference(show_ignored))
final_ignored = frozen_set(show_ignored) + (frozen_set(global_ignored).difference(show_required))

sharkykh:
Lol, I thought you wanted to change those notes...

gui/slick/views/editShow.mako
sharkykh: I agree that it might need a refactor,
but just so you know, my first thought was [why is this an `elif` and not an `if`?](https://github.com/SickRage/SickRage/blob/3877fab198fa54a9773860251f8894a875f6cb98/sickbeard/show_name_helpers.py#L91-L94)
This is where it seems to be ignoring globally ignored words if show ignored words are set.

You actually didn't say that it should do that in your first comment.
I believe it should, maybe as a pair of checkboxes per show -
`append to globally ignored words` and `append to globally required words` (both default to yes)

sharkykh: ~One more thing:
`ignored_words` becomes a `set` a few lines later,
so any duplicate ignored words will be purged, if I'm right.~

Edit: Just realized I got confused again because the `ignored` and `required` variable names are so similar.

miigotu:
We can change that anytime if needed. Will have to make some tests for that and make sure the logic works first, and maybe when editting show level ignored/required we can update a small infotip under it with what the final ignored/required calculates to with regard to the global setting

PR #3557, opened 2017-03-31
Auto grunt and update translations

sharkykh:
@miigotu 
In order to test it further, I need you to set the environment variable CROWDIN_API_KEY for Travis.

I could probably also use your help setting up the git remote
so we could actually push the updated translations (needs a token, or ssh keys).

I don't know if this is exactly what you meant in https://github.com/SickRage/SickRage/commit/1faad224a89a4c0ee588627fe65065e2f0d0ffc2#comments, but it's a start.


SickRage2:

Thanks!

miigotu:
You can call grunt in travis-ci.yml

```bash
- [ ! -z $CROWDIN_API_KEY ] &&  [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" != "master" ] && grunt update_trans
- [[ $(git status locale/ | sed '/^##/d') != "" ]] && grunt commit_and_push_trans
```

sharkykh:
True.

I needed a full script to help set it up first, though.

Also it's been such a long time since I've written anything in bash, so I mostly copied stuff.

miigotu:
Another thing, it should only be updating translations when a release is made ^ (push was to master)

sharkykh:
are you sure? doesn't that mean it would create another commit after you already pushed,

Now I'm thinking pushing will spawn another build...

miigotu:
Its ok that it creates a new commit, but the point is to make sure the newest translations are added on release.

miigotu:
Otherwise it would run and add a commit everytime someone pushes to a SickRage branch

sharkykh:
Ok, so I'll change this to update on master only once it's fully working.

sharkykh:
Can you add the Crowdin key please?

And I'm guessing you want @SickRage2 to push it,
so I also need a `BOT_GH_TOKEN` env var containing the account's personal token (and it should have privileges to push)

.travis.yml
```diff
@@ -17,8 +17,11 @@ before_install:
   - bower install
 
 install:
-    - pip install --upgrade pip
-    - pip install --upgrade tox
+  - pip install --upgrade pip
+  - pip install --upgrade tox
+  - if [[ ! -z "$CROWDIN_API_KEY" ]] && [[ "$TRAVIS_PULL_REQUEST" == "false" ]] && [[ "$TRAVIS_BRANCH" != "master" ]]; then
```
miigotu: You can just pip install here unconditionally

.travis.yml
```diff
@@ -33,6 +36,22 @@ cache:
...
+after_success:
+  - if ! { [[ ! -z "$CROWDIN_API_KEY" ]] && [[ "$TRAVIS_PULL_REQUEST" == "false" ]] && [[ "$TRAVIS_BRANCH" != "master" ]]; }; then exit 0; fi
+  - echo -e "Updating translations..."
+  - git config --global user.name "SickRage2"
+  - git config --global user.email sickrage2@gmail.com
+  - git config --global push.default simple
+  - git remote rm origin
+  - git remote add origin https://SickRage2:$BOT_GH_TOKEN@github.com/SickRage/SickRage.git
+  - grunt update_trans
+  - if [[ ! -z $(git diff-index HEAD -- locale/) ]]; then
```
miigotu: You should just check the git diff-index in the grunt task and exit there if nothing needs pushed. That will clean this up some, and I can use the commit_and_push_trans locally if I wanted also

miigotu (approved):
Small adjustments, if you don't mind =P

.travis.yml
sharkykh: The reason behind this is I thought it would be a bit faster if we skip it when not needed (which will be most of the times, if we'll only update translations on pushes to master)

.travis.yml
sharkykh: 👍  on it.

sharkykh:
I didn't change the `pip` line, I still think it's a good idea to only install those dependencies when they're needed..

.travis.yml
miigotu: The virtualenvironment is cached, so after the first time it wont take more than a second to check if its already installed.

sharkykh: Oh, okay then!

sharkykh:
Okay, [everything looks good so far](https://travis-ci.org/SickRage/SickRage/builds/217282508#L1792).

## TODO
- [ ] Add CROWDIN_API_KEY env var
- [x] Remove the  [fake CROWDIN_API_KEY from `.travis.yml`](https://github.com/SickRage/SickRage/blob/59914f5548b9161095d3b46a3be49f70c4bc9680/.travis.yml#L11-L11)

- [ ] Add GH_TOKEN env var and give SickRage2 write access for master.
A simpler way is to **use your username and put your token in the env var**

👍

sharkykh:
Oh, and of course
- [x] Change code to only update translations on pushes to master.

sharkykh:
Changed some things:

- Update translations only on pushes to master
- Use $GH_CRED env var instead of SickRage2:$GH_TOKEN,
Maybe it's better to not give any information to the public.
$GH_CRED should be set to `github_username:github_token`

- Changed git user.name to SickRage
- Removed the fake crowdin key.

sharkykh:
@miigotu I believe it's ready, only the push is untested.
Do you want to merge it in this state?

sharkykh:
I have another idea:
Shouldn't Travis push the new commit to both master & develop at the same time?
(To avoid having to FF develop to match the new updated master)

sharkykh:
Okay, so:
* New task for Travis: _**travis_update_and_push**_ (replaces commit_and_push_trans).
This task will run the default task and update translations.
If the files have changed it will commit the changes to both master and develop
(actually, to master and rebase develop), then push the branches.
* Fixed 'publish' task:
The task will now run the default task and update the translations **before** creating a new release and generating changes.
**When we decide to merge this, Travis should be able to do everything without any user interaction.**
Then, because you won't have to run grunt and update translations anymore -
I'll remove that part from the task.
* Use `git status -s` instead of `git diff-index HEAD` (simpler)
* Combine simple git commands (checkout, pull, merge, etc.) into one dynamic command,

----

In case it's not implied - the message of the commit below this comment is supposed to say:
>We need exit code 0 in commit_changed_files
So Travis doesn't deem the build as failed just because we didn't update *.min.js/*.min.css _**or locale files**_.

sharkykh:
I just found out that, apparently,
there's a way to debug nodejs/grunt from within PyCharm!

image

It would have been much easier to work with had I known about this!

Issue #3528, opened 2017-03-30
Impossible to listen on port 80


jeremyfritzen:
### Before submitting your issue:

Hi,

I installed Sickrage on new fresh installation of Debian.
It's working well on port 8080 but also on other ones (8081, 8082, etc.).

As I want to make the address more "easy to type", I want to setup the standard port to port 80. This way, I won't have to type manually "8080" in my browser.

But when I change the port to 80 (on Sickrage web interface or directly in the /opt/sickrage/config.ini file), the web interface is no more accessible (80 or 8080).
Log files say that the port 80 is already in use but the netstat command shows me that nothing is listening on port 80!

Do you have any idea?

Thank you very much!

Here are some information on my configuration:

OS: Debian jessie 8.7
Logs:
```
2017-03-30 22:14:21 INFO     TORNADO :: Starting SickRage on http://0.0.0.0:80/
2017-03-30 22:14:21 INFO     TORNADO :: Could not start webserver on port 80, already in use!
```
Net stat command "sudo netstat -tulpn |grep :80" doesn't show anything.

SickRage2:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Include basic information: Branch/Commit, OS, What you did, What happened, What you expected
- [ ] Enable debug logging (be sure to disable after the bug is fixed)
- [x] Post debug logs, either inline (for smaller logs) or using gist
- [x] Please wrap inline logs inside \`\`\` and \`\`\` for readability

Please make sure you also read [how to create an issue](https://github.com/SickRage/sickrage-issues/blob/master/README.md) and followed all of the steps. 

The title should describe your issue. Having "SR not working" or "I get this bug" for 100 issues, isn't really helpful. We will close issues if there isn't enough information.

Sometimes the devs may seem like grunts and respond with short answers. This isn't (always) because the dev hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let us know, and this bot may get updated to automatically answer you.

Thanks!

sharkykh:
I don't have *nix but on Windows it works fine for me, **as long as the port is not being used**.
Does the command you used show you anything without using `grep`?

NickMolloy:
Ports below 1024 can be opened only by root. SickRage is just catching all exceptions, and so doesn't distinguish between the actual reasons binding to a port fails.
[...]

Issue #3585, opened 2017-04-03
[APP SUBMITTED]: type object 'GithubException' has no attribute 'RateLimitExceededException'


BlackOnix80:
### INFO
Python Version: **2.7.13 (default, Mar 13 2017, 18:01:51) [GCC 4.9.2]**
Operating System: **Linux-3.16.0-4-amd64-x86_64-with-debian-9.0**
Locale: UTF-8
Branch: **master**
Commit: SickRage/SickRage@a1f5e6750493b00b032536c9a4274da57b5b9531
No Log available with ERRORS:
### ERROR
```
Thread-15 :: [a1f5e67] Failed doing webui callback: Traceback (most recent call last):
  File "/opt/sickrage/sickbeard/webserve.py", line 286, in async_call
    result = function(**kwargs)
  File "/opt/sickrage/sickbeard/webserve.py", line 5367, in submit_errors
    submitter_result, issue_id = logger.submit_errors()
  File "/opt/sickrage/sickbeard/logger.py", line 469, in submit_errors
    return Wrapper.instance.submit_errors(*args, **kwargs)
  File "/opt/sickrage/sickbeard/logger.py", line 414, in submit_errors
    except GhEx.RateLimitExceededException as ex:
AttributeError: type object 'GithubException' has no attribute 'RateLimitExceededException'

```
---
_STAFF NOTIFIED_: @SickRage/owners @SickRage/moderators

BlackOnix80:
Error you can close it.

PR #3587, opened 2017-04-03
Log the exception when starting the web server fails

NickMolloy:
Helps to solve confusion seen in #3528.

### DIFF
--- a/sickbeard/webserveInit.py
+++ b/sickbeard/webserveInit.py
@@ -148,11 +148,11 @@ class SRWebServer(threading.Thread):  # pylint: disable=too-many-instance-attrib
         try:
             self.server = self.app.listen(self.options['port'], self.options['host'], ssl_options=ssl_options,
                                           xheaders=sickbeard.HANDLE_REVERSE_PROXY, protocol=protocol)
-        except Exception:
+        except Exception as ex:
             if sickbeard.LAUNCH_BROWSER and not self.daemon:
                 sickbeard.launchBrowser('https' if sickbeard.ENABLE_HTTPS else 'http', self.options['port'], sickbeard.WEB_ROOT)
                 logger.log("Launching browser and exiting")
-            logger.log("Could not start webserver on port {0}, already in use!".format(self.options['port']))
+            logger.log("Could not start webserver on port {0}: {1}".format(self.options['port'], ex))
             os._exit(1)  # pylint: disable=protected-access
 
         try:
--- a/sickbeard/webserveInit.py
+++ b/sickbeard/webserveInit.py
@@ -4,6 +4,7 @@ from __future__ import print_function, unicode_literals
 
 import os
 import threading
+from socket import error as SocketError, errno
 
 from tornado.ioloop import IOLoop
 from tornado.routes import route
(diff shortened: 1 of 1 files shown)


sharkykh (commented):
I don't think Traceback is necessary in this case, should just display the error message.
```
15:46:59 INFO::TORNADO :: Starting SickRage on http://0.0.0.0:8082/
15:46:59 ERROR::TORNADO :: [1908114] Could not start webserver on port 8082: [Errno 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted
  File "C:\Github\SickRage-Dev\SickRage\sickbeard\webserveInit.py", line 150, in run
    xheaders=sickbeard.HANDLE_REVERSE_PROXY, protocol=protocol)
  File "C:\Github\SickRage-Dev\SickRage\lib\tornado\web.py", line 1850, in listen
    server.listen(port, address)
  File "C:\Github\SickRage-Dev\SickRage\lib\tornado\tcpserver.py", line 126, in listen
    sockets = bind_sockets(port, address=address)
  File "C:\Github\SickRage-Dev\SickRage\lib\tornado\netutil.py", line 194, in bind_sockets
    sock.bind(sockaddr)
  File "C:\Python27\lib\socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
error: [Errno 10048] Only one usage of each socket address (protocol/network address/port) is normally permitted
```

NickMolloy:
It is not possible to log to Error and not have it print a traceback, so it will have to be logged to info (or maybe warning).

sharkykh:
@NickMolloy I think that's fine, it was like this before too.

sharkykh:
I might be picky, but if I'm honest,
I think we should still have that nice "already in use!" one.
It's simple and people would understand it better.

```diff
--- a/sickbeard/webserveInit.py
+++ b/sickbeard/webserveInit.py
@@ -5,6 +5,8 @@ from __future__ import print_function, unicode_literals
 import os
 import threading
 
+from socket import error as SocketError
+
 from tornado.ioloop import IOLoop
 from tornado.routes import route
 from tornado.web import Application, RedirectHandler, StaticFileHandler
@@ -148,10 +150,16 @@ class SRWebServer(threading.Thread):  # pylint: disable=too-many-instance-attrib
         try:
             self.server = self.app.listen(self.options['port'], self.options['host'], ssl_options=ssl_options,
                                           xheaders=sickbeard.HANDLE_REVERSE_PROXY, protocol=protocol)
+        except SocketError as ex:
+            if ex[0] == 10048:  # "Only one usage of each socket address (protocol/network address/port) is normally permitted"
+                if sickbeard.LAUNCH_BROWSER and not self.daemon:
+                    sickbeard.launchBrowser('https' if sickbeard.ENABLE_HTTPS else 'http', self.options['port'], sickbeard.WEB_ROOT)
+                    logger.log("Launching browser and exiting")
+                logger.log("Could not start webserver on port {0}, already in use!".format(self.options['port']))
+            else:
+                logger.log("Could not start webserver on port {0}: {1}".format(self.options['port'], ex))
+            os._exit(1)  # pylint: disable=protected-access
         except Exception as ex:
-            if sickbeard.LAUNCH_BROWSER and not self.daemon:
-                sickbeard.launchBrowser('https' if sickbeard.ENABLE_HTTPS else 'http', self.options['port'], sickbeard.WEB_ROOT)
-                logger.log("Launching browser and exiting")
             logger.log("Could not start webserver on port {0}: {1}".format(self.options['port'], ex))
             os._exit(1)  # pylint: disable=protected-access
 

```

NickMolloy:
`if ex[0] == 10048` is not cross-platform.
For instance on Linux the equivalent error code is 98: `[Errno 98] Address already in use`.

sharkykh:
Thank you for pointing that out.

sharkykh (approved):
I think we should still wait for @miigotu's approval.

NickMolloy:
It seems instead of hardcoding the error code, [socket.errno](https://docs.python.org/2/library/errno.html#module-errno) can be used to do error code comparsions:

```
from socket import errno
...
except SocketError as ex:
    if ex.errno == errno.EADDRINUSE:
    ...
```

sharkykh:
That works for me, do you want me to push this?

NickMolloy:
Sure, sounds good.

miigotu (approved):

miigotu:
@sharkykh there is an as ex in retrySearchSubtitles

sharkykh:
@miigotu  Sorry, what?

sharkykh:
If you're talking about the GitHub Exception, this isn't the same thing. 
Because webserve.py doesn't import anything 'as ex'.

miigotu:
But it does import the ex function:
```python
from sickrage.helper.exceptions import CantRefreshShowException, CantUpdateShowException, ex
```

sharkykh:
So I don't know what caused the error in #3585

miigotu:
https://github.com/SickRage/SickRage/commit/6e875d4977deb0b0a5af0b7e5c505451d08c1a05

miigotu:
It was catching like `github.GithubException.RateLimitExceededException`, which doesnt exist

sharkykh:
Oh so I imported the wrong module
damn lol

BTW
I don't think you need this line too now, https://github.com/SickRage/SickRage/commit/6e875d4977deb0b0a5af0b7e5c505451d08c1a05#diff-1716b2a4b5a6b712365fc5f089022831R40

PR #3591, opened 2017-04-03
Add --force-update argument

sharkykh:
**Note:** I removed an old code from versionChecker.py regarding the old unrar library.
I think it's not needed anymore because we're using rarfile.

---

Add `--force-update` start argument that forces SickRage to update by downloading the master-tarball from Github (using the SourceUpdateManager class from versionCheck.py)
This will help resolve issues when SickRage doesn't load because of a breaking commit.

~This is intended for users that installed using source install, and SickRage directory isn't a git repo...~
This is intended for both git and source installation types.
If this is a git installation, it will try to use git first.
If updating using git failed, it will fallback to source update (download master-tarball from Github).

### Usage:
```
SickBeard.py --force-update
```

After it's done it will exit.
Then the user can load SickRage using his/hers normal start command.


SickBeard.py
```diff
@@ -173,6 +174,22 @@ def start(self):  # pylint: disable=too-many-branches,too-many-statements
 
         args = SickRageArgumentParser(sickbeard.PROG_DIR).parse_args()
 
+        if args.force_update:
+            if ek(os.path.isdir, ek(os.path.join, sickbeard.PROG_DIR, '.git')):
+                print('It is recommended to use Git to update SickRage.')
+                print('See https://github.com/SickRage/SickRage/wiki/FAQ\'s-and-Fixes#update-problems-try-this')
```
miigotu: We also support updating with a "source" install, which is a downloaded zip from github that is extracted into PROG_DIR.

You are calling the SourceUpdateManager here, you can also use the GitUpdateManager.

```python
            self.install_type = self.find_install_type()
            if self.install_type == 'git':
                self.updater = GitUpdateManager()
            elif self.install_type == 'source':
                self.updater = SourceUpdateManager()
```

miigotu (changes requested):

SickBeard.py
sharkykh: I know. But I think many of the stuff GitUpdateManager() needs to update are not initialized so early into the code.

I didn't want to add a messy code to SickBeard.py, so I compromised on this being a solution for Source installations only.

sharkykh: > ~We also support updating with a "source" install, which is a downloaded zip from github that is extracted into PROG_DIR.~

~If you hadn't noticed, I **am** using SourceUpdateManager.
It's GitUpdateManager that I'm not using.~

miigotu: Git updates shouldnt require anything other than git reset --hard and git pull.  The repo info is held in the .git/config anyways so git handles that all externally.

Its like 
```python
gh =  Github(sickbeard.PROG_DIR)
gh.reset()
gh.pull()
```
or something like that lol

sharkykh: I didn't go into the code again just yet,
but I don't understand - you still think GitUpdateManager would work?
It seemed to me like it uses config vars and Github to do all kinds of funky stuff.

**Or** do you think it's better to just use git commands in `SickBeard.py`..?

miigotu: Git commands will work just fine. force a reset/clean and pull and restart

sharkykh: I'll look into that. Sounds like a bitch to test this though. 😆

sharkykh:
Not sure if this is sufficient. It isn't fully tested yet.

sharkykh:
I just tested and it works 😄

Issue #3379, opened 2017-03-10
Exception when adding shows from trending lists


sharkykh:
When adding a show from trending(=Trakt) shows (popular=IMDB uses the same method),
and the show was deleted from TVDB, or the indexer ID is invalid, an exception is thrown.

Branch: trakt-images (**Shouldn't matter**)
Commit: 1697245f7a0438f4a81632f22cb0cc29aaa05d5c

Traceback
-------
```
2017-03-10 23:47:15 Thread-26 :: [1697245] Failed doing webui callback: Traceback (most recent call last):
File "C:\Github\SickRage-Dev\SickRage\sickbeard\webserve.py", line 282, in async_call
result = function(**kwargs)
File "C:\Github\SickRage-Dev\SickRage\sickbeard\webserve.py", line 2798, in addShowByID
show_name = get_showname_from_indexer(1, indexer_id)
File "C:\Github\SickRage-Dev\SickRage\sickbeard\helpers.py", line 1775, in get_showname_from_indexer
s = t[int(indexer_id)]
File "C:\Github\SickRage-Dev\SickRage\lib\tvdb_api\tvdb_api.py", line 946, in __getitem__
self._getShowData(key, self.config['language'], True)
File "C:\Github\SickRage-Dev\SickRage\lib\tvdb_api\tvdb_api.py", line 841, in _getShowData
self.config['url_seriesInfo'] % (sid, getShowInLanguage)
File "C:\Github\SickRage-Dev\SickRage\lib\tvdb_api\tvdb_api.py", line 629, in _getetsrc
raise tvdb_error(e)
tvdb_error: HTTP error None while loading URL http://thetvdb.com/api/F9C450E78D99172E/series/318852/en.xml
```

AJAX Response Preview (what the call to addShowByID returns)
-----
```
Error

HTTP error None while loading URL http://thetvdb.com/api/F9C450E78D99172E/series/318852/en.xml

Traceback

Traceback (most recent call last): 
File "C:\Github\SickRage-Dev\SickRage\lib\tornado\web.py", line 1469, in _execute result = yield result 
File "C:\Github\SickRage-Dev\SickRage\lib\tornado\gen.py", line 1015, in run value = future.result() 
File "C:\Github\SickRage-Dev\SickRage\lib\tornado\concurrent.py", line 237, in result raise_exc_info(self._exc_info) 
[...]

PR #3598, opened 2017-04-04
Fix error when adding show with invalid tvdbid

h3llrais3r:
Fixes https://github.com/SickRage/SickRage/issues/3379


sharkykh (approved):
Same solution as implemented in Trending shows.

sharkykh:
Maybe also run grunt and push it here. (I'm on mobile)

sickbeard/helpers.py
```diff
@@ -1790,16 +1791,21 @@ def tvdbid_from_remote_id(indexer_id, indexer):  # pylint:disable=too-many-retur
...
+        s = t[int(indexer_id)]
 
-    if hasattr(s, 'data'):
-        return s.data.get('seriesname')
+        if hasattr(s, 'data'):
+            return s.data.get('seriesname')
+    except (sickbeard.indexer_error, IOError) as e:
+            logger.log("Show id " + str(indexer_id) + " not found on " + sickbeard.indexerApi(indexer).name +
+                       ", not adding show: " + ex(e), logger.WARNING)
+            return None
```
sharkykh: This is indented too much (remove 4 spaces from all 3 lines)

sharkykh (changes requested):

sickbeard/helpers.py
h3llrais3r: Will do this evening.

h3llrais3r:
Why running grunt? 
I thought @miigotu does this when merging into master only?

sharkykh:
He uses an automated script we built together and it doesn't run grunt...
So I'm not sure.

sharkykh (approved):

h3llrais3r:
Ok, but shouldn't you run the script before merging into master only?
I thought @miigotu also only did the minimize of the js files when merging in master...

sharkykh:
In short, doesn't really matter.
I mean, you **can**, but the way PRs are merged here (squash), it looks better if the PR updates all of the necessary files. Also makes it easier to revert if it's ever needed.

---

**If you push grunt here,** when the PR gets merged into develop, everything is ready and updated for a new release. It's also cleaner this way, IMO, because pull requests are squashed into a single commit.

Because only miigotu can commit straight it to develop,
and like I said, the script miigotu uses doesn't do the grunt task just yet (because of reasons),
I think it's better to run grunt and push it to the PR before merging it.

**Also, think about this:**
Some people use develop but don't have developer enabled (granted, not many).
So when you push core.js without core.min.js,
core.min.js is still the old version, and that can cause issues.

sharkykh:
Only downside I can think of is it would create a merge conflict on core.min.js,
when you click "Update branch" on another PR that edits core.js.

h3llrais3r:
That's what I also thought 😄 
But I think we can't ask everyone to run the grunt script when creating PR's?

sharkykh:
I agree that it should be automated.
I'm working on [automatically updating the translations](https://github.com/SickRage/SickRage/pull/3557) first (on a push to master).
If that proves to be a good solution we'll make sure to run grunt too.

h3llrais3r:
@sharkykh @miigotu Can we merge?

h3llrais3r:
@sharkykh @miigotu Can we merge?

sharkykh:
I have already approved :)

sickbeard/webserve.py
```diff
@@ -2832,15 +2832,18 @@ def addShowByID(  # pylint: disable=unused-argument
         show_name = get_showname_from_indexer(1, indexer_id)
         show_dir = None
 
+        if not show_name:
+            ui.notifications.error(_('Unable to add show'))
+            return self.redirect('/home/')
+
```
sharkykh, line 2838: I just had a chance to actually test it, (was, or rather still, having some issues with my laptop)
and because it tried to add the show even when it didn't find it on TheTVDB,
some errors were logged, and a traceback was printed...

I think it's better to not try and add the show if it wasn't not found on TVDB.

h3llrais3r: Ok for me.

h3llrais3r: Can you merge?

sharkykh:
@h3llrais3r 
>But I think we can't ask everyone to run the grunt script when creating PR's?

https://github.com/SickRage/SickRage/commit/1d28b358576b35b101bb0f662858ae5296d2ef8b


@miigotu We won't have to once you add the needed environment variables:
`$GH_CRED` and `$CROWDIN_API_KEY`

Issue #3579, opened 2017-04-02
Error on tableSorter when adding a new show


beschoenen:
### Before submitting your issue:

Enable debug logging in SickRage settings, reproduce the error (be sure to disable after the bug is fixed)

Branch/Commit: 792a8637bfbd5adcb144d3b331cd3ebcc325fe8f
OS: Windows
What you did: Added a new show
What happened: JS died
What you expected: JS not to die
Logs:
```
core.js?1972:2229 Uncaught TypeError: Cannot read property 'toLowerCase' of undefined
    at 3 (core.js?1972:2229)
    at Object.getElementText (vender.min.js?1972:30)
    at Object.detectParserForColumn (vender.min.js?1972:30)
    at Object.setupParsers (vender.min.js?1972:30)
    at Object.setup (vender.min.js?1972:30)
    at HTMLTableElement.<anonymous> (vender.min.js?1972:31)
    at Function.each (vender.min.js?1972:1)
    at ea.fn.init.each (vender.min.js?1972:1)
    at ea.fn.init.b.fn.tablesorter (vender.min.js?1972:31)
    at Object.index (core.js?1972:2224)
```

SickRage2:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Include basic information: Branch/Commit, OS, What you did, What happened, What you expected
- [ ] Enable debug logging (be sure to disable after the bug is fixed)
- [ ] Post debug logs, either inline (for smaller logs) or using gist

Please make sure you also read [how to create an issue](https://github.com/SickRage/sickrage-issues/blob/master/README.md) and followed all of the steps. 

The title should describe your issue. Having "SR not working" or "I get this bug" for 100 issues, isn't really helpful. We will close issues if there isn't enough information.

Sometimes the devs may seem like grunts and respond with short answers. This isn't (always) because the dev hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let us know, and this bot may get updated to automatically answer you.

Thanks!

sharkykh:
I don't think need to do toLowerCase() at all.
https://mottie.github.io/tablesorter/docs/index.html#Configuration

Property | Type | Default | Description
--- | --- | --- | ---
ignoreCase | Boolean | true | When true, text sorting will ignore the character case. If false, upper case characters will sort before lower case. (v2.2).

h3llrais3r:
I think you're right. It can be removed everywhere the tablesorter is used.
[...]

PR #3601, opened 2017-04-05
Fix tablesorter

h3llrais3r:
Fixes tablesorter undefined error when calling toLowerCase() in certain cases. (Fixes https://github.com/SickRage/SickRage/issues/3579)
Fix sorting for quality.


miigotu:
You have to keep the lowercase.

sharkykh:
@miigotu I disagree: https://github.com/SickRage/SickRage/issues/3579#issuecomment-290999083

miigotu:
How does it not affect the filtering?

miigotu:
Also, ignoreCase is not set anywhere as far as I can tell.

miigotu:
Ok, seems you are right. ignoreCase became default in v2.2, filter_ignoreCase became default in 2.3.4
```
filter_ignoreCase	Boolean	true	Filter widget: Set this option to false to make the column content search case-sensitive, so typing in "a" will not find "Albert". (v2.3.4)
```

sharkykh:
Yep.
You figured it out by the time it took me to [find it](https://mottie.github.io/tablesorter/docs/example-widget-filter.html#options) using my phone... :)
I simply remembered that I checked it once some time ago and it worked out of the box.

miigotu:
Anyways, I will have to test because I know the search/replace for the quality extractor does not work on all of those.  We use the text because that is what people search by

sharkykh:
I just remembered that I checked it one time some time ago and it worked out of the box.

h3llrais3r:
@miigotu I fixed the search on quality :rofl: 
It's part of the PR because I saw it was not working on all tables.

OmgImAlexis:
This is just bound to create issues later on, use data attributes like the docs state.

h3llrais3r:
You're right, but this is just a quick fix for something that was not working as it should.

h3llrais3r:
@sharkykh @miigotu Can we still merge?

OmgImAlexis:
@h3llrais3r adding quick fixes is the reason sickrage ends up with so much broken code. If you're going do it you might as well just do it right from the start.

Issue #2526, opened 2016-10-23
GUI uses a lot of space and leaves sides empty / No horiz. Scrollbar / buttons not shown


udochrist:
Hi,

don't know if this is a Bug or enhancement:
Lately the GUI leaves large spaces on both sides of the browser window empty and i have to resize the sickrage window to use nearly all of my desktop to see all columns.

Esp when i'm looking at "Schedule" i don't even see the "Search" button to the right (and theres no horizontal scroll-bar at the bottom) unless i resize my browser window to at least 90% of my desktop width.

So: Can the gui please use - or at least shrink - the unused spaces left and right?

I've performed tests using Chrome on Windows 10 and macOs Sierra as well as the "native" Safari and Edge Browsers. The Behaviour is at least consistent.

I'm using the latest published version of SR.

Branch: master
Commit: 8bf53e8d03dc2b5bb946123dbaef2afb8667730c
Version: v2016.10.20-1
Database Version: 44.0

cheers
udo

SickRage2:
Thanks for the issue report! Before a real human comes by, please make sure your report has all the below criteria checked

- [ ] Include basic information: Branch/Commit, OS, What you did, What happened, What you expected
- [ ] Enable debug logging (be sure to disable after the bug is fixed)
- [ ] Post debug logs, either inline (for smaller logs) or using gist

Please make sure you also read [how to create an issue](https://github.com/SickRage/sickrage-issues/blob/master/README.md) and followed all of the steps. 

The title should describe your issue. Having "SR not working" or "I get this bug" for 100 issues, isn't really helpful. We will close issues if there isn't enough information.

Sometimes the devs may seem like grunts and respond with short answers. This isn't (always) because the dev hates you, but because he's on mobile or busy fixing bugs. If something isn't clear, please let us know, and this bot may get updated to automatically answer you.

Thanks!

beschoenen:
Can you give me some screenshots? I've had more complaints about this.

udochrist:
here you are. attached are screenshots with the browser occupying 75% and 90% of my 1080p wide screen.

latest sr/latest chrome/latest win10 fwiw.

75%
![screenshot_chrome_win10_75pct](https://cloud.githubusercontent.com/assets/869637/19627209/22889308-9942-11e6-8b57-db48a7d68856.PNG)

90%
[...]

PR #3604, opened 2017-04-06
Fix schedule table layout

h3llrais3r:
Fixes https://github.com/SickRage/SickRage/issues/2526


h3llrais3r:
@sharkykh @miigotu Can we merge?

sharkykh:
~I'm not seeing any difference on the page?~

~There's still a horizontal scrollbar.~
~https://cloud.githubusercontent.com/assets/10238474/24831127/9c016150-1c9c-11e7-95a8-931e0244c639.png~

---

Edit: Was looking at this the wrong way. I'll check again later.

sharkykh:
Still not seeing a change. It works fine for me as it is - without any changes [on a27ef187afda4399f39b076bfbbc4157e2fbd156]

h3llrais3r:
@sharkykh  you only have the problem when you have series with long names. Then you have an overflow and scrollbars are shown. With the fix, the long names are wrapped an no scrollbars are shown.

miigotu:
iirc making it not wrap was intentional.

h3llrais3r:
I think it's still better to wrap it and have no scrollbars...
But feel free to merge it or not... I never use the schedule anyway. :rofl:

miigotu:
I prefer to not have scroll bars too, I just cannot remember why they wanted it to not wrap

PR #3616, opened 2017-04-10
Snatched episodes in Schedule view

### LINKED ISSUE #3600: [Feature] Show snatched on schedule screen
When an episode has been snatched it disappears from the scheduled screen.

When that download fails, the only way to attempt to find the failure is to look under history and try and find episodes that don't have a future download status.

The scheduled screen should continue to show snached episodes until it has reached a state of downloaded. This should either be shown under "missing" or a maybe a new status.

This also should maybe have a timeout so that older snatched items are only visible for a number of days (perhaps 7 days, or a user configurable value).

psylenced:
Fixes #3600

- Schedule screen now shows "snatched" episodes
- Snatched episodes will only be displayed if they are new (location is empty) - ie. new snatches not update quality snatches
- Snatches will be removed once server updates status to download
- This uses the "Missed episodes range" setting to tell list how far back to find snatched episodes

This feature will hopefully easily identify any episodes that have been snatched and have failed or are currently in progress.

Apologies in advance for any code issues, it's my first time using python.


miigotu:
@sharkykh see any problems with this?

sharkykh:
I think I do...
I'll take a closer look later today or tomorrow.

sharkykh:
Right from the get-go:
I snatched an episode through the schedule screen, and now `gui/slick/js/ajaxEpSearch.js` keeps removing it from the list whenever I refresh.

Traced it to [this line](https://github.com/SickRage/SickRage/blob/7ba52ea56efd8ceb72e2fc23ba344c8599e75b0d/gui/slick/js/ajaxEpSearch.js#L97) but I don't know if it has any bad effect in other places.
It doesn't affect manual search in displayShow.

sickrage/show/ComingEpisodes.py
```diff
@@ -111,7 +113,11 @@ def get_coming_episodes(categories, sort, group, paused=sickbeard.COMING_EPS_DIS
             result[b'airs'] = str(result[b'airs']).replace('am', ' AM').replace('pm', ' PM').replace('  ', ' ')
             result[b'airdate'] = result[b'localtime'].toordinal()
 
-            if result[b'airdate'] < today:
+            if result[b'location'] and int(result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
+                continue
+            elif not result[b'location'] and int(result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
+                category = 'snatched'
```
sharkykh: I'd prefer:
```python
if int(result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
    if not result[b'location']:
        category = 'snatched'
    else:
        continue
```
This is one less `elif` and looks nicer IMO.

gui/slick/views/schedule.mako
```diff
@@ -115,8 +118,12 @@
                                     cur_ep_enddate = cur_result[b'localtime']
                                     if run_time:
                                         cur_ep_enddate += datetime.timedelta(minutes = run_time)
-
-                                    if cur_ep_enddate < today:
+                                    
+                                    if cur_result[b'location'] and int(cur_result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
+                                        continue
+                                    elif not cur_result[b'location'] and int(cur_result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
+                                        show_div = 'listing-snatched'
```
sharkykh: Same as above,
```python
if int(cur_result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
    if not cur_result[b'location']:
        show_div = 'listing-snatched'
    else:
        continue
```

----

You can even go simpler:
```python
if not cur_result[b'location'] and int(cur_result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
    show_div = 'listing-snatched'
```
However that might be a little bit slower, but I'm not too sure about that.

gui/slick/views/schedule.mako
```diff
@@ -222,6 +229,10 @@
                                             <% continue %>
                                         % endif
 
+                                        % if cur_result[b'location'] and int(cur_result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
```
sharkykh: Calendar view doesn't use the "listing-keys".
Instead we should just ignore all snatched episodes.
```python
% if int(cur_result[b'epstatus']) in Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST:
```

gui/slick/views/schedule.mako
```diff
@@ -318,7 +334,7 @@
                     <div>
                         % if cur_segment != cur_ep_airdate:
                             % if cur_ep_enddate < today and cur_ep_airdate != today.date() and not missed_header:
-                                <h2 class="day">${_('Missed')}</h2>
+                                <h2 class="day">${_('Missed')} / ${_('Snatched')}</h2>
```
sharkykh: I'm unable to test this...

And...
I'm not saying this is wrong, only odd - 
shouldn't "Snatched" get its own segment?

sharkykh (changes requested):
1. My comment about `gui/slick/js/ajaxEpSearch.js` still stands.
Need to make sure it only affects the Schedule view.
2. @miigotu [What about the API?](https://github.com/SickRage/SickRage/blob/599431d9c5a99ad57512c2e84c54ff15bb493a88/sickbeard/webapi.py#L709-L726)

psylenced:
I'll have a look at some of those code style changes later tonight and respond to the other queries.

From a quick look the ajax is only affecting schedule (as that's the only one with a id=forceUpdate attribute).

But this was just checked using github viewer, I'll have better access to the files once I'm home from work.

psylenced:
I've made the code style changes as suggested above.

Also while looking into the `ajaxEpSearch.js` I found a bug that was that pre-existing before my change. For Banner/Poster views, it was finding the wrong `tr` and removing half the panel (just the plot part) and not the full row, as those views have multiple `tr` elements per episode panel.

With the Missing / Snatched category part, that's a bit trickier when sorting by date. I had a look in to it and the problem with it is that episodes will be intermixed with different statuses. This also fails when a snatched episode is "today" as it will show under that category rather than snatched/missing.

example-poster-datesorting

The best thing I can think of is to alter the date sorting to be:
* Sort by snatched = true first
* Then sort by date

```
sorts = {
        'date': (lambda a, b: a[b'localtime'] < b[b'localtime']),
        'network': (lambda a, b: (a[b'network'], a[b'localtime']) < (b[b'network'], b[b'localtime'])),
        'show': (lambda a, b: (a[b'show_name'], a[b'localtime']) < (b[b'show_name'], b[b'localtime'])),
    }
```
That way all the snatched will be in the top panel (even if it's today's date), then missing will show in the second panel, then today, then days of week, then later.

The logic in that part of `schedule.mako` for panel + date sorting is a bit confusing so probably will need to be cleaned up a bit to use a variable containing each header type as it goes through rather than individual bool flags per type.

sharkykh:
I'll test/review it again soon.

If you feel like doing a code clean up in `schedule.mako`, then go for it.

psylenced:
I think once i get the sorting / rewrite of the grouping it should be ready to push in.

Main issue is it's my first time using python so need to get my head around how lambdas and the sorting works in it.

If you're able to give me a pointer on how to change this:

`'date': (lambda a, b: a[b'localtime'] < b[b'localtime']),`

To sort by `statusid = snatched`, then by `localtime`, that will be a massive help.

Then the change to `schedule.mako` should be fairly simple for me to implement.

sharkykh (dismissed):

sharkykh:
I now think that when you actually snatch an episode it should move it to the "Snatched" segment/section.
But I'm still not in the right mood for jump in the code and actually do it.
I'm not even sure if it's possible with the current setup...

psylenced:
Not sure that items should move sections. If you take all the other snatch actions (show screens, etc), they all just stay inline and change colour - so this is just following the same process until page is manually refreshed.

There is one last change that needs to be done for this feature to be complete, and that's the sorting for poster date view (as mentioned above). The snatched episodes need to be listed first and then the rest sorted by date order after that. Currently if there are inter-twined missing/snatched the headers will show up multiple times.

@sharkykh - Are you able to help with that one last bit?

sharkykh:
@chris-gh Can you check again? I can't seem to be able to reproduce this issue.

psylenced:
example-poster-datesorting2

Yep - still doing it. The fix it to somehow change:

```
sorts = {
        'date': (lambda a, b: a[b'localtime'] < b[b'localtime']),
        'network': (lambda a, b: (a[b'network'], a[b'localtime']) < (b[b'network'], b[b'localtime'])),
        'show': (lambda a, b: (a[b'show_name'], a[b'localtime']) < (b[b'show_name'], b[b'localtime'])),
    }
```
to:
```
        'date': (lambda a, b: (a[b'epstatus'] in Snatched[] ? 0 : 1, a[b'localtime']) < (b[b'epstatus'] in Snatched[] ? 0 : 1, b[b'localtime'])),
```

Does that make sense?

sharkykh:
OK, now I'm able to reproduce (after removing some episodes from my database..)
I'll try to make it work.

sharkykh:
@chris-gh I haven't had much luck so far...

psylenced:
@sharkykh just went to have another go at this myself and from the looks of it, the lambda sorting is not working at all.

I switched back to the master branch and scheduled poster view is always sorting by airdate, no matter what.

This would have been why I couldn't get mine working - I made assumption that the sorting code was working.

There is the following code:
```
        results.sort(key=itemgetter(b'localtime'))
        results.sort(ComingEpisodes.sorts[sort])
```
From the looks, the top sort is working by sorting airdate (for all cases), and then the second re-sort does nothing.

I'll have a look into the second sorting part and see if I can work it out - once that's done I'm guessing it'll just be a simple case of adding the second (snatched) sort field in and it'll work.

psylenced:
Fixed it - went back through old versions of the file and found that it previously was using cmp() in the lambda statements which sorted correctly. I've put that through in a separate commit (7491297).

With the final commit (4f43760) I think this issue is now complete and ready to go back in, once it's had a final review.

sharkykh:
@chris-gh 
Tried that now. It works well for me.

sharkykh:
@chris-gh Your code gave me an idea. Please test my latest commit.
It's a cleaner and more readable code. Used itemgetter and `key=` instead of `cmp=` (default for `list.sort()`) which is an old technique.
I **can't** confirm that it keeps the snatched and missed together.
If they mash together, it's easy to add more comparisons. (I commented out the other solution)

psylenced:
Can confirm your new way of sorting worked.

Just need to re-add my code back in. Would you like me to do that?

sharkykh:
@chris-gh I'll do that, I have some more small changes.
You mean we still need to sort using `snatchedsort`, correct?

psylenced:
Yes, still need to sort by snatchedsort. 

By uncommenting locally the 2 lines including snatched sort (the sorter and the setter) everything is working 100% from my testing.

sharkykh (approved):

sharkykh:
I can confirm that this is indeed working as expected now :smile: 

@miigotu Got any notes on this before I merge?

sharkykh:
@chris-gh 
Doesn't look so good on my phone (Galaxy S5).
<table>
<thead>
<td align="center"><b>Before</b></td>
<td align="center"><b>After</b></td>
</thead>
<tr>
<td></td>
<td></td>
</tr>
</table>

I'll try to fix this but I don't remember web design that well.


**Edit:**

image

PR #3630, opened 2017-04-15
Fix travis:update task

sharkykh:
- Incorrect syntax in grunt.run.task()
- Because Travis uses Git v1.8.x, I needed to remove '&&',
  so I split the 'git add' and 'git commit' commands.
- `git checkout branch` didn't work because Travis checks out the commit hash,
  fixed it by adding `git fetch`.
- It's a good thing I tested it again...
because **git credentials were exposed ~if `git push` failed~ on git push**
It will now censor the credentials:
```
Running "exec:git_push:test:develop" (exec) task
>> Exited with code: 128.
remote: Invalid username or password.
fatal: Authentication failed for 'https://[censored]@github.com/SickRage/SickRage.git/'
Warning: Task "exec:git_push:test:develop" failed. Use --force to continue.

Aborted due to warnings.
```
```
Running "exec:git_push:origin:master develop" (exec) task
To https://[censored]@github.com/sharkykh/SickRage.git
   5a74b96..d31ca27  develop -> develop
   5a74b96..d31ca27  master -> master
```
- Change task name from `travis` to `ci`.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -311,11 +311,11 @@ module.exports = function(grunt) {
                     }
                     return 'git add -- ' + paths;
                 },
-				callback: function(err) {
-					if (!err) {
-						grunt.task.run('exec:git:commit:-m "' + grunt.config('commit_msg') + '"');
-					}
-				}
+                callback: function(err) {
+                    if (!err) {
+                        grunt.task.run('exec:git:commit:-m "' + grunt.config('commit_msg') + '"');
+                    }
+                }
             },
             'git_get_last_tag': {
                 cmd: 'git for-each-ref --sort=-refname --count=1 --format "%(refname:short)" refs/tags/v20[0-9][0-9]*',
@@ -360,6 +360,14 @@ module.exports = function(grunt) {
                         pushCmd += ' --dry-run';
                     }
                     return pushCmd;
+                },
+                stderr: false,
+                callback: function(err, stdout, stderr) {
+                    if (err && err.code == 128 || stderr.includes('Authentication failed for')) {
+                        grunt.fatal('Git remote config contains invalid credentials.');
+                    } else {
+						grunt.fatal(stderr);
+					}
                 }
             },
             'git_list_tags': {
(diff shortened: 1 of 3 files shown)


OmgImAlexis:
@sharkykh is there really any point in censoring the username they use since it'll be on Github with their push/PR anyways?

sharkykh:
@OmgImAlexis Not really, no.
But,
I already decided to use one environment variable instead of two ($GH_USER, $GH_TOKEN),
**and** making it censor only the token (by matching the colon in $GH_CRED) is really not necessary...
And on top of that, @miigotu's the only one who would actually ~be able to see~ know the value,
so I figured, why not keep it censored everywhere.

btw, it will only do this on a push to master. Which is also, usually, @miigotu that does that.
And the username/email that will appear on GitHub/Git would be @SickRage2's

sharkykh:
There is still a small issue that I can't fix right now (tried a few times) because I don't have the right tools,
The commit message has quotation marks around it.

I'll leave that for when I have some more time to mess with it.
**Other than that, it's working great now :smile:**

image

Gruntfile.js
```diff
@@ -313,7 +313,7 @@ module.exports = function(grunt) {
                 },
                 callback: function(err) {
                     if (!err) {
-                        grunt.task.run('exec:git:commit:-m "' + grunt.config('commit_msg') + '"');
+                        grunt.task.run('exec:git:commit:-m \'' + grunt.config('commit_msg') + '\'');
```
OmgImAlexis: Since you're using node just use back ticks for the string and then you can use both types of quotes.

sharkykh: Yeah I know about that, you can also use `${variable}` instead of concatenating.
But that's not the issue, that old Git seems to be the issue..

OmgImAlexis: I get that but using backticks would solve the issue of having to escape everything and would make it a lot easier to read.

sharkykh: Thanks, I'll probably change stuff like that in another branch I'm working on, `SickRage:CI-s` (should've actually named it "Automation" or something like that)

sharkykh:
Found a way around it by writing a temp file, and using `git commit -F message-file.txt`
image

sharkykh:
Maybe I shouldn't link it to @SickRage2 (just remove the email address).
I only added it to get the icon to show.

PR #3655, opened 2017-04-24
Add maximum size

ArthurGarnier:
Fixes : http://feathub.com/SickRage/SickRage/+77

- Add field in "edit show" to set a maximum size in MB
- Default value is -1, means ignore the parameter


sickbeard/webapi.py
```diff
@@ -1904,6 +1904,7 @@ def run(self):
         else:
             show_dict["rls_ignore_words"] = []
 
+        show_dict["max_size_MB"] = show_obj.max_size_MB
```
OmgImAlexis: This should be max size in bytes. The only places MB should be used is when a human needs to see the value. Everything else should be translates back and forth to bytes.

sickbeard/webserve.py
```diff
@@ -1509,7 +1509,7 @@ def editShow(self, show=None, location=None, anyQualities=None, bestQualities=No
                  air_by_date=None, sports=None, dvdorder=None, indexerLang=None,
                  subtitles=None, subtitles_sr_metadata=None, rls_ignore_words=None, rls_require_words=None,
                  anime=None, blacklist=None, whitelist=None, scene=None,
-                 defaultEpStatus=None, quality_preset=None):
+                 defaultEpStatus=None, quality_preset=None, max_size_MB=None):
```
OmgImAlexis: Change to use bytes.

sickbeard/webserve.py
```diff
@@ -1650,6 +1650,7 @@ def editShow(self, show=None, location=None, anyQualities=None, bestQualities=No
                 show_obj.dvdorder = dvdorder
                 show_obj.rls_ignore_words = rls_ignore_words.strip()
                 show_obj.rls_require_words = rls_require_words.strip()
+                show_obj.max_size_MB = max_size_MB
```
OmgImAlexis: Change to use bytes.

sickbeard/tv.py
```diff
@@ -1165,7 +1169,8 @@ def saveToDB(self, forceSave=False):
                         "rls_ignore_words": self.rls_ignore_words,
                         "rls_require_words": self.rls_require_words,
                         "default_ep_status": self.default_ep_status,
-                        "sub_use_sr_metadata": self.subtitles_sr_metadata}
+                        "sub_use_sr_metadata": self.subtitles_sr_metadata,
+                        "max_size_MB": self.max_size_MB }
```
OmgImAlexis: Change to use bytes.

sickbeard/tv.py
```diff
@@ -823,6 +825,8 @@ def loadFromDB(self):  # pylint: disable=too-many-branches, too-many-statements
 
             self.subtitles_sr_metadata = int(sql_results[0][b"sub_use_sr_metadata"] or 0)
 
+            self.max_size_MB = sql_results[0][b"max_size_MB"]
```
OmgImAlexis: Change to use bytes.

sickbeard/tv.py
```diff
@@ -152,6 +153,7 @@ def __init__(self, indexer, indexerid, lang=""):
     rls_require_words = property(lambda self: self._rls_require_words, dirty_setter("_rls_require_words"))
     default_ep_status = property(lambda self: self._default_ep_status, dirty_setter("_default_ep_status"))
     subtitles_sr_metadata = property(lambda self: self._subtitles_sr_metadata, dirty_setter("_subtitles_sr_metadata"))
+    max_size_MB = property(lambda self: self._max_size_MB, dirty_setter("_max_size_MB"))
```
OmgImAlexis: Change to use bytes.

sickbeard/tv.py
```diff
@@ -110,6 +110,7 @@ def __init__(self, indexer, indexerid, lang=""):
         self._rls_ignore_words = ""
         self._rls_require_words = ""
         self._default_ep_status = SKIPPED
+        self._max_size_MB = -1
```
OmgImAlexis: Change to use bytes.

OmgImAlexis:
The update for the translations shouldn't be done in this PR. @sharkykh

ArthurGarnier:
Update done, just not sure if I have to do some conversion in webapi ?

sickbeard/webserve.py
```diff
@@ -1650,7 +1650,7 @@ def editShow(self, show=None, location=None, anyQualities=None, bestQualities=No
                 show_obj.dvdorder = dvdorder
                 show_obj.rls_ignore_words = rls_ignore_words.strip()
                 show_obj.rls_require_words = rls_require_words.strip()
-                show_obj.max_size_MB = max_size_MB
+                show_obj.max_size = str(int(max_size_MB)*1000*1000 if max_size_MB != -1 else -1)
```
OmgImAlexis: This is still using `max_size_MB`.

OmgImAlexis:
The API should be fine with returning it in bytes since a developer uses the api and can convert it back to MB if they would like it that way.

OmgImAlexis:
After reading through the feature request I'm not 100% sold on this being the right way to add it as this adds in a global size limit vs limits per quality setting meaning it doesn't really do what the feature request wanted.

Ideally you'd have user changeable size limits min/max for each quality so for example a `1080p WEB-DL` would be allowed to be from `1GB` - `5GB` where as a `SDTV` copy would only be allowed to go from `100MB` - `500MB`.

sickbeard/webserve.py
ArthurGarnier: Yes it's the value from the field of the UI, as you mentionned the final user have to enter a value in MB (or any other unit, but for the moment I just support MB)

ArthurGarnier:
I agree that, it's not exactly what the Feature Request asks, but this request is open for more than a year, so it's a good start I think. 
I can do this for each quality but it will take more time

OmgImAlexis:
@miigotu thoughts?

miigotu:
More time is fine, there is no rush on any db updates quite yet =P
Nearly there for the next db update, just not yet.

ArthurGarnier:
Just a question, it should be a set of parameter by show or a global one ?

beschoenen:
I'd like to have a minimum size, too. Eliminating the issue of download really bad 720p videos.
Is that something you could add in here? Or maybe another PR.

sharkykh:
@ArthurGarnier in my mind per show with default setting (think like the show settings)

miigotu:
Yeah, both global and per-show, with the show's setting overriding the global setting

ArthurGarnier:
@beschoenen Yes it could be added too, probably in the same PR, it's almost the same work

OmgImAlexis:
This really needs not only global and per show settings but per quality settings otherwise this only prevents the high res qualities downloading large files but doesn't really prevent things like a 5GB SD file vs a 5GB HDTV file.

ArthurGarnier:
Yes, 
So we will have : 2 tables in show settings and global settings : 
* Quality 1 : Min size | Max Size
* Quality 2 : Min size | Max Size
* ...

beschoenen:
The best solution, in my opinion, is to move qualities to a separate table.
This will also give us the ability to make custom qualities, with custom names, etc..

julienld:
@ArthurGarnier 
It also could have been a size relative to the length of the show : X MB per minute of the show. For example a 40 MB/min limit for a 20 minutes show would have an 600MB limit and a 40 minute show would have an 1.2GB limit... This would almost never require a per show setting to be set.

The number of minute per episode is in show.imdb_info['runtimes']. It's also available in the 'runtime' property of TheTVDB (but not saved right now... I think).

This is not perfect because some episodes are double feature and the runtime does not take this into account... but it still an issue with the current way to do it.

Also, i'm not sure this should have been implemented as 2 hard limit (low and high). I could have one setting : "Ideal size". From the ideal size, you could calculate a penalty for a too big of too small file.  The less ideal the file size is, bigger the penalty. This penalty could be applied when selecting the source for the episode.

As side effect, this would ensure a double-feature would still be downloaded because it's would not be an hard limit.

OmgImAlexis:
@julienld there would be a lot of false positives since things can change the torrent size like the codec used, if it has ads, trailers, etc., if it's compressed in a rar file, if there's a sample, etc..

mgaulton:
Curious if there's been any additional progress on this? I'd really like this particular feature.

ArthurGarnier:
Personally, didn't get any time for this. But I think I'll take few hours in next few weeks

YaannnTech:
I want to bump this up. I would love a file size limit for episodes, so that I can just tell SickRage to download files in "HD" and trust that whatever files it chooses to download don't exceed the individual file size limit I set (i.e. 2GB). Can this development be dusted off and integrated into the master?

sjdaws:
I have developed a branch which checks sizes for usenet however it currently uses arrays of static sizes based on quality as I don't know how to implement this into the UI.

If someone wants to run with it: https://github.com/SickRage/SickRage/compare/master...sjdaws:master

Update due to repo change: https://github.com/SickChill/SickChill/compare/master...sjdaws:master

rwyarbrough:
This functionality would be greatly appreciated in the community.  More and more files are being created with less compression. Some of them are enormous. Not having an option to set a min or max file size is resulting in them having to be discarded and smaller sized files manually added much more frequently of late which takes away the benefit of the automatic operation of SickBeard.

Tiwas:
*bump*

I have some other requests and takes on requests here in this thread: #4595

Nosirus (changes requested):
I need too, it downloads me otherwise seasons of 80Gb see more

miigotu:
I was going to use this as a base for the DB upgrade, but the patch/diff is gone since the repo is gone...

mgaulton:
@miigotu Is this a dead feature or are you just reevaluating?

miigotu:
@mgaulton I will be doing it myself, since this PR isnt able to be applied anymore

YaannnTech:
@miigotu Thank you -- we all appreciate it and are looking forward to this feature! :)

PR #3679, opened 2017-04-27
Fix the webcal protocol when running behind a reverse proxy

### LINKED ISSUE #1622: Link to calendar is broken if running behind a reverse proxy
For instance, when running behind an nginx proxy at `/sickrage` the subscribe link on the schedule page points to `webcal://mydomain.com:8081/calendar`, instead of `webcal://mydomain.com/sickrage/calendar`.

I think changing the this url to a relative one that uses `srRoot` ( [ like this](https://github.com/xxChinasaurxx/SickRage/commit/041dd14aa7171b9f09d1312656216dcf746fe062)) fixes this, but doing so means that the url no longer specifies the webcal protocol.

h3llrais3r:
Fixes #1622


gui/slick/views/schedule.mako
```diff
@@ -24,7 +24,7 @@
                     <span class="listing-key listing-default">${_('Soon')}</span>
                     <span class="listing-key listing-toofar">${_('Later')}</span>
                 % endif
-                <a class="btn btn-inline forceBacklog" href="webcal://${sbHost}:${sbHttpPort}/calendar">
+                <a class="btn btn-inline forceBacklog btn-cal-subscribe" href="${srRoot}/calendar">
```
OmgImAlexis: A better idea would be to add nothing in the href and then use js to add it that way there's never a broken link even before js loads.

NickMolloy:
Tested this, fixes the issue.

gui/slick/views/schedule.mako
sharkykh: @h3llrais3r seems easy enough - you have `srRoot` as a javascript var you can use.

OmgImAlexis: @sharkykh why would you even need that? Just get the current url and replace `/schedule` with `/calendar`.

sharkykh: @OmgImAlexis That is also an option. Didn't think of that :smile:

h3llrais3r: I can change it, but isn't it better to actually have a url in it?
If we keep my solution, the href stays `/calendar` even when the javascript is not loaded?
Or we can change it to `href="/calendar"` instead of `href="${srRoot}/calendar"`? Then we still have the output of `/calendar` in the browser instead of nothing?

OmgImAlexis: JS has to load before 99% of pages work anyways so there isn't much issue doing what I suggested.

h3llrais3r: I know... 😄 
New proposition:
```
<a class="btn btn-inline forceBacklog btn-cal-subscribe" href="#">
```
```
var calSubscribeLink = document.location.href.replace('http', 'webcal').replace('schedule', 'calendar');
$(".btn-cal-subscribe").attr("href", calSubscribeLink);
```

OmgImAlexis: Use this incase they're using https.
`window.location.href.replace(/https?/, 'webcal')`.

Looks good otherwise.

h3llrais3r: What about the empty href?
Standard practice is to use href="#" or sometimes href="javascript:;"

h3llrais3r: Nice remark, didn't think of https...

OmgImAlexis: Standard is to just omit completely it since we're expecting js to set it. Anchor elements don't need a `href` attr onload just when a user uses the element.

OmgImAlexis (approved):

sharkykh (approved):

h3llrais3r:
@NickMolloy can you do a final test again?

OmgImAlexis:
Looks good to me. The vender file shouldn't have been changed though? @sharkykh any ideas?

sharkykh:
@OmgImAlexis 
I thought about that but I removed node_modules then ran `npm install && grunt`.
Some of the versions in bower.json are set with ~ and <, maybe some updated versions were installed, as grunt removes bower_components and runs `bower install` from scratch.
I don't think there's an easy way to find out what exactly got updated, but you can be sure something did, like `tablesorter`.

OmgImAlexis:
All good then just don't really like seeing vender updates randomly in PRs ideally they should be in either separate commits and/or PRs.

sharkykh:
I'm following [miigotu's decision](https://github.com/SickRage/SickRage/commit/1d28b358576b35b101bb0f662858ae5296d2ef8b) for now, although we are set up for this task to be completely automated on push to master or whatever (missing API keys in configuration, that only miigotu has).

NickMolloy:
It no longer works, because a trailing slash is added, giving a 404.
This is because the [calendar handler regex](https://github.com/SickRage/SickRage/blob/4b3dad8eec3e238b1cadc809acf8e0e26deb9eb3/sickbeard/webserveInit.py#L102) doesn't accept an optional trailing slash, like the other handlers.

OmgImAlexis:
Let's add an optional ending slash to that handler then.

miigotu:
Working?

NickMolloy:
No, I get an exception:

```
03:38:46 ERROR::Uncaught exception GET /sickrage/calendar/ (***.***.***.***)
HTTPServerRequest(protocol=u'https', host='mydomain.com', method='GET', uri='/sickrage/calendar/', version='HTTP/1.0', remote_ip='***.***.***.***', headers={'Accept-Language': 'en-US,en;q=0.8', 'Accept-Encoding': 'gzip, deflate, sdch, br', 'X-Forwarded-For': '***.***.***.***', 'Host': 'mydomain.com', 'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8', 'Upgrade-Insecure-Requests': '1', 'Connection': 'close', 'Cookie': 'sickrage_user="**********"', 'Pragma': 'no-cache', 'Cache-Control': 'no-cache', 'X-Real-Ip': '***.***.***.***', 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/58.0.3029.81 Chrome/58.0.3029.81 Safari/537.36'})
  File "/opt/SickRage/lib/tornado/web.py", line 1467, in _execute
    result = method(*self.path_args, **self.path_kwargs)
TypeError: get() takes exactly 1 argument (2 given)
```

NickMolloy:
I think this is because the signature for `get` in **[CalendarHandler](https://github.com/SickRage/SickRage/blob/4b3dad8eec3e238b1cadc809acf8e0e26deb9eb3/sickbeard/webserve.py#L502)** should be ```def get(self, *args, **kwargs):```, but instead is only ```def get(self):```.

NickMolloy:
Just tested with the method signature changed and all is to be working. It seems the trailing slash is being passed as an arg to get.

sharkykh:
Thanks @NickMolloy , pushed a fix

PR #3686, opened 2017-04-28
Fix split-RAR files detection for "Treat as video" unpack option

### LINKED ISSUE #3674: "Treat as video" (#3271) doesn't process .r{00..99} files
Branch/Commit: master
OS: Ubuntu ARM64
What you did: Use the new "Treat as video" #3271 feature
What happened:
Sickrage correctly links `.rar` files, but doesn't take `.r00` - `r.99` with them. Kodi only plays the first few seconds.
What you expected:
Sickrage correctly links `.rar` files, but **does** take `.r00` - `r.99` with them. Kodi plays the whole episode.

sharkykh:
Should fix #3674 

See:
* http://i.imgur.com/yz0wxSR.png
* https://regex101.com/r/eMPi3p/9 (`only_first = False`)
* https://regex101.com/r/yM9BOp/3 (`only_first = True` default)


rubdos:
> https://regex101.com/r/eMPi3p/9 (only_first = False)
> https://regex101.com/r/yM9BOp/3 (only_first = True default)

Why is `only_first` per default ~~`False`~~`True`, shouldn't it be false? Most scene has the older style.

sharkykh:
@rubdos 
On the contrary, the default is `True`.

The reason for that is,
`is_rar_file` is used by the postprocessor to determine which files need to be extracted.
Without `only_first` (or when it's set to False), it would try to unpack/unrar every possible rar file:
`.rar, .part1.rar, .part9.rar, .r00, .r09`, etc.

I use `only_first` only for when unpack method is set to "Treat as video" and I need to process all of the rar files and parts to move them all together, like you requested.

So it has nothing to do with the actual style that is used in the downloaded release.

rubdos:
> On the contrary, the default is True.

Whooooops, typo :)

> The reason for that is,
> is_rar_file is used by the postprocessor to determine which files need to be extracted.
> Without only_first (or when it's set to False), it would try to unpack/unrar every possible rar file:
> .rar, .part1.rar, .part9.rar, .r00, .r09, etc.
> 
> I use only_first only for when unpack method is set to "Treat as video" and I need to process all of the rar files and parts to move them all together, like you asked.

Ok, I get it. It seemed a bit weird to me, that's why I asked. Thanks for the fix !

sickbeard/helpers.py
```diff
@@ -205,7 +207,7 @@ def is_media_file(filename):
     # ignore samples
     try:
         assert isinstance(filename, six.string_types), type(filename)
-        is_rar = is_rar_file(filename)
+        is_rar = is_rar_file(filename, only_first=False)
```
miigotu, line 210: doing this unconditionally wont work, its going to process ear rar piece as a separate video and process them all, overwriting the previous one as it goes (unless renaming is off)

miigotu (changes requested):

sharkykh:
@miigotu 
In that case, associated files is the way to go. That complicates things a bit.

sickbeard/helpers.py
sharkykh: @miigotu 
In that case, associated files is the way to go. That complicates things a bit.

miigotu:
I would suggest adding a check just after the treat_as_video and move/link/copy/etc them separately, idk

PR #2153, opened 2016-07-28
Support reversed symlink => the original file doesn't move

ArthurGarnier:
Proposed changes in this pull request:
- This commit permits users to Symlink files without moving. In the "sickrage directory" there is only symlink.
- The aim ? Personally I manage my files with Transmission, when I delete a file from Transmission I want it deleted on my HDD, a script can easily delete orphan symlink (can be added to SickRage if necessary)

PR #3687, opened 2017-04-28
Ignore broken symlink

andrewleech:
Related to #2153 
When using reversed symlink, the destination folder for post-processing gets a symlink in it pointing to the original file.

If that file is then deleted / moved / cleaned up a broken symlink is left behind, which is not necessarily a problem. 
In my case it's useful to have it there as a cheap reference of what was previously downloaded.

Currently however at startup I get an error entry in the log for ever file as get_size() on it fails with an exception.

- Ignore broken symlinks in get_size() - they're essentially treated as 0 length files with this PR

- [X] Read [contribution guide](https://github.com/SickRage/SickRage/blob/master/.github/CONTRIBUTING.md)


sharkykh (changes requested):
1. When you open a pull request you have the option to choose the target branch, please make sure you manually select develop every time.

2. I don't think this is the right way to fix this.
I think you should use [`os.path.isfile`](https://docs.python.org/2/library/os.path.html#os.path.isfile).

What I have in mind is:
```python
def get_size(start_path='.'):
    """
    Find the total dir and filesize of a path

    :param start_path: Path to recursively count size
    :return: total filesize
    """

    if not ek(os.path.isdir, start_path):
        return -1

    total_size = 0
    for dirpath, dirnames_, filenames in ek(os.walk, start_path):
        for f in filenames:
            fp = ek(os.path.join, dirpath, f)
            if ek(os.path.islink, fp) and not ek(os.path.isfile, fp):
                logger.log("Unable to get size for file {0} because the link to the file is not valid".format(fp), logger.DEBUG)
                continue
            try:
                total_size += ek(os.path.getsize, fp)
            except OSError as error:
                logger.log("Unable to get size for file {0} Error: {1}".format(fp, error), logger.ERROR)
                logger.log(traceback.format_exc(), logger.DEBUG)
    return total_size
```

andrewleech:
Ah sorry, I obviously read the guide wrong, thought it was just the pr branch coming from develop, missed setting the target to develop too.

Yes I like your suggestion. It's certainly better to not silently ignore the issue, log to debug level is much better.

I've tested it, get the new message when debug logging is turned on. Thanks.

OmgImAlexis:
Maybe it's better to add an option for enabling this as an error? I'm sure not everyone wants broken links to now silently be ignored.

sharkykh:
@OmgImAlexis 
You're right, done.
There's no need for an error.
Once you treat the underlying issue that causes the OSError, it can be logged as a warning.

@miigotu 
I'd like your take on it, too.

sharkykh (approved):

PR #3706, opened 2017-05-04
qBittorrent: Verify torrent was added

PancakeZik:
The error has been happening for months.

Fixes [https://github.com/SickRage/old-sickrage-issues/issues/557](https://github.com/SickRage/old-sickrage-issues/issues/557)

Just return True so it doesn't try to set priority, which is not working.

### DIFF
--- a/sickbeard/clients/qbittorrent_client.py
+++ b/sickbeard/clients/qbittorrent_client.py
@@ -18,8 +18,11 @@
 # You should have received a copy of the GNU General Public License
 # along with SickRage. If not, see <http://www.gnu.org/licenses/>.
 
+from time import sleep
+
 import sickbeard
 from sickbeard.clients.generic import GenericClient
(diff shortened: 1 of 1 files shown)


sharkykh:
@PancakeZik 
Could you please try with the fix I pushed?
I don't have qBittorrent, but reading through the [wiki](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation) I realized that we don't know if the torrent was actually added. So I added a check for it.
It will check for 10 seconds (maximum) to see if the torrent was added.

Also, it might not be needed at all for `_add_torrent_file`.
If I'm right it only happens with magnet links.
But as I said, I can't test it myself.

PancakeZik:
@sharkykh 
It works for me

sharkykh (approved):

Issue #3725, opened 2017-05-07
Episode not found in file re-scan


TheRealNoob:
So this is a pretty simple bug.  Episode not found in file scan due to the name of the episode.  I have followed recommended naming convention.
File name:  Marvel's Agents of S.H.I.E.L.D. - S01E02 - 0-8-4 - 720p WEB-DL.mkv

I suspect it is the use of the dashes that causes the issue.  When I remove the episode name (0-8-4) from the file it is identified correctly.  Logs below.

Branch/Commit:  Master / e1af3a8af8dacccfb9c7adf45f08090a68875af2
OS:  Windows-2012ServerR2-6.3.9600
What you did:  Re-Scan Files on TV Show
What happened:  Episode not found
What you expected:  Episode to be found
Logs:
```
2017-05-07 11:01:45 INFO SHOWQUEUE-REFRESH :: 263365: No episode number found in D:\TV Shows\Marvel's Agents of S.H.I.E.L.D\Season 1\Marvel's Agents of S.H.I.E.L.D. - S01E02 - 0-8-4 - 720p WEB-DL.mkv, ignoring it
2017-05-07 11:01:45 INFO SHOWQUEUE-REFRESH :: 263365: parse_result: S01 [ABD: False] [ANIME: False] [whichReg: [u'season_only']]
2017-05-07 11:01:41 INFO SHOWQUEUE-REFRESH :: Performing refresh on Marvel's Agents of S.H.I.E.L.D.
```

h3llrais3r:
It's most likely related to the dashes in the episode name like you mentioned.
Since the separtor in the pattern is a dash, it tries to split it on the dashes with this result.
Can I advise you another naming pattern which should not give any problems?
I use it for all my series and didn't have an issue yet:

![image](https://cloud.githubusercontent.com/assets/1945295/25783729/759bab98-3361-11e7-90bf-1a9c5c2f2bde.png)

h3llrais3r:
I did some debugging and I found out that there is something in the code that causes the 'no match' result.
@miigotu @sharkykh 
Any idea why this is done?
https://github.com/SickRage/SickRage/blob/develop/sickbeard/name_parser/parser.py#L149

Because the 'no_season' regex parses more than 4 extra_ep_num values, the matches are cleared and nothing could be found. I just tested it myself by putting the value to > 10 and now it parses with success.

What's the purpose of this check? Perhaps this can even be removed?

sharkykh:
[...]

PR #3742, opened 2017-05-10
Fix name parser for regex which thinks there are more than 4 episodes

h3llrais3r:
Fixes https://github.com/SickRage/SickRage/issues/3725

### DIFF
--- a/sickbeard/name_parser/parser.py
+++ b/sickbeard/name_parser/parser.py
@@ -147,7 +147,7 @@ class NameParser(object):
                 if 'extra_ep_num' in named_groups and match.group('extra_ep_num'):
                     tmp_episodes = range(ep_num, self._convert_number(match.group('extra_ep_num')) + 1)
                     if len(tmp_episodes) > 4:
-                        break
+                        continue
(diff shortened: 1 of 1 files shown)


sickbeard/name_parser/parser.py
```diff
@@ -147,7 +147,6 @@ def _parse_string(self, name, skip_scene_detection=False):  # pylint: disable=to
                 if 'extra_ep_num' in named_groups and match.group('extra_ep_num'):
                     tmp_episodes = range(ep_num, self._convert_number(match.group('extra_ep_num')) + 1)
                     if len(tmp_episodes) > 4:
-                        matches = []
                         break
```
sharkykh: @h3llrais3r 
I feel like this (`break`) should be `continue` if you want it to keep trying other regexes too.

h3llrais3r: You're right... made a stupid mistake... forgot to change it when I copied my changes in the branch... 😢

sharkykh (approved):
LGTM

PR #3811, opened 2017-05-29
Provider Danishbits updated to use their API. 

MikkelGN:
Danishbits no longer allows systems to send many requests to their torrents.php.
Instead they provide an API at couchpotato.php which is used in this update.


mattail:
Daily-search has been broken since this commit.
Backlog and manual search are working though. 
There's a thread about this on Danishbits forum.
Would be nice if you could fix it.

PR #3840, opened 2017-06-06
switch to yarn, xo and ava

OmgImAlexis:
Fixes #

-
-
-

- [ ] Read [contribution guide](https://github.com/SickRage/SickRage/blob/master/.github/CONTRIBUTING.md)


OmgImAlexis:
@sharkykh @miigotu is the `_` function for get text or underscore? You have both in the bower file.

OmgImAlexis:
I'll be using webpack for bundling which means we can finally use ES6 features.

OmgImAlexis:
I know this is going to be a massive PR but there really isn't any other way todo this.

OmgImAlexis:
@miigotu @sharkykh can one or both of you have a look at the travis file and see if you can work out why yarn isn't installing? It should be using the apt source as [per](https://github.com/travis-ci/travis-ci/issues/6720#issuecomment-285981417).

sharkykh:
@OmgImAlexis 
I'm guessing...
Maybe because the yarn package is installed before NVM installs a different Node.js version,
**but probably** because this is container-based and the yarn source **is not whitelisted**.
The list is [here](https://github.com/travis-ci/apt-source-whitelist), the docs about this - [here](https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon)

I'll try and make it work.
Also, since NPM is pre-installed, we can use it. I don't think it makes any difference to the CI..

sharkykh:
It works now :+1:
Well, `yarn install`, at least

sharkykh:
> is the `_` function for get text or underscore? You have both in the bower file.
@OmgImAlexis 

In core.js, [`underscore` is moved to `__` (double underscore)](https://github.com/SickRage/SickRage/blob/b5b2262ad0acfd0ca679b1964950a147dfc5e193/gui/slick/js/core.js#L86)
so we can use [`_` (single underscore) and `_n` for gettext and ngettext](https://github.com/SickRage/SickRage/blob/b5b2262ad0acfd0ca679b1964950a147dfc5e193/gui/slick/js/core.js#L4050-L4064)

This happens in core.js, and underscore is only being used there.
It's only being used for `__.debounce`, if you think there's another/better solution (not my code),
you may try dropping this dependency.

miigotu:
Failing when grunt ci task is called (which calls jshint and mocha)

OmgImAlexis:
That's fine since I'm replacing them both.

OmgImAlexis:
@miigotu can you explain the changes in https://github.com/SickRage/SickRage/commit/ca8499dba5c6e3ae21c9baa3f604c94d2629a86d in relevance to `stupidOptions`, I don't understand why it was changed this way since `stupidOptions` now just gets set but isn't actually used?

OmgImAlexis:
Can we move the icon from `ajaxNotifications.js` to a file?

sharkykh:
> Can we move the icon from ajaxNotifications.js to a file?

I guess.
We already have [gui/slick/images/ico/favicon-196.png](https://github.com/SickRage/SickRage/blob/master/gui/slick/images/ico/favicon-196.png), which is close enough
(base64 encoded image is 180x180, I think)

miigotu:
Stupid options is used, it gets extended when manual search occurs, and then it changes the icon first to a clock (queued), then spinner while searching, and then back to a magnifying glass
see stupidOptions.loadingClass, etc.

Regardless, dont rely on my rookie js if its ugly, feel free to change it as long as it still works afterwards.

OmgImAlexis:
Having a few issues with index files so just putting this as a note to myself.
Once https://github.com/webpack/webpack/issues/5020 is fixed we can change imports back to using directory imports.

miigotu:
Just a note for me, to adjust the travis-ci.yml later to split the build tests into jobs for python, and javascript, and others...
Link: https://stackoverflow.com/a/44054333/1338570

OmgImAlexis:
@sharkykh would you mind reviewing what I've done so far?

sharkykh:
@OmgImAlexis 
I don't exactly understand this process but I'll try, I guess...

BTW, I seem to get a different yarn.lock when I install
and the UI is broken for me (`formwizard.js:7 Uncaught SyntaxError: Unexpected token import`),
I guess it's probably not done.

```diff
--- a/yarn.lock
+++ b/yarn.lock
@@ -2460,14 +2460,14 @@ lodash@^3.10.1, lodash@~3.10.1:
   version "3.10.1"
   resolved "https://registry.yarnpkg.com/lodash/-/lodash-3.10.1.tgz#5bf45e8e49ba4189e17d482789dfd15bd140b7b6"

-lodash@^4.0.0, lodash@^4.13.1, lodash@^4.14.0, lodash@^4.3.0, lodash@~4.16.4:
-  version "4.16.6"
-  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.16.6.tgz#d22c9ac660288f3843e16ba7d2b5d06cca27d777"
-
-lodash@^4.17.4:
+lodash@^4.0.0, lodash@^4.13.1, lodash@^4.14.0, lodash@^4.17.4, lodash@^4.3.0:
   version "4.17.4"
   resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.17.4.tgz#78203a4d1c328ae1d86dca6460e369b57f4055ae"

+lodash@~4.16.4:
+  version "4.16.6"
+  resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.16.6.tgz#d22c9ac660288f3843e16ba7d2b5d06cca27d777"
+
 lodash@~4.3.0:
   version "4.3.0"
   resolved "https://registry.yarnpkg.com/lodash/-/lodash-4.3.0.tgz#efd9c4a6ec53f3b05412429915c3e4824e4d25a4"

```

OmgImAlexis:
Nope not done, about half way I'd say. I just need to move the rest of the routes into their files and cleanup a few things.

OmgImAlexis:
The import issue is expected since I'm still moving scripts around, we should end up with just a `bundle.js` in a script tag with the rest all bundled together.

gui/slick/js/ajaxEpSearch.js
```diff
@@ -1,112 +1,113 @@
-var searchStatusUrl = srRoot + '/home/getManualSearchStatus';
-var failedDownload = false;
-var qualityDownload = false;
-var selectedEpisode = '';
+import PNotify from 'pnotify';
+
+const searchStatusUrl = srRoot + '/home/getManualSearchStatus';
```
sharkykh, line 3: Of course, correct me if I'm wrong,
**I think** you need to
```js
import config from './config';
const searchStatusUrl = config.srRoot + '/home/getManualSearchStatus';
```
Since srRoot isn't defined here?

gui/slick/js/core.js
```diff
@@ -1,136 +1,117 @@
...
     array.shift();
     return array;
 }
 
-var __ = _; // Moves `underscore` to __
-_ = function(str) { return str; }; // Temporary (gets replaced)
+const __ = _; // Moves `underscore` to __
+_ = function(str) {
+    return str;
+}; // Temporary (gets replaced)
```
sharkykh, line 45: I'd rather ditch this and import `underscore` directly to `__`, if that's possible with ES6

gui/slick/js/core.js
```diff
@@ -1,136 +1,117 @@
...
                         '<span>' + siteMessages[key].message + '</span><span class="glyphicon glyphicon-check site-message-dismiss pull-right" data-id="' + key + '"/>' +
                         '</div>');
                 }
             }
         }
     });
 
-    $('#site-messages').on('click', '.site-message-dismiss', function () {
-        var messageID = $(this).data('id');
+    $('#site-messages').on('click', '.site-message-dismiss', function() {
```
sharkykh: Missed, I think
```js
$('#site-messages').on('click', '.site-message-dismiss', () => {
```

gui/slick/js/core.js
```diff
@@ -2514,70 +2203,73 @@ var SICKRAGE = {
...
                 $('#pickShow option:selected').next('option').prop('selected', 'selected');
-                $("#pickShow").change();
+                $('#pickShow').change();
             });
 
-            $('#changeStatus').on('click', function(){
-                var srRoot = $('#srRoot').val();
-                var epArr = [];
+            $('#changeStatus').on('click', () => {
+                const srRoot = $('#srRoot').val();
```
sharkykh, line 1410: You removed the `srRoot` hidden input element from the template.
https://github.com/SickRage/SickRage/pull/3840/files#diff-3a9e689c8230966780b2688544445f9fL30

So I think that wouldn't work right now.
There might be more of these here (not sure)

It can be `const srRoot = config.srRoot` for a quick fix
or remove this and change the references from `srRoot` to `config.srRoot`

gui/slick/js/core.js
```diff
@@ -2514,70 +2203,73 @@ var SICKRAGE = {
...
-                    switch (found) {
-                        case 2:
-                            return false;
-                        case 1:
-                            this.checked = lastCheck.checked;
+                    if (found === 2) {
+                        return false;
+                    }
+                    if (found === 1) {
+                        this.checked = lastCheck.checked;
```
sharkykh, line 1454: Why did you change the `switch-case` to `if`s?

gui/slick/js/core.js
```diff
@@ -2586,47 +2278,47 @@ var SICKRAGE = {
...
                 });
             });
 
-            // handle the show selection dropbox
-            $('#pickShow').on('change', function () {
-                var srRoot = $('#srRoot').val();
-                var val = $(this).val();
+            // Handle the show selection dropbox
+            $('#pickShow').on('change', function() {
+                const srRoot = $('#srRoot').val();
```
sharkykh, line 1485: Here's another `#srRoot` reference (element was removed from template).

gui/slick/js/core.js
```diff
@@ -2871,44 +2571,48 @@ var SICKRAGE = {
...
+                sortList: [[3, 0], [4, 0], [2, 1]]
             });
         },
-        restart: function(){
-            var currentPid = srPID;
-            var checkIsAlive = setTimeout(function () {
-                setInterval(function() {
-                    $.post(srRoot + '/home/is-alive/', function(data) {
+        restart() {
+            let currentPid = srPID;
```
sharkykh, line 1794: `config.srPID`

gui/slick/js/core.js
```diff
@@ -2920,16 +2624,16 @@ var SICKRAGE = {
...
                             $('#refresh_message').show();
                             srPID = currentPid = data.msg;
-                            checkIsAlive = setInterval(function() {
-                                $.post(srRoot + '/home/is-alive/', function() {
+                            checkIsAlive = setInterval(() => {
+                                $.post(config.srRoot + '/home/is-alive/', () => {
                                     clearInterval(checkIsAlive);
-                                    setTimeout(function(){
+                                    setTimeout(() => {
                                         window.location = srRoot + '/' + srDefaultPage + '/';
```
sharkykh, line 1813:
```js
window.location = config.srRoot + '/' + config.srDefaultPage + '/';
```

gui/slick/js/utils/notifications.js
```diff
@@ -0,0 +1,52 @@
...
+    dir1: 'up',
+    dir2: 'left',
+    firstpos1: 25,
+    firstpos2: 25
+};
+PNotify.prototype.options.styling = 'bootstrap3';
+PNotify.prototype.options.width = '340px';
+PNotify.prototype.options.desktop = {
+    desktop: true,
+    icon: '/images/ico/favicon-196.png'
```
sharkykh: We changed this to
`icon: config.srRoot + '/images/ico/favicon-196.png'`

package.json
```diff
@@ -1,39 +1,112 @@
...
+        "jquery-tokeninput": "^1.6.0",
+        "jquery-ui": "^1.12.1",
+        "lodash": "^4.17.4",
+        "loglevel": "^1.4.1",
+        "pnotify": "^3.2.0",
+        "qtip2": "^3.0.3",
+        "tablesorter": "2.28.4",
+        "timeago": "^1.5.4",
+        "underscore": "^1.8.3",
+        "xo": "^0.18.2"
```
sharkykh, line 40: This is the linter, right? Is that supposed to be in `devDependencies`?

sharkykh (commented):
Tried my best. but `core.js` is just too much to check IMO.
I mean, I did go over it, but couldn't stay focused / in-focus.

gui/slick/js/utils/notifications.js
```diff
@@ -0,0 +1,52 @@
...
+            .replace(/<[/]?b(?:\s[^>]*)?>/ig, '*')
+            .replace(/<i(?:\s[^>]*)?>/ig, '[').replace(/<[/]i>/ig, ']')
+            .replace(/<(?:[/]?ul|\/li)(?:\s[^>]*)?>/ig, '').replace(/<li(?:\s[^>]*)?>/ig, '\n* ')
+    };
+    new PNotify(opts); // eslint-disable-line
+};
+
+const checkNotifications = timeout => {
+    $.getJSON(config.srRoot + '/ui/get_messages', data => {
+        $.each(data, (name, {type, title, message}) => displayPNotify({type, title, message, name}));
```
sharkykh, line 44: #3868 changes this.
If I understand this correctly:
```js
$.each(data, (name, {type, title, message, hash}) => displayPNotify({type, title, message, hash}));
```

gui/slick/js/core.js
OmgImAlexis: Since this uses `$(this)` I can't use an arrow function so I haven't moved it over yet, I'm switching most of these to `(index, element) => {}` to completely remove the need for `$(this)`.

gui/slick/js/core.js
OmgImAlexis: underscore will be imported by named module instead so we can use lodash and underscore together with gettext without any of them needing to override each other.

_ will be gettext in all files imported as `import {gettext as _} from './utils';` where as lodash/underscore we'll import as `import {set} from lodash;`.

gui/slick/js/core.js
OmgImAlexis: Switches aren't a good practise in js and I'm most likely going to just a toggle util that returns a flipped boolean instead so it'd be like `x = flipBoolean(this.checked);` and we'd get the opposite returned. If statements and functions also allow for better testing where as switches can have issues with default branches being used by accident, etc.

package.json
OmgImAlexis: Yes I'll move that over, xo tends to install itself in deps by default.

gui/slick/js/utils/notifications.js
OmgImAlexis: Yep I'll switch that over. The first half `{type, title, message, hash}` are the deconstructed fields coming out of data, we then just pass them to displayPNotify since ES6 allows shorthand.

Ref: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Object_initializer#Property_definitions

OmgImAlexis:
I'll fixup the srRoot and other comments, thanks.

miigotu:
Boo conflicts =P

OmgImAlexis:
@sharkykh once this is done you should be able to use something like this. This will create the element, add it to the DOM, add a click handler for the dismiss button and if the tag is used before it'll replace the content with the new message from Sickrage. I think we should change tag to be called id though as that's really what it is. Then in js we can just use `id = new UUID();` when submitting them.

```
const message = new SiteMessage(level, tag, message);
```

sharkykh:
> [...] and if the tag is used before it'll replace the content with the new message from Sickrage. I think we should change tag to be called id though as that's really what it is. Then in js we can just use `id = new UUID();` when submitting them.
@OmgImAlexis

I took a look at the `utils/site-message.js` code, hopefully I understand it correctly... 😅 
First, **`tag` is not an id for the message**, it's more of like a... category.
[We store the messages in Python as a dictionary and use incremental ids as its keys.](https://github.com/SickRage/SickRage/blob/fa5b11c7092984b3e31f9b0cfe5f513481041ac4/sickbeard/helpers.py#L1847-L1873)

For the off chance you didn't realize that yourself, I'll say:
When we have, for example, a message tagged `git_newer_version_available`
and we check for version again, and there's an even newer version,
we replace the similarly **tagged** message if such message exists, **in Python**,
by rewriting the existing message with the new message (You can see that in the Python code linked above)

I don't see a reason to change this and do this in JS because...
1. Even when we post a request to add a new site message to the server,
**and we don't do this anywhere in JS currently**,
the server/Python should handle it, as this is where it's stored.
2. Since messages are dismissable by their id (or key in said dictionary),
I don't think we really need UUIDs. They're not **_actual_** notifications, they're just a global message (or global messages) that are stored until dismissed or until the web server restarts.

miigotu:
I think the other way would be useful for things like "Settings saved" etc, but that is a totally different purpose and should probably be set separately, or in a js_messages tag

miigotu:
I'm just re-opening this and saving the branch so I can reference it while I try to figure out how webpack works. Ive been trying over the past week or so and it's fairly complicated.

I'm mostly interested in removing bower and grunt entirely. I have assumed webpack would be the best compliment to yarn, with the grunt tasks ported into yarn tasks.
Let me know if any of you guys are interested in breaking this into smaller tasks and tackling them, like porting over the grunt publish tasks to yarn only or helping me with webpack.

@SickChill/collaborators @SickChill/moderators @OmgImAlexis @sharkykh

ghost:
Remove all of the outdated babel deps and webpack v3.
Install the latest webpack, etc.

`npm i -D webpack babel-loader @babel/core @babel/preset-env @babel/polyfill`

<details><summary>webpack.config.js</summary>

```js
const path = require('path');

module.exports = {
    devtool: 'inline-source-map',
    module: {
        rules: [{
            // Only run `.js` and `.jsx` files through Babel
            test: /\.jsx?$/,
            // Skip any files outside of your project's `src` directory
            include: [
                path.resolve(__dirname, './gui/slick/js/**/*')
            ],
            exclude: /node_modules/,
            use: {
                loader: 'babel-loader',
                options: {
                    presets: [
                        ['@babel/preset-env', {
                            useBuiltIns: true // Include required parts of the polyfill.
                            // modules: false // Is this still necessary for tree shaking?
                        }],
                    ]
                }
            }

        }]
    },
    entry: [
        // Add your application's scripts below
        './gui/slick/js/index'
    ],
    output: {
        filename: 'bundle.js',
        path: path.resolve(__dirname, './gui/slick/js/dist/')
    }
};
```
</details>

miigotu:
@pro-src yeah I actually had a much more involved webpack.config.js (and parts split into webpack.parts.js) but I had some pretty nasty undefineds, expecially gettext, lodash, and $/jQuery.
I think the whole thing needs reworked to move the js out of the gui/slick/js and into a src folder and then add index/entry to the src folder and then output to gui/slick/js/core.webpack.js, etc.  The hardest part is that its several removals at the same time because we have to remove bower, rewrite the grunt tasks into package.json, and also add webpack. It's a pita lol

I'm not sure if I made changes to this trying to get it to work but this is the closest I came:
<details><summary>webpack.config.js</summary>

```js
const webpack = require('webpack');
const path = require('path');
const merge = require('webpack-merge');
const parts = require('./webpack.parts');

const config = merge([
    {
        entry: './gui/slick/js/core.js',
        output: {
            path: path.resolve(__dirname, 'gui/slick/js'),
            filename: 'core.webpack.js'
        },
        mode: 'development',
        module: {
            rules: [
                {
                    test: /\.js$/,
                    use: 'babel-loader',
                    exclude: /node_modules/
                },
                {
                    test: /\.css$/,
                    use: [
                        'style-loader',
                        'css-loader'
                    ]
                },
                {
                    test: /\.less$/,
                    use: [
                        'style-loader',
                        'css-loader',
                        'less-loader'
                    ]
                },
                {
                    test: /\.scss$/,
                    use: [
                        'style-loader',
                        'css-loader',
                        'sass-loader'
                    ]
                },
                {
                    test: /\.png$/,
                    use: [
                        {
                            loader: 'url-loader',
                            options: {
                                mimetype: 'image/png'
                            }
                        }
                    ]
                },
                {
                    test: /\.svg$/,
                    use: 'file-loader'
                }
            ]
        },
        plugins: [
            new webpack.ProvidePlugin({
                'window.jQuery': 'jquery',
                'window.$': 'jquery',
                'jQuery': 'jquery',
                '$': 'jquery',
            })
        ],
        externals: {
            lodash : {
                commonjs: 'lodash',
                amd: 'lodash',
                root: '_' // indicates global variable
            },
            srRoot: {
                root: 'srRoot'
            }
        }
    },
    parts.minifyJavaScript(),
    parts.minifyCSS({
        options: {
            discardComments: {
                removeAll: true,
            },
            // Run cssnano in safe mode to avoid
            // potentially unsafe transformations.
            safe: true,
        },
    })
]);

module.exports = config;

```

</details>

<details><summary>webpack.parts.js</summary>

```js

const OptimizeCSSAssetsPlugin = require(
  "optimize-css-assets-webpack-plugin"
);
const cssnano = require("cssnano");

exports.minifyCSS = ({ options }) => ({
  plugins: [
    new OptimizeCSSAssetsPlugin({
      cssProcessor: cssnano,
      cssProcessorOptions: options,
      canPrint: false,
    }),
  ],
});

const TerserPlugin = require("terser-webpack-plugin");

exports.minifyJavaScript = () => ({
  optimization: {
    minimizer: [new TerserPlugin({ sourceMap: true })],
  },
});

```

</details>

ghost:
Great, I'll try to get around to assisting more.

<details><summary>Notes on Babel Polyfill</summary>

The Babel Polyfill has been deprecated. (Quoting https://babeljs.io/docs/en/babel-polyfill)

> As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions):
```js
import "core-js/stable";
import "regenerator-runtime/runtime";
```

There is an advantage of using the `@babel/preset-env` with the `useBuiltins: "usage"` option. It should only include parts of the polyfill that are actually needed.

You can see the affect of importing the polyfill directly by viewing it's source. It bulls down to https://github.com/babel/babel/blob/master/packages/babel-polyfill/src/noConflict.js

It's best to go ahead and use corejs and the regenerator directly as recommended.
This explains things very well: https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md
Instructions (same page): https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#babelpolyfill

If you haven't already, you can dump the babel related config from package.json as well.
</details>

Do you have another branch with that config on it?

miigotu:
I do not have a branch with it pushed up yet, but I will push my WIP branch up soon. I have split webserve.py into separate files and am working on reorganizing the classes and urls, as well as using named url patterns so we can reverse_url('home') for example.

In the process of switching from chrome to firefox atm though.

PR #3854, opened 2017-06-12
Update MoreThanTV Provider - Allow for proper season searches

### LINKED ISSUE #2496: MoreThan.tv not catching season packs

On morethan.tv most of the season packs are under the correct show and named by what season they are. The season packs will only work if it can use which show the pack belongs to rather than the name itself.  

Example of pack caught now:
ShowName - ShowName.S01.720p.BluRay.X264-Scene [2009] [Scene] // TV

example of a non-caught release:
Example of pack caught now:
ShowName - Season 1.720p.BluRay.X264-Scene [2009] [Scene] // TV

Branch/Commit:
21c9621a36dabe345fc4879200c3fba39b84c8ce 
OS:
Linux
What you did:
Search show for season pack only

What happened:
Didn't capture season pack because of name

What you expected:
Should of caught the shows name

Logs:
2016-10-17 13:09:55 DEBUG    SEARCHQUEUE-BACKLOG-305288 :: [MoreThanTV] :: Fallback episode search initiated
2016-10-17 13:09:55 DEBUG    SEARCHQUEUE-BACKLOG-305288 :: [MoreThanTV] :: Unable to match Season.1 to a show in your database. Parser result: S01 [ABD: False] [ANIME: False] [whichReg: ['season_only']]
2016-10-17 13:09:55 DEBUG    SEARCHQUEUE-BACKLOG-305288 :: [MoreThanTV] :: Unable to match Season.1 to a show in your database. Parser result: S01 [ABD: False] [ANIME: False] [whichReg: ['season_only']]
2016-10-17 13:09:55 DEBUG    SEARCHQUEUE-BACKLOG-305288 :: [MoreThanTV] :: Found result: Season 1 with 80 seeders and 2 leechers
2016-10-17 13:09:55 INFO     FINDSUBTITLES :: No subtitles to download
2016-10-17 13:09:55 DEBUG    SEARCHQUEUE-BACKLOG-305288 :: [MoreThanTV] :: Found result: Season 1 with 62 seeders and 1 leechers

Narkyzz:
Fixes #2496

Currently the MoreThanTV provider is only able to process ungrouped torrents, however most of the season packs are grouped as "Season XX" torrent groups so it is unable to download them.

- Ability to do backlog season searches for both grouped and normal torrents.
- Relies on torrent's folder name when it's grouped to obtain a release name, not just "Season XX"
- If it's not grouped, process like a normal torrent/episode.
- Obtains the search results from both "SXX" and "Season XX" search strings and merges the torrent table as one to process both at the same time. In case one or both search strings returns no torrent, it handles accordingly to avoid errors.


sharkykh:
@Narkyzz 
If you need to change Season search strings for the specific provider,
please override the [`_get_season_search_strings`](https://github.com/SickRage/SickRage/blob/12e314818381add3323e77ba26867bdd2b1ef347/sickrage/providers/GenericProvider.py#L430-L449) function instead.
(If you need to do this for episodes, override [`_get_season_episode_strings`](https://github.com/SickRage/SickRage/blob/12e314818381add3323e77ba26867bdd2b1ef347/sickrage/providers/GenericProvider.py#L392-L428))

**_Perhaps_** like in [BTN](https://github.com/SickRage/SickRage/blob/96b53dee713dc35ebf295d0f37a67b2a36fb4ea3/sickbeard/providers/btn.py#L201-L261).

Narkyzz:
Okay that's fixed for season search, the episode search strings stay the same.

sickbeard/providers/morethantv.py
```diff
@@ -127,9 +128,12 @@ def process_column_header(td):
                     logger.log("Search string: {0}".format
                                (search_string.decode("utf-8")), logger.DEBUG)
 
-                search_params['searchstr'] = search_string
+                if mode == 'Season':
+                    searchedSeason = re.match('.*\s(Season\s\d+|S\d\d+)', search_string).group(1)
```
sharkykh: If I'm reading this right (sorry, late at night),
The right side of the RegEx "or" can be `S\d+`, since `+` is 1 or more of the previous character.

```
searchedSeason = re.match('.*\s(Season\s\d+|S\d+)', search_string).group(1)
```

sickbeard/providers/morethantv.py
```diff
@@ -153,6 +157,31 @@ def process_column_header(td):
...
+                                if 'Season' in title:
+                                    torrentid = urljoin(self.url, result.find('span', title='Download').parent['href'])
+                                    torrentid = re.match('.*?id=([0-9]+)', torrentid).group(1)
+
+                                    group_params = {
+                                        'torrentid': ''
+                                    }
+
+                                    # Obtain folder name to use as title
+                                    group_params['torrentid'] = torrentid
```
sharkykh: This is not needed.
You can just do this with the previous statement:
```python
# Obtain folder name to use as title
group_params = {
    'torrentid': torrentid
}
```

sharkykh (commented):
Forgot to say, thanks for the PR.
Two small notes, other than that looks good, I guess.

Make sure you test this, as we can't since it's a private tracker.

Narkyzz:
It could be `S\d+` but it's less likely to make a mistake (I think) with `S\d\d+` since there's always a leading zero for the seasons.
Also I have tested it properly and it works like it should.

sharkykh:
@Narkyzz 
I totally missed that the check is on the `search_string`.

But still, it's really the same. Just try this out:
https://regex101.com/r/eM4qrX/2

^ This actually says it takes the regex parser one less step in comparison..

Narkyzz:
I was just trying to minimize the risk in case a show has "S\d" in the name without it being the actual season number, but it probably wouldn't happen.

sharkykh:
> I was just trying to minimize the risk in case a show has "S\d" in the name without it being the actual season number, but it probably wouldn't happen.
@Narkyzz 

I just checked in Python, it only gives you the last match.
And you know that the last thing in the string is the season number you put in `_get_season_search_strings`

So yeah, won't happen, this is all good.

sharkykh (approved):

PR #3858, opened 2017-06-13
Restore `bencode` library to vanilla state

sharkykh:
* Restore lib `bencode` to **vanilla** state + update libs list
* Handle 'data after valid prefix' exception where bencode is used

Tested by raising a similar exception outside of SickRage.


sharkykh:
Shoot. bencode won't return the value if it's the exception's raised.

PR #3859, opened 2017-06-13
ajaxNotifications.js: Replace base64 image with actual image

sharkykh:
As suggested https://github.com/SickRage/SickRage/pull/3840#issuecomment-306988965
Tested and looks good.


OmgImAlexis:
Not 100% sure this will work if someone uses a subdirectory or srRoot. @sharkykh could you test, if it's all good we can merge. I'll make the same change on the PR I've got open.

sharkykh:
Good catch! I forgot about srRoot. Just tested and it doesn't work without it (had to disable browser cache to verify). Updated my code.

OmgImAlexis (approved):

PR #3861, opened 2017-06-13
Update subscenter domain to .info

sagilo:


sharkykh:
If .org works why change it?

sagilo:
the API doesn't work with .info.
See :
http://www.subscenter.org/he/cst/data/series/sb/greys-anatomy/13/22
And :
http://www.subscenter.info/he/cst/data/series/sb/greys-anatomy/13/22/

sharkykh:
Thanks for clarifying

sharkykh (approved):

PR #3873, opened 2017-06-18
Use `static_url` for linking static files

sharkykh:
Generate relative links to static files in templates. This takes `sickbeard.WEB_ROOT` into account and also generates an "E-Tag" of the static file, so it gets cached when unchanged.


OmgImAlexis (approved):
This will be even better once we merge all the webpack as we'll only have a single file that needs to be cached then.

miigotu:
The cache breaker is not for tornado, its for the client to know if the file has changed.  Chrome for example has no way to know if a file has changed on the server, so without the cachebreaker it makes a request and gets an etag saying it hasnt changed, so the client loads from cache.

The cache breaker is intended to remove the need for the request altogether, and we can set high expirations for them on the first request.
Once the static file handler is working properly, tornado would add the v= cachebreaker for us, you import the static method into the template and call it like a template tag

sharkykh:
> Once the static file handler is working properly, tornado would add the v= cachebreaker for us, you import the static method into the template and call it like a template tag
> @miigotu 

I learned that from messing with Flask, but I couldn't get the [`static_url` from `tornado.web`](https://github.com/tornadoweb/tornado/blob/9291944071ec17f9496667920ba89bec72dfba06/tornado/web.py#L1371-L1391) to be available in the Mako template.
Is that what you're referring to?

miigotu:
Yes, I had it in the template before, its just a bit tricky with the wonky inheritance we have.  Not sure if rh is still available in the templates but it was rh.static_url

gui/slick/views/config_subtitles.mako
```diff
@@ -14,10 +14,10 @@
                 preventDuplicates: true,
                 prePopulate: [${','.join("{\"id\": \"" + code + "\", name: \"" + subtitles.name_from_code(code) + "\"}" for code in subtitles.wanted_languages())}],
                 resultsFormatter: function(item) {
-                    return "<li><img src='${srRoot}/images/subtitles/flags/" + item.id + ".png' onError='this.onerror=null;this.src=\"${srRoot}/images/flags/unknown.png\";' style='vertical-align: middle !important;' /> " + item.name + "</li>"
+                    return "<li><img src='${ static_url('images/subtitles/flags/" + item.id + ".png') }' onError='this.onerror=null;this.src=\"${ static_url('images/flags/unknown.png') }\";' style='vertical-align: middle !important;' /> " + item.name + "</li>"
```
sharkykh: This obviously wouldn't (doesn't) work. Anything JavaScript/CSS related, really.

sharkykh:
Anything loaded by JavaScript / CSS in templates or in `.js`/`.css` files obviously can't use `static_url`.
But it seems to work... I think.
I still see megabytes of transfers on the show list, but that's because of the showPoster route.

OmgImAlexis:
@sharkykh does the static url function take into account reverse proxies?

sharkykh:
@OmgImAlexis It generates a relative path (no schema/domain/port). Shouldn't be an issue.

OmgImAlexis:
I'd highly suggest you test that before merging just incase. I'm sure there's a reason we hadn't switched this over before now.

sharkykh:
I would, but I don't have a reverse proxy.

The `src`s and `href`s are very similar. Only change is the `?v=` param is set to some kind of hash that's generated based on the static file, rather than set to the process ID.

> I'm sure there's a reason we hadn't switched this over before now.

I tried to search for this `static_url` function using git history, but found no mention of it.
Maybe I missed something, but looks like it never made it into master? (nothing on github either)
@miigotu

gui/slick/views/config_subtitles.mako
miigotu: build the url before the return

sharkykh: Mako doesn't have the value of the **JavaScript variable** `item.id` when it compiles the template.
I'll probably revert that and similar lines to use `srRoot`, for now.

OmgImAlexis: All js can be left as is since it's already fixed in the webpack stuff.

sharkykh: @OmgImAlexis You didn't change this file in your PR...
I don't think there's a way to do this in JavaScript **without calling Tornado to generate a `static_url` first** (like, a route in Python that will return a static url from a relative path string)
```
request	  > GET /ui/static_url?file=images/subtitles/flags/eng.png
response  < '/web_root/images/subtitles/flags/eng.png?v=174981723981230'
```

miigotu:
"""
I tried to search for this static_url function using git history, but found no mention of it.
Maybe I missed something, but looks like it never made it into master? (nothing on github either)
@miigotu
"""
I was not able to get it working before =P... Is this all working now?

sharkykh:
Looks like it. I've had this on my installation since my last commit and haven't hit any issues.
It doesn't seem to break anything, so that's good.
About the reverse proxy, as I said, the generated path or url is relative and doesn't include the hostname/port, and also the web root is in it,
It's basically similar to what we currently have.
So I don't believe there will be an issue.

Safe to merge, in my opinion.

**Edit:** On the other hand, not so sure it fully works...
The Network tab in Chrome's (v59) DevTools logs that `style.css` (for example) is fetched in full (gzipped ~13kb) almost every refresh or when loading a new page...
However `vender.min.js` which is roughly 1mb or so, is cached properly....
It looks like only some css files are not cached?
image

miigotu:
Might be another fix for that later, this is a move in the correct direction

PR #3877, opened 2017-06-21
Update Python dependencies - 2nd batch

sharkykh:
List of changes in this pull request:
- [x] 	Update `Unidecode` to 0.04.20 (was 0.04.12)
- [x] Vanilla `babelfish` @ 0.5.5 (was 0.5.5-dev -- no change in package code)
- [x] Update `markdown2` to 2.3.4 (was 596d48b ~ 2.3.1)
  News / Changelog pages tested in SickRage
- [x] Update `xmltodict` to 0.11.0 (was 0.9.2)
- [x] 	Update `python-dateutil` to 2.6.0 (was d05b837 ~ 2.5.0)
- [x] 	Replace unmaintained SocksiPy, with maintained PySocks
adds a dependency for `win-inet-pton` that is required on Windows systems.
**I'm still not sure if we actually use this anywhere.
In import lines, it's mostly a fallback if other modules are not available.**
I would (ultimately) like to remove this completely, if possible.
- [x] 	Remove `simplejson` (unused)
I think this is a remnant from way back when SR supported Python 2.6?
In any case, json is available in Python >= 2.7, and no module explicitly imports `simplejson`.
That is mostly a fallback on `ImportError`, maybe the other way around (doesn't matter, really)
- [x] Update `IMDbPY` to 5.1.1
Install arguments:
`pip install imdbpy --no-deps --global-option="--without-sqlobject" --global-option="--without-sqlalchemy"`
- [x] Update `PyGithub` to v1.34 vanilla
  Dependencies:
    - `pyjwt` @ 1.5.0 (new)
- [x] Update `tzlocal` to 1.4
- [x] Restore `FeedParser` to vanilla (now using VCS install)

Other:
- [x] **List the versions** of `decorator` + `validators`
**Upgrading `validators` breaks url validation.**
`validators.url()` was changed in newer version, and AFAICT, it fails on urls with spaces in them.
Maybe `decorator` can be upgraded, I still haven't checked.


sharkykh:
Added `PyGithub`, `IMDbPY`, `tzlocal`, and `FeedParser`
I'll leave `PySocks` in for now, and leave `decorator` + `validators` in their current versions for now.

sharkykh:
`IMDbPY` :angry:
> **Still can't install this from requirements.txt**, because we need to be able to use `--no-deps`, but it's unsupported in the context of requirement files.
A workaround seems to be to use `pip install --no-deps -r requirements.txt`, which is fine as long as we list ALL dependencies (which is planned).
Also, I don't think `install_requires` in `setup.py` supports anything more than simple package version constraints, maybe VCS sources (git/mercurial/etc..) - untested.

BTW, the `hachoir` libs have issues with installing from VCS.
Maintainer [had ditched the Python2 project some time ago](https://bitbucket.org/haypo/hachoir),
and is now actively working on a [Python3-**only** port](https://github.com/haypo/hachoir3).

PR #3901, opened 2017-07-03
Docker support

kylegordon:
- This enhancement correctly supports docker, in that it creates a static binary image containing the desired version of Sickrage. The linuxserver.io implementation performs a git pull at startup, meaning that the version of sickrage cannot be guaranteed, but they are happy with that. https://github.com/linuxserver/docker-sickbeard/issues/8#issuecomment-307518311
- Creation of docker a docker image can be done with the following command
```export SICKRAGE_VERSION=v2017.06.05-1; docker build . --build-arg SICKRAGE_VERSION=${SICKRAGE_VERSION} -t sickrage:${SICKRAGE_VERSION}```
- The resultant image can then be pushed to Docker Hub with a command similar to ```docker push sickrage/sickrage:v2017.05.10-1```
- To run a particular image, ```docker run --rm --name sickrage -it -v /docker/sickrage/data:/data -v /docker/sickrage/config:/config -p 8081:8081 sickrage:v2017.06.05-1```
- If you use 'master' in as the version, sickrage will continue as normal to update to master via git. Unfortunately I'm not sure how best to disable update checks in config.json at runtime for non-master versions. Whilst it can be preconfigured in /config/config.json at build time, that directory is ignored when the user starts with an empty persistent volume.
- A docker-compose.yaml file is also included, such that Sickrage docker can be configured in a static file and deployed with a command such as ```docker-compose -f docker-compose.yaml up -d```

### CODE REVIEW

File: docker-compose.yaml [hunk 1]
```diff
@@ -0,0 +1,15 @@
+version: '2'
+
+services:
+  sickrage:
+    image: sickrage:v2017.06.05-1
+    container_name: sickrage
+    restart: unless-stopped
+    environment:
+      - PUID=200
```
OmgImAlexis: `PUID` and `PGID` should be 100 by default.

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM python:2.7.13-alpine
+# Run build with...
+# export SICKRAGE_VERSION=v2017.06.05-1; docker build . --build-arg SICKRAGE_VERSION=${SICKRAGE_VERSION} -t sickrage:${SICKRAGE_VERSION}
+# Run docker image with
+# docker run --rm --name sickrage -it -v /docker/sickrage/data:/data -v /docker/sickrage/config:/config -p 8081:8081 sickrage:v2017.06.05-1
```
OmgImAlexis: Can you please remove this whole comment. @sharkykh do you want to add this as a npm script or via grunt?

OmgImAlexis (changes requested):

docker-compose.yaml
kylegordon: That's actually an unused hangover from the linuxserver.io compose file that I plagiarised. I've now removed it. 

Is it a strict requirement to run as 100? Leaving it out will cause it to run as root, but controlled by the standard docker/cgroups/namespace restrictions. I'll have a look at editing the startup procedure to ensure it runs as that particular user, if need be.

OmgImAlexis: Whoops, meant 1000 as that should set them as *nix OSs mainly use 1000+ for users so 1000 would be user 1 essentially.

kylegordon: Ah, I understand :-)

I've reworked it so that when it's build it creates an internal user of UID 1000. Short of using an overlay such as maybe s6, the user is set during build time, so I've removed the env vars from docker-compose.yaml.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ENV PGID 1000
+
+RUN apk update ; apk add git
+
+RUN addgroup -g ${PGID} sickrage ; adduser -u ${PUID} -D -S -G sickrage sickrage
+
+RUN git config --global advice.detachedHead false ;
+RUN git clone --quiet https://github.com/SickRage/SickRage/ --branch $SICKRAGE_VERSION --single-branch --depth=1 /app/sickrage
+
+RUN mkdir /var/run/sickrage/ ; chown sickrage. /var/run/sickrage/
```
OmgImAlexis: What's the reason for using `;` over `&&`? If it's style choice then can we switch to `&&` since it's what most people are familiar with and will lead to less confusion when editing the Dockerfile.

Other than that the PR looks good.

docker-compose.yaml
```diff
@@ -0,0 +1,12 @@
+version: '2'
+
+services:
+  sickrage:
+    image: sickrage:v2017.06.05-1
```
OmgImAlexis, line 5: @sharkykh we should add either a script to `package.json` or a grunt task for bumping this when we release a new git tag.

We also need to look into building and pushing a new container to docker hub when we release a git tag.

sharkykh: Can't `master` be used?

Dockerfile
sharkykh: @OmgImAlexis If this is like a run script, wouldn't a simple script in [runscripts](https://github.com/SickRage/SickRage/tree/develop/runscripts) do the job?

docker-compose.yaml
OmgImAlexis: No that's the whole point of this PR. To change it so we're building containers that are tagged with certain versions as that's how Docker is meant to be used instead of how we have it now where the container just builds the latest.

sharkykh: Sorry, I've never really used Docker,
So we have to build a docker image with each new version and publish it somewhere?

OmgImAlexis: Yep they get pushed to [docker hub](https://docs.docker.com/docker-cloud/builds/push-images/). [For example](https://sebest.github.io/post/using-travis-ci-to-build-docker-images/).

kylegordon: Certainly, master can be used. I just selected v2017.06.05-1 as that was the latest release tag. Just drop in whatever tag is desired into the SICKRAGE_VERSION env before build, and it'll build a docker image of that version. 

The image is then present locally, and needs pushed to a repository, such as Dockers own docker hub. If it's Travis doing the heavy lifting, then something similar to https://sebest.github.io/post/using-travis-ci-to-build-docker-images/ could be used.

Once it's been uploaded to Docker hub, a user can run it in a couple of ways. Either, using bare docker commands, such as 'docker run --rm -v /docker/sickrage/data:/data -v /docker/sickrage/config:/config -p 8081:8081 sickrage:v2017.06.05-1'

The --rm flag is to tidy up after it closes down, along with -v to mount the host directories in the docker image, so that the /data and /config persist on the users drive even when the docker image is stopped or deleted. -p is mapping container ports to the host. 

The other way to bring up a docker image is with docker-compose.yaml, which has a lot of the parameters predefined. Using the example one I've provided, it would be a case of ```docker-compose up -d```, and that will pull the image down and run it with the provided config.

I hope I've managed to make sense whilst describing it :-)

Dockerfile
kylegordon: I debated with myself over ; vs && \ and just made the wrong choice :-) 

In terms of bash behaviour, the exit codes should get passed up to docker build, so ; vs && on it's own shouldn't be an issue. 

Are newlines for each command ok style wise?

kylegordon: Have moved it over to && \ multiline style :-)

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+VOLUME ["/config","/data"]
+
+USER sickrage
+
+WORKDIR /app/sickrage/
+
+CMD /usr/local/bin/python SickBeard.py -q --nolaunch --pidfile=${PID_FILE} --config=${CONF_DIR}/config.ini --datadir=${DATA_DIR} ${EXTRA_DAEMON_OPTS}
+
+EXPOSE 8081
```
OmgImAlexis, line 39: @sharkykh should we be exposing 443 incase the user would like to use the built in SSL or should be just assume they'll be using this with a reverse proxy?

OmgImAlexis:
If @miigotu and/or @sharkykh wants to review this I think we should be good to merge.

OmgImAlexis (approved):

Dockerfile
miigotu: these files are for a template. if the user doent know how to use/edit them for their needs they shouldnt be using them

OmgImAlexis: I get that but if we push this to docker hub then people using them via `docker pull` won't have access to the `Dockerfile` so by adding 80 and 443 after 8081 it'll allow them to use it with things like Dokku/Flynn/UNRAID, etc. without any extra effort.

OmgImAlexis: `EXPOSE 8081 80 443` allows any of those ports to be used by Sickrage inside the container.

miigotu: Yeah, I understand what it does (I do a lot of docker with work, run our own registry and all.... no one else knows wtf it does there haha).
For now this is fine, ill set up an offical docker hub account for SR and then we can worry about it later.

sharkykh:
I never used Docker so I couldn't really give a real review @OmgImAlexis

CHBMB:
Just an update, after the latest DMCA debacle our version (linuxserver.io) now doesn't git pull on start but rather does so at build time.  The image being updated each week.

PR #3912, opened 2017-07-07
Update requirements list

sharkykh:
**TL;DR** This doesn't affect any actual code. Safe to merge.

- Update the list to account for the changes made in #3877 
- Specify VCS sources for packages that are on a specific commit, and
- Enable VCS sources for some packages (mostly the unmodified ones)
- Pin `cfscrape` to <1.8.0 so it doesn't say "outdated" on requires.io
- Change `html5lib` version to 1.0b10 (new version style)
- Change `pgi` version to the actual commit hash

VCS sources won't work with setuptools (`setup.py`) in the current configuration,
but [there might be a possible solution](https://stackoverflow.com/a/18563923/7597273).

### DIFF
--- a/lib/html5lib/__init__.py
+++ b/lib/html5lib/__init__.py
@@ -22,4 +22,4 @@ __all__ = ["HTMLParser", "parse", "parseFragment", "getTreeBuilder",
            "getTreeWalker", "serialize"]
 
 # this has to be at the top level, see how setup.py parses this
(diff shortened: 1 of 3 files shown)


sharkykh:
> VCS sources won't work with setuptools (setup.py) in the current configuration.

They actually do seem to work.. The SO answer I looked at was probably outdated.

PR #3916, opened 2017-07-10
Swap jshint with xo (+2 more small fixes)

OmgImAlexis:
@sharkykh since I'm still working on my other PR I thought I might as well do the linting changes separately so they can be used before the webpack stuff is finished.


OmgImAlexis:
@sharkykh can you review and merge this when you get a chance. Not sure why codecov is down since I didn't touch any python files. 😕

OmgImAlexis:
Also can you look into Grunt with Travis and Appveyor I don't know how you intended on the grunt ci task to be used so I've just removed the stuff that's no longer needed.

sharkykh:
> Also can you look into Grunt with Travis and Appveyor I don't know how you intended on the grunt ci task to be used so I've just removed the stuff that's no longer needed.

Yeah the old previous `travis` task had `jshint` and `mocha`, so when we added AppVeyor I changed the task name to `ci`. Then I added `ci:update` task just for Travis to update translations and push them (because miigotu didn't like having more bash scripts).

So basically what you did is fine, just `grunt ci` is no longer needed, because it can be replaced with `npm run test`, which is better.

I'll take at look at the other changes now.

sharkykh:
> @sharkykh can you review and merge this when you get a chance. Not sure why codecov is down since I didn't touch any python files. 😕 

Here, now it's fixed, lol. Codecov gets confused sometimes. It's was just the build failing, I think.
(Because the CIs still sent the coverage reports even though the JS tests failed)

sharkykh:
Umm, @OmgImAlexis , you're adding ES6 syntax here. How will that work if the browser doesn't support it?
(In case there's an end user with an old browser)

OmgImAlexis:
What parts are using ES6? `const` and `let` are usable in all browsers and the ones that have issues with it just treat them both as vars.

I made sure to replace the arrow functions with normal functions as well as not using shorthand object fields as that only works with ES6(Chrome,FF,Edge).

sharkykh:
Okay, good to know on `const` and `let`.
What about http://es6-features.org/#MethodProperties.
For example, in [parsers.js](https://github.com/SickRage/SickRage/pull/3916/files#diff-eef321619506a694c62fa49ad7f9eba9L3)

~or am I just comparing it wrong again?~

> as well as not using shorthand object fields as that only works with ES6(Chrome,FF,Edge).

Then they were just missed...?


Damn, I comment without fully reading your comment.

OmgImAlexis:
Let me know if you find anything else. I'll switch them over once you finish the review.

OmgImAlexis:
Since I've added the rule to xo it'll force longform until we can switch to webpack then we can just remove the rule from `package.json` and run `xo --fix` to switch back to shorthand.

sharkykh:
Good, this is much easier to review now, thanks :)

gui/slick/js/configProviders.js
```diff
@@ -222,34 +218,34 @@ $(document).ready(function(){
...
         }
     };
 
     $.fn.makeNewznabProviderString = function() {
-        var provStrings = [];
+        const provStrings = [];
 
-        for (var id in newznabProviders) {
-            if (newznabProviders.hasOwnProperty(id)) {
+        for (const id in newznabProviders) {
```
sharkykh: From experience, this kind of loop isn't supported in IE **at all**. https://github.com/nyaadevs/nyaa/pull/246/commits/685936bed9b977dc1404e2461744c1ae6ba5a8fc
(it's supported in Edge, though)
Some people still use this shitty browser.

gui/slick/js/configProviders.js
```diff
@@ -222,34 +218,34 @@ $(document).ready(function(){
...
+            if ({}.hasOwnProperty.call(newznabProviders, id)) {
```
sharkykh: I don't understand why this is needed? (couldn't find an answer)

gui/slick/js/configProviders.js
```diff
@@ -277,205 +273,203 @@ $(document).ready(function(){
...
+            $('#torrentrss_delete').removeAttr('disabled');
         }
     };
 
     $.fn.makeTorrentRssProviderString = function() {
-        var provStrings = [];
-        for (var id in torrentRssProviders) {
-            if (torrentRssProviders.hasOwnProperty(id)) {
+        const provStrings = [];
+        for (const id in torrentRssProviders) {
```
sharkykh: same as above

gui/slick/js/configProviders.js
```diff
@@ -277,205 +273,203 @@ $(document).ready(function(){
...
 
-
     $.fn.refreshProviderList = function() {
-        var idArr = $("#provider_order_list").sortable('toArray');
-        var finalArr = [];
+        const idArr = $('#provider_order_list').sortable('toArray');
+        const finalArr = [];
         $.each(idArr, function(key, val) {
-            var checked = + $('#enable_'+val).prop('checked') ? '1' : '0';
+            const checked = Number($('#enable_' + val).prop('checked')) ? '1' : '0';
```
sharkykh: What is happening here?
Did you wrap it in "Number" to replace the casting to number (the plus sign)?
Can this be the following, dropping the condition completely?
```js
const checked = String(Number($('#enable_' + val).prop('checked')))
```

gui/slick/js/parsers.js
```diff
@@ -18,19 +17,19 @@ $.tablesorter.addParser({
...
+            hd720p: 5,
             '720p': 4, // Custom: Only 720p
-            'hd': 3,
-            'sd': 2,
-            'any': 1,
-            'best': 0
+            hd: 3,
+            sd: 2,
+            any: 1,
+            best: 0
```
sharkykh: I'm not how this is handled in JS, or ES5 for that matter.
Is `replacements['best']` equivalent to `replacements.best`? (see the next line of the code)
If not, this might break `tablesorter`.

gui/slick/js/plotTooltip.js
```diff
@@ -1,18 +1,18 @@
...
-                    url: $("#srRoot").val() + '/home/plotDetails',
+                    url: $('#srRoot').val() + '/home/plotDetails',
                     type: 'GET',
                     data: {
                         show: match[1],
                         episode: match[3],
                         season: match[2]
                     },
-                    success: function (data) {
+                    success: function(data) { // eslint-disable-line object-shorthand
```
sharkykh: You should remove this comment now that you added this to the "xo" configuration,
so when you remove this it will change to the shorthand...

gui/slick/js/ratingTooltip.js
```diff
@@ -1,7 +1,7 @@
-$(function () {
+$(function() {
     $('.imdbstars').qtip({
         content: {
-            text: function() {
+            text: function() { // eslint-disable-line object-shorthand
```
sharkykh: (comment is not needed)

gui/slick/js/sceneExceptionsTooltip.js
```diff
@@ -1,16 +1,16 @@
...
                 text: _('Loading...'),
                 ajax: {
-                    url: $("#srRoot").val() + '/home/sceneExceptions',
+                    url: $('#srRoot').val() + '/home/sceneExceptions',
                     type: 'GET',
                     data: {
                         show: match[1]
                     },
-                    success: function (data) {
+                    success: function(data) { // eslint-disable-line object-shorthand
```
sharkykh: (comment is not needed)

gui/slick/js/configProviders.js
OmgImAlexis: It's the "more correct" way to check instead of calling hasOwnProperty on the object itself you call it on a clean new object.

Ref: http://eslint.org/docs/rules/guard-for-in

gui/slick/js/configProviders.js
OmgImAlexis: This was caused by xo's `--fix`. We could actually use this instead which is easier to read and guards against checked changing to a boolean or int(1/0).

```js
const checked = $('#enable_' + val).is(':checked') ? '1' : '0';
```

OmgImAlexis:
Currently busy with something so if you could please make those changes I think we're good to merge.

sharkykh:
@OmgImAlexis Sure

OmgImAlexis:
If you get an xo warning after changing the for loop just add it to the `package.json` like I did for `object-shorthand` so `xo` knows we want the inverse of the rule.

OmgImAlexis:
@sharkykh we good to merge?

sharkykh:
@OmgImAlexis I'm just lightly testing in the browser

sharkykh:
image

gui/slick/js/apibuilder.js
```diff
@@ -56,52 +58,52 @@ $(document).ready(function() {
...
+                    label: (season === 0) ? 'Specials' : 'Season ' + season
                 }));
             }
         }
     });
 
     // Enable command search
     $('#command-search').typeahead({
-        source: commands, // jshint ignore:line
+        source: commands // eslint-disable-line no-undef
```
sharkykh: Error is here

OmgImAlexis: `commands` should be passed from the mako file for the api as far as I know.

sharkykh: LOL, it's commented out in the mako!

OmgImAlexis: @miigotu that's not how to fix things... 

https://github.com/SickRage/SickRage/blame/develop/gui/slick/views/apiBuilder.mako#L223

OmgImAlexis: I think we should just open an issue about that since it was obviously an issue before now.

sharkykh: Okay, I have a fix for it.
I don't think the command search ever worked on the API builder.
It will now...

OmgImAlexis: It use to about a year or two ago. I don't think anyone really uses it so the issue never came up.

sharkykh: Oh, the comments didn't get updated for me.
Umm, I just pushed 2 commits to fix this and add an auto-scroll-to-command feature.
Might as well fix that now...

sharkykh (approved):

gui/slick/js/apibuilder.js
```diff
@@ -110,12 +110,12 @@ $(document).ready(function() {
         const command = $(this).typeahead('getActive');
 
         if (command) {
-            const commandObj = $('[href="#command-' + command.replace('.', '-') + '"]')
+            const commandObj = $('[href="#command-' + command.replace('.', '-') + '"]');
             commandObj.click();
 
-            setTimeout(function () {
+            setTimeout(function() {
```
OmgImAlexis: If you're using setTimeout or setInterval you can just pass the function instead of wrapping it in another one. Just a little tip to keep in mind.

```js
setTimeout(commandObj.goTo, 250);
```

sharkykh: Completely forgot about this, thanks.
I tried a few things before writing this. I was just happy it finally worked... lol

Issue #3832, opened 2017-06-05
qbittorrent: Unable to send Torrent 


ragtag-fleet:
Hi,

Since updating to qbittorrent v3.3.13, I am unable to send to webui. I think something has changed in webui. Authentication is fine i have tested it, changed it and verified it by logging into the webui. 

Branch/Commit: master/10a0c3012e600ed61cdff6a6108b2bb55062907b
OS: Windows 10
What you did: Manual search eposide to transfer via webui to qbitorrent
What happened: torrent not send to qbitorrent webui
What you expected: torrent send to qbitorrent webui
Logs: https://pastebin.com/pJUPeimj

sharkykh:
https://github.com/qbittorrent/qBittorrent/commit/087856d3d8d6d77881c3d8378a95d80d693af2dd
This was implemented around that version.

We need to account for that change... I look into that, but I don't use qBT.

jabbink:
Tested, and the fix works.

ragtag-fleet:
Could you please push this fix to master branch, I am awaiting for this fix...

jabbink:
In the meantime you can switch to the `develop` branch in the advanced settings.

on `/config/general/`

![image](https://user-images.githubusercontent.com/2461972/26927311-a3d566ee-4c51-11e7-89a1-3969892af89d.png)

Downside: you get this warning on all pages for the time being

![image](https://user-images.githubusercontent.com/2461972/26927333-bc192330-4c51-11e7-858d-5cd423268e0b.png)

ragtag-fleet:
@jabbink This fix has not been pushed to the develop branch either! Did you test on the devlop branch?

jabbink:
It has been on `develop` since 4 days ago: https://github.com/SickRage/SickRage/commit/be0c5eaac12d5736c49e9a54edb4b4cd5b260720

![image](https://user-images.githubusercontent.com/2461972/26973828-8fc4605c-4d18-11e7-8dad-821c8b59f7d8.png)

procdaddy11:
Issue is with qBitTorrent 3.3.13, downgrading to .12 fixed the issue

https://sourceforge.net/projects/qbittorrent/files/qbittorrent-win32/qbittorrent-3.3.12/

.14 is in the works to resolve this issue.

Ambilinc:
downgrading from 3.3.13 to 3.3.12 did it for me.. now my downloads are once again working

jabbink:
[...]

PR #3919, opened 2017-07-10
Lint core.js

OmgImAlexis:
The first commit is from xo's `--fix` so it's mainly just whitespace that was changed. Should be easy enough to review even with the 1k+ changes.

I'll add another few commits to this to bring `core.js` to a 100% linted state. Ci is expected to fail 100% on this PR until all issues with `core.js` are fixed as I removed the ignore from `package.json`.


sharkykh:
I add `?w=1` to the url on Github so the diffs ignore whitespace changes :)

gui/slick/js/core.js
```diff
@@ -333,32 +338,34 @@ var SICKRAGE = {
...
                         $(this).after('<span><img src="' + srRoot + '/images/loading16' + themeSpinner + '.gif"> Saving...</span>');
                         $(this).hide();
                     });
                 },
-                success: function(){
-                    setTimeout(function () {
-                        "use strict";
+                success: function() {
+                    setTimeout(function() {
+                        'use strict';
```
sharkykh: You removed this in `ajaxNotifications` in the previous PR. Is this eligible too?

OmgImAlexis: Yep added that to the next commit already. 👍

gui/slick/js/core.js
```diff
@@ -2067,42 +2078,44 @@ var SICKRAGE = {
...
+                $('#service_order_list > #' + id).remove();
             };
 
             $.fn.refreshServiceList = function() {
-                var idArr = $("#service_order_list").sortable('toArray');
+                var idArr = $('#service_order_list').sortable('toArray');
                 var finalArr = [];
                 $.each(idArr, function(key, val) {
-                    var checked = + $('#enable_'+val).prop('checked') ? '1' : '0';
+                    var checked = Number($('#enable_' + val).prop('checked')) ? '1' : '0';
```
sharkykh: Like previously:
```js
var checked = $('#enable_' + val).is(':checked') ? '1' : '0';
```

gui/slick/js/core.js
```diff
@@ -2975,78 +3028,102 @@ var SICKRAGE = {
...
+                    return;
+                }
 
                 var submitForm = $(
-                    "<form method='post' action='" + srRoot + "/manage/massEdit'>" +
-                        "<input type='hidden' name='toEdit' value='" + editArr.join('|') + "'/>" +
-                    "</form>"
+                    '<form method=\'post\' action=\'' + srRoot + '/manage/massEdit\'>' +
+                        '<input type=\'hidden\' name=\'toEdit\' value=\'' + editArr.join('|') + '\'/>' +
+                    '</form>'
```
sharkykh: Using double quotes in the html itself would be much more readable in this case.

gui/slick/js/core.js
```diff
@@ -3342,91 +3441,104 @@ var SICKRAGE = {
...
+                        }; // Quality
+                    } else {
+                        compactExtract[4] = function(node) {
+                            return $(node).find('span').text().toLowerCase();
+                        }; // Quality
+                    }
 
-                        return compactExtract;
+                    return compactExtract;
+                })(),
```
sharkykh: Is the change from `}()),` to this significant?

OmgImAlexis: It's meant to be a cleaner way to handle a iife. I also removed the function's name.

sharkykh:
OK, I'm done. Looks good!

Also, worth noting, the previous PR and this PR will make your yarn/webpack PR much easier to review!

gui/slick/js/core.js
sharkykh: The function in textExtraction didn't have a name, but ok!

sharkykh:
Also, a fix for the comments for functions being on the incorrect line:
image

Find [Regex]:
```regex
function\((.*?)?\) {(.*)?(\n\s*return.*\n\s*\}[,;]?)\s*//\s*(.*)
```
Replace [Regex]:
```regex
function\(\1\) { // \4\2\3
```

(Tested on `core.js`)

**Edit:** Fixed regex to append the current line comment.

gui/slick/js/core.js
OmgImAlexis: Oh maybe we're talking about different parts. Let me check.

gui/slick/js/core.js
```diff
@@ -82,14 +83,14 @@ function shiftReturn(array) {
...
 var __ = _; // Moves `underscore` to __
-_ = function(str) {
+_ = function(str) { // eslint-disable-line no-global-assign
     return str;
 }; // Temporary (gets replaced)
 
 var SICKRAGE = {
     common: {
         init: function() {
-            (function init() {
```
sharkykh, line 92: Yeah, you were talking about this...

gui/slick/js/core.js
OmgImAlexis: Ah yep I was thinking of another section further up. It's still a cleaner way to handle an iife.

OmgImAlexis:
Ref: http://benalman.com/news/2010/11/immediately-invoked-function-expression/

```
  (function(){ /* code */ }()); // Crockford recommends this one
  (function(){ /* code */ })(); // But this one works just as well
```

OmgImAlexis:
Does appveyor not have a setting to fail the build as soon as anything exists with a failure?

sharkykh:
@OmgImAlexis I think it only has the "fail next environments if any of the environments fail", an environment for us being a test for 32 bit or 64 bit.
And what I put in is what they recommend in their docs...
`set -e` is much better!
In any case, it works for both of them now.

OmgImAlexis:
This may help. http://help.appveyor.com/discussions/questions/851-is-it-possible-to-programmatically-eg-via-ps-or-cmd-fail-a-build

gui/slick/js/core.js
```diff
@@ -2270,12 +2236,13 @@ var SICKRAGE = {
...
-                    stickyHeaders_offset: 50,
-                    filter_saveFilters: true,
-                    filter_functions: {
-                        5: function(e, n, f) {
+                    filter_columnFilters: true, // eslint-disable-line camelcase
+                    filter_hideFilters: true, // eslint-disable-line camelcase
+                    stickyHeaders_offset: 50, // eslint-disable-line camelcase
+                    filter_saveFilters: true, // eslint-disable-line camelcase
+                    filter_functions: { // eslint-disable-line camelcase
+                        // @TODO: The fuck is e, n, f? We need to start naming variables properly.
```
sharkykh: #tablesorter_stuff
https://mottie.github.io/tablesorter/docs/example-widget-filter-custom.html#notes

sharkykh:
Yep, that's what I did. They put the code in https://www.appveyor.com/docs/build-configuration/#script-blocks-in-build-configuration

> If you need to check execution results of commands in the middle of PowerShell script you can verify `$LastExitCode` after calling a command and terminate earlier, for example:
> 
> ```ps
> vstest.console myassembly1.dll
> if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
> 
> vstest.console myassembly2.dll
> if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
> ```

gui/slick/js/core.js
```diff
@@ -2749,13 +2750,14 @@ var SICKRAGE = {
...
-                var forEpisode = $(this).attr('data-for-episode');
-                var m = $(this).val().match(/^(\d+)x(\d+)$/i);
-                var onlyEpisode = $(this).val().match(/^(\d+)$/i);
-                var sceneSeason = null,
-                    sceneEpisode = null;
-                var isValid = false;
+                let forSeason = $(this).attr('data-for-season');
+                let forEpisode = $(this).attr('data-for-episode');
+                let m = $(this).val().match(/^(\d+)x(\d+)$/i);
+                let onlyEpisode = $(this).val().match(/^(\d+)$/i);
```
sharkykh, line 2761: ~both~ the four of these can be const

OmgImAlexis: Yep, going back through now. I just swtiched over from var since let/const is better for scope.

gui/slick/js/core.js
```diff
@@ -3846,8 +3848,7 @@ var SICKRAGE = {
                 }
 
                 // Also toggle the add show button
-                if (($('#rootDirs option:selected').length || ($('#fullShowPath').length && $('#fullShowPath').val().length)) &&
-                    ($('input:radio[name=whichSeries]:checked').length) || ($('input:hidden[name=whichSeries]').length && $('input:hidden[name=whichSeries]').val().length)) {
+                if (($('#rootDirs option:selected').length || ($('#fullShowPath').length && $('#fullShowPath').val().length)) && ($('input:radio[name=whichSeries]:checked').length) || ($('input:hidden[name=whichSeries]').length && $('input:hidden[name=whichSeries]').val().length)) {
```
sharkykh: That line is too long. What about http://eslint.org/docs/rules/max-len ?

gui/slick/js/core.js
```diff
@@ -3750,19 +3837,18 @@ var SICKRAGE = {
...
+                // If not then it's unknown
                 } else {
                     $('#displayText').html(sampleText.replace('||', '??'));
                 }
 
-                // also toggle the add show button
-                if (($("#rootDirs option:selected").length || ($('#fullShowPath').length && $('#fullShowPath').val().length)) &&
-                    ($('input:radio[name=whichSeries]:checked').length) || ($('input:hidden[name=whichSeries]').length && $('input:hidden[name=whichSeries]').val().length)) {
+                // Also toggle the add show button
+                if (($('#rootDirs option:selected').length || ($('#fullShowPath').length && $('#fullShowPath').val().length)) && ($('input:radio[name=whichSeries]:checked').length) || ($('input:hidden[name=whichSeries]').length && $('input:hidden[name=whichSeries]').val().length)) {
```
sharkykh: That line is too long. What about http://eslint.org/docs/rules/max-len ?

gui/slick/js/core.js
```diff
@@ -4038,33 +4129,38 @@ var SICKRAGE = {
...
+        UTIL.exec('common');
         UTIL.exec(controller);
         UTIL.exec(controller, action);
     }
 };
 
 // Handle js-gettext + load javascript functions
-var gt = null, _n = null;
+let gt = null;
+let _n = null;
```
sharkykh: This might break the gettext functions. They should remain `var`s, IMO.
They are both called everywhere, even outside the scope of `core.js`.
So they need to be in the global scope. (`_` + `_n` both use `gt`)

gui/slick/js/core.js
OmgImAlexis: Most likely going to be moving it into a function tomorrow morning. Heading to bed so I'll push the last few commits after I wake up.

gui/slick/js/core.js
OmgImAlexis: I'll move them to `window.gt` and `window._n` then.

gui/slick/js/core.js
sharkykh: GN

OmgImAlexis:
@sharkykh would you mind reviewing the latest changes when you get a chance. I want to see if the way I setup get text that it still works since it's now bound to the window instead of just being a global var.

I've also added `eslint-disable-line` on a bunch of lines as I plan on going back to fix as many of them as I can after this PR.

OmgImAlexis:
Also thoughts on removing uglify? With webpack we be using another type of minifier anyways since uglifyjs doesn't work well with ES6 or even ES6(let/const).

gui/slick/js/core.js
```diff
@@ -1558,16 +1556,17 @@ var SICKRAGE = {
...
-                        $('#unpack').qtip('toggle', false);
-                        break;
-                    case 1: // Unpack
-                        $('#content_unpack').fadeIn('fast', 'linear');
-                        isRarSupported();
-                        break;
+                const value = parseInt(this.value, 10);
+
+                // Treat as video
+                if (value === 2) {
```
sharkykh:
```js
// Treat as video or ignore
if (value === 2 || value === 0) {
```

Edit: fixed.

sharkykh:
> @sharkykh would you mind reviewing the latest changes when you get a chance. I want to see if the way I setup get text that it still works since it's now bound to the window instead of just being a global var.

I'll probably test and maybe push some stuff later today...

> Also thoughts on removing uglify? With webpack we be using another type of minifier anyways since uglifyjs doesn't work well with ES6 or even ES6(let/const).

We still need something like it until the webpack branch is live and we ditch the old setup completely.
So maybe [uglify-es](https://www.npmjs.com/package/uglify-es) or [uglify-js-es6](https://www.npmjs.com/package/uglify-js-es6)?

For grunt, https://stackoverflow.com/questions/44848273/how-to-integrate-uglify-es-in-grunt

OmgImAlexis:
Yep `uglify-es` is the way to go. I was trying to think of what it was called since it's actually just the harmony branch uglify.

Would you mind swapping it over? I think `core.js` is done after that.

sharkykh:
@OmgImAlexis I did. Harmony branch is uglify-es : https://github.com/gruntjs/grunt-contrib-uglify/blob/harmony/package.json
Going to sleep now...

sharkykh (dismissed):
I also saw some more `var`s around that line..

This is not a full review. This is just something that caught my eye.

gui/slick/js/core.js
```diff
@@ -1176,7 +1173,8 @@ var SICKRAGE = {
 
             function loadShowNotifyLists() {
                 $.getJSON(srRoot + '/home/loadShowNotifyLists', function(list) {
-                    var html, s;
+                    let html = null;
+                    let s = null; // @TODO: What is an "s"?
```
sharkykh: Okay, I went through it.
So, basically `s` is just a variable for the for loops.
What this code does is get [the data of your tv shows with the corresponding emails list and prowl list](https://github.com/SickRage/SickRage/blob/a240f9f5293c19c2a34350ec55b96808048766f9/sickbeard/webserve.py#L1124-L1149), for example:
```json
{
  [...]
  "5": {
    "list": "",
    "id": 5,
    "prowl_notify_list": "",
    "name": "The Expanse"
  },
  [...]
  "8": {
    "list": "",
    "id": 8,
    "prowl_notify_list": "",
    "name": "The Grand Tour (2016)"
  },
  [...]
}
```

In order to populate them in the lists for the Email / Prowl notifiers, like so...
image

gui/slick/js/core.js
```diff
@@ -11,7 +11,7 @@ const loading = '<img src="' + srRoot + '/images/loading16' + themeSpinner + '.g
 
 let srPID = getMeta('srPID');
 
-var configSuccess = function() {
+function configSuccess() {
```
OmgImAlexis: I'd actually recommend using `const configSuccess = function(){}` since that's what we'll use in webpack when we move the functions to a util file and export them.

gui/slick/js/core.js
```diff
@@ -2240,35 +2246,38 @@ var SICKRAGE = {
...
+                                '>': function(a, b) {
+                                    return a > b;
+                                }
+                            };
+
+                            if (filterInput === '') {
                                 test = true;
                             } else {
-                                var result = f.match(/(<|<=|>=|>)\s+(\d+)/i);
+                                let result = filterInput.match(/(<|<=|>=|>)\s?(\d+)/i);
```
OmgImAlexis, line 2272: Does this change the regex matching?

sharkykh: In a word: **No**.
More specifically, this is for the better.
As before this change, it wouldn't catch filters like `< 50`, only `<50`.
This can also be an asterisk (for 0 or more spaces), but I don't think that's even needed.

BTW I tested it along with the new "compare mapper" I added, so I know it doesn't break anything.

gui/slick/js/core.js
sharkykh: Oh, okay. I just looked at the other ones like the meta functions and matched that. It just annoyed me... lol
But I guess the others **are needed in the global scope** (getMeta and the likes)

gui/slick/js/core.js
```diff
@@ -2313,37 +2322,37 @@ var SICKRAGE = {
...
                     }
                 });
 
                 // When posters are small enough to not display the .show-details
                 // table, display a larger poster when hovering.
-                var posterHoverTimer = null;
+                let posterHoverTimer = null;
                 $('.show-container').on('mouseenter', function() {
-                    var poster = $(this);
+                    const poster = $(this);
```
OmgImAlexis: I think we should stop using `$(this)` and instead use `$(event.currentTarget)` that way if we want we can switch them all to using arrow functions later on.

gui/slick/js/core.js
```diff
@@ -2858,8 +2863,8 @@ var SICKRAGE = {
             });
 
             $('#addSceneName').click(function() {
```
OmgImAlexis: This needs to be on click, not sure why it wasn't ever changed over.

gui/slick/js/core.js
OmgImAlexis: Are we able to use template strings for most browsers?

sharkykh: [No IE, no Android](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Browser_compatibility). I guess not for now.

gui/slick/js/core.js
sharkykh: BTW, forgot to reply. Tested and the functions work.

gui/slick/js/core.js
OmgImAlexis: 😞  Guess I'll swap this over to double quotes in the HTML then.

gui/slick/js/core.js
OmgImAlexis: Might want to check the other regex strings around that area then since some srae still using the `s+` instead of `s?`.

OmgImAlexis:
If you're busy I'll add these changes in the next hour, just need to get lunch.

sharkykh:
@OmgImAlexis I just need to go to sleep (5.30am). If you have the time, go ahead :)

gui/slick/js/core.js
sharkykh: From what I can tell, it's only the `-` or `to` filter that still uses `\s+` and `\s`. It can be `\s?`.
ln2280:
```js
result = filterInput.match(/(\d+)\s(-|to)\s+(\d+)/i); // Before
result = filterInput.match(/(\d+)\s?(-|to)\s?(\d+)/i); // After
```

OmgImAlexis: Okay looks like I hadn't pulled the latest changes down. All good now.

OmgImAlexis:
I think I got all the changes. If you could give it a once over I'm happy with merging.

sharkykh:
We've managed to keep this (mostly) centered around `core.js`. I'd suggest we keep it that way and move the changes to other files to a small, separate PR. I'll do that then we'll merge them both.

#3922

sharkykh (approved):
I think I'm good with this, if you want to merge.

sharkykh:
Erm.. You should've squashed the commits...

OmgImAlexis:
Yeah thought about that after I merged. Oh well.

gui/slick/js/core.js
```diff
@@ -2586,47 +2603,46 @@ var SICKRAGE = {
...
             });
 
-            // initially show/hide all the rows according to the checkboxes
-            $("#checkboxControls input").each(function() {
-                var status = $(this).prop('checked');
-                $("tr." + $(this).attr('id')).each(function() {
-                    if(status) {
+            // Initially show/hide all the rows according to the checkboxes
+            $('#checkboxControls input').each(function() {
+                const status = $(this).checked;
```
sharkykh, line 2643: Hey @OmgImAlexis, just so you know and for future reference:
It appears `$('selector').checked` is not the same as `$('selector').prop('checked')`.
This is what caused the issue described in https://github.com/SickRage/SickRage/issues/3832#issuecomment-315377641
I'll look for all these changes and change them back.

```html
<input type="checkbox" id="good" checked="checked">
```

```js
> $('#good')
< [input#good, context: document, selector: "#good"]
> $('#good').checked
< undefined
> $('#good')[0].checked
< true
> $('#good').prop('checked')
< true
```

PR #3921, opened 2017-07-12
Fix issue with variable in static_url call, add more static calls

miigotu:
static_url works with normal paths to add the root/port, which prevents issues with reverse proxy
@sharkykh can you work on the other sbRoot calls maybe?


sickbeard/webserve.py
```diff
@@ -169,6 +169,7 @@ def __init__(self, *args, **kwargs):
         self.startTime = time.time()
 
         super(BaseHandler, self).__init__(*args, **kwargs)
+        self.include_host = True
```
miigotu, line 172: This makes static_url add the request url, port, and protocol

gui/slick/views/addShows.mako
```diff
@@ -12,7 +12,7 @@
     </div>
     <div class="row">
         <div class="col-md-12">
-            <a href="${srRoot}/addShows/newShow/" id="btnNewShow" class="btn btn-large">
+            <a href="${ static_url("addShows/newShow") }" id="btnNewShow" class="btn btn-large">
```
miigotu: static_url calls to non-static assets just adds proto, originating ip or domain, and port. The v= cachebreaker is not added as it is autodetected as a route rather than a static asset

miigotu:
...

gui/slick/views/apiBuilder.mako
```diff
@@ -76,7 +77,7 @@
         <link rel="stylesheet" type="text/css" href="${ static_url('css/print.css') }" />
 
         %if sickbeard.THEME_NAME != "light":
-            <link rel="stylesheet" type="text/css" href="${ static_url('css/${sickbeard.THEME_NAME}.css') }" />
+            <link rel="stylesheet" type="text/css" href="${ static_url(urljoin('css/', '.'.join((sickbeard.THEME_NAME, 'css')))) }" />
```
miigotu: This fixes dark theme not loading

gui/slick/views/layouts/main.mako
```diff
@@ -95,12 +94,13 @@
         <link rel="stylesheet" type="text/css" href="${ static_url('css/country-flags.css') }"/>
 
         % if sickbeard.THEME_NAME != "light":
-            <link rel="stylesheet" type="text/css" href="${ static_url('css/${sickbeard.THEME_NAME}.css') }" />
+            <link rel="stylesheet" type="text/css" href="${ static_url(urljoin('css/', '.'.join((sickbeard.THEME_NAME, 'css')))) }" />
```
miigotu: Fixes dark theme not loading

gui/slick/views/layouts/main.mako
```diff
@@ -244,7 +245,7 @@
                                         <li><a href="${srRoot}/logout" class="confirm logout"><i class="fa fa-fw fa-sign-out"></i>&nbsp;${_('Logout')}</a></li>
                                     % endif
                                     <li role="separator" class="divider"></li>
-                                    <li><a href="${ static_url("home/status") }"><i class="fa fa-fw fa-info-circle"></i>&nbsp;${_('Server Status')}</a></li>
+                                    <li><a href = "${ static_url("home/status", include_version=False) }"><i class="fa fa-fw fa-info-circle"></i>&nbsp;${_('Server Status')}</a></li>
```
OmgImAlexis: There shouldn't be a space `href = "`

miigotu:
@OmgImAlexis @sharkykh 
This adds ip/url, port, protocol to all without using srRoot, please test/review

miigotu:
Idk why it hid my comments as outdated reviews... those comments are still valid.

gui/slick/views/layouts/main.mako
miigotu: Yeah, bad regex. It's fixed in later commit

sharkykh:
> Idk why it hid my comments as outdated reviews... those comments are still valid.

Because you rebased, I think.

***

I haven't had the time to thoroughly review this, but, @miigotu, `static_url` is not Flask's `url_for`... I don't know if I like this type of usage.
~http://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.reverse_url  maybe?~

sharkykh:
What I mean by don't like, is that it's prone to break if we change something with the static files configuration.

sickbeard/webserve.py
sharkykh: This feels like bad idea.
Why have it at all? Relative paths are good.

If the `static_url` doesn't work with reverse proxies out of the box in some way,
we should fix it by subclassing something or adding a custom function...
That's my opinion.

gui/slick/views/layouts/main.mako
```diff
@@ -101,7 +101,7 @@
 
         % if sickbeard.CUSTOM_CSS:
             ## TODO: check if this exists first
-            <link rel="stylesheet" type="text/css" href="${static_url('ui/custom.css')}" />
+            <link rel="stylesheet" type="text/css" href="${static_url('ui/custom.css', include_version=False)}" />
```
sharkykh: Nope. this is my bad.
I shouldn't have used `static_url` at all for this, simply because it's not possible [with the way it's implemented](https://github.com/SickRage/SickRage/blob/develop/sickbeard/webserve.py#L653-L658).

gui/slick/views/addShows.mako
```diff
@@ -12,7 +12,7 @@
     </div>
     <div class="row">
         <div class="col-md-12">
-            <a href="${srRoot}/addShows/newShow/" id="btnNewShow" class="btn btn-large">
+            <a href="${static_url("addShows/newShow", include_version=False)}" id="btnNewShow" class="btn btn-large">
```
sharkykh, line 15: The fact that you add the `include_version=False` argument to these types of calls to `static_url`, after saying
>static_url calls to non-static assets just adds proto, originating ip or domain, and port. **_The v= cachebreaker is not added as it is autodetected as a route rather than a static asset_**

just proves the point that it's not intended to be used this way.
(since `static_url` is derived from `StaticFileHandler` class and these are not static files)

gui/slick/views/layouts/main.mako
```diff
@@ -219,31 +221,39 @@
...
-                                    <li><a href="${srRoot}/home/restart/?pid=${sbPID}" class="confirm restart"><i class="fa fa-fw fa-repeat"></i>&nbsp;${_('Restart')}</a></li>
-                                    <li><a href="${srRoot}/home/shutdown/?pid=${sbPID}" class="confirm shutdown"><i class="fa fa-fw fa-power-off"></i>&nbsp;${_('Shutdown')}</a></li>
+                                    <li><a href="${static_url("home/updateCheck?pid={}".format(sbPID), include_version=False)}><i class="fa fa-fw fa-wrench"></i>&nbsp;${_('Check For Updates')}</a></li>
+                                    <li><a href="${static_url("home/restart/?pid={}".format(sbPID), include_version=False)}" class="confirm restart"><i
+                                        class="fa fa-fw fa-repeat"></i>&nbsp;${_('Restart')}</a></li>
+                                    <li><a href="${static_url("home/shutdown/?pid={}".format(sbPID), include_version=False)}" class="confirm shutdown"><i class="fa \
+                                    fa-fw \
+
+                                    fa-power-off"></i>&nbsp;${_('Shutdown')}</a></li>
```
sharkykh, line 251: Weird looking thing with the line breaks. I guess it was caused by something automatic.
Just have it like the restart above it.

sharkykh (commented):
**_Unless_** the changes I've made with my PR had caused breakages for reverse proxies,
I don't think I get the reasoning behind these changes.

Although I do agree on some of the changes, like the fix to dark theme and the change in `self.arguments['sbHttpsEnabled']`.

miigotu:
Im going to merge this as-is so I can release to master. We can change them to use reverse_url later

sharkykh:
@miigotu 
Removing the **trailing forward slash** in the links to the config sections (for example) caused the config forms to stop saving.
It tries to POST to `/config/saveNotifications` instead of `/config/notifications/saveNotifications`, raising a 404 error and exception that `Config` doesn't have `saveNotifications`.

Took me a while to figure that one out.

sharkykh:
@miigotu Don't release a new version before fixing this...

PR #3926, opened 2017-07-14
core.js: Switch .checked on jQuery elements back to .prop('checked')

sharkykh:
It appears `$('selector').checked` is not the same as `$('selector').prop('checked')`.
```html
<input type="checkbox" id="good" checked="checked">
```
```js
> $('#good')
< [input#good, context: document, selector: "#good"]
> $('#good').checked
< undefined
> $('#good')[0].checked
< true
> $('#good').prop('checked')
< true
```

This is what caused the issue described in https://github.com/SickRage/SickRage/issues/3832#issuecomment-315377641:

> I've been using develop, but now downloaded episodes aren't showing anymore under develop. 
>  
> Master:
> image
> Develop:
> image

This PR reverts only those changes.

@OmgImAlexis


sharkykh:
> I was just saying that I had switched to develop to skirt around the qBT issue for now, and this new issue popped up.
> ...
> Anyway.. thanks for looking into it!

@mickloud 
We (developers) can only do and test so much.
So we only find out about stuff that's broken when it gets merged to master, and that's bad.
This is why we need more users actually trying out the develop branch...

OmgImAlexis:
Actually this isn't the right way to fix it. Setting `checked` on an input will default to true. `checked="checked"` is an old compatibility way of setting it and isn't recommended at all in HTML5.

OmgImAlexis:
If we just switch them to `[0].checked` it'll fix the issue. I'll look into why it's not working like that since it should work on the jquery element without needing to select the first element with [0].

OmgImAlexis:
The other option would be to use `.is(':checked');` the reason I swapped it out in the first place is that `prop('checked')` is the slowest way to do it and is prone to false positives where as `.checked` and `.is(':checked')` don't have that issue.

sharkykh:
> If we just switch them to [0].checked it'll fix the issue. I'll look into why it's not working like that since it should work on the jquery element without needing to select the first element with [0].

https://stackoverflow.com/a/3334945/7597273
?

OmgImAlexis:
Keep in mind that's from 2010. It's changed a LOT since then.

sharkykh:
Oops, missed that

OmgImAlexis:
Ah so looks like jQuery no longer proxies the DOM element to `.checked` that's why you need to use `[0]` or `.get(0)`. I'm fine with any changes. I'll mark this as approved for now and look into the best way later on.

OmgImAlexis (approved):

sharkykh:
FTR I'm also fine with `.is(':checked')`

OmgImAlexis:
If you'd like to switch to `.is(':checked')` then go ahead otherwise let's merge this. 👍

PR #3433, opened 2017-03-19
Cleanup some templating, remove resorting of show lists and iterable …

miigotu:
Don't display shows in the showList if they are in the queue to be removed
    Cleanup some templating, remove resorting of show lists and iterable container type changes
    Update some bower/npm packages,
    Fix shecdule_missed_range default to be like min_max

Fixes #3432 too [edited by @sharkykh]

PR #3930, opened 2017-07-16
add ava testing

OmgImAlexis:
This adds ava so we can start adding tests for the js files finally.


package.json
```diff
@@ -11,10 +11,13 @@
     },
     "homepage": "https://github.com/SickRage/SickRage#readme",
     "scripts": {
-        "test": "xo && grunt",
+        "test": "xo --verbose && ava --verbose && grunt",
```
sharkykh: Seeing how grunt is no longer used for testing, I'd suggest adding a `"build": "grunt"` script and removing it from `test`.

I'll eventually move the release commands into a Python script and remove Gruntfile.js completely (after it's no longer used for building the assets)
I've read that article about "Why we shouldn't use Grunt & Gulp?" multiple times and it bugs me.

package.json
```diff
@@ -28,11 +31,17 @@
         "grunt-contrib-watch": "^1.0.0",
         "grunt-exec": "^2.0.0",
         "grunt-po2json": "git+https://github.com/rockykitamura/grunt-po2json.git",
+        "jquery": "^3.2.1",
```
sharkykh, line 34: Why is `jquery` added?

package.json
```diff
@@ -28,11 +31,17 @@
...
         "load-grunt-tasks": "^3.5.2",
         "npm-check-updates": "^2.10.5",
         "snyk": "^1.28.1",
         "xo": "^0.18.2"
     },
+    "ava": {
+        "require": [
+            "./test/helpers/setup-browser-env.js"
```
sharkykh: I don't like that it [adds the `test` folder again](https://github.com/SickRage/SickRage/pull/3433/commits/b210afdf17ff231d81d0a61fd8fde40387f5be18), but whatever.

test/index.js
```diff
@@ -0,0 +1,11 @@
+import test from 'ava';
+
+test.failing('getMeta', t => {
+    const meta = document.createElement('meta', {
+        'data-var': 'srRoot',
+        content: 'http://localhost:8081/'
```
sharkykh: I realize this is just a placeholder,
but a real value would be `/sickrage`.

package.json
OmgImAlexis: Sounds good to me. I only left it there since I wasn't 100% sure on what grunt was used for.

package.json
OmgImAlexis: This is so we can use it in testing with ava. All the other bower deps will need to be added eventually so testing and webpack can work.

test/index.js
OmgImAlexis: I'll fix that.

package.json
OmgImAlexis: If we use `tests/ava` or `tests/js` it means more config and I'd rather just be able to use it out of the box without needing to configure anything that's why I just used `test`.

package.json
sharkykh: A remnant of the past. :stuck_out_tongue:

package.json
sharkykh: Okay, but in [`bower.json` the jQuery version is pinned to `<3.0`](https://github.com/SickRage/SickRage/blob/b5ae6efbe48c8150b4120b6ba4881024c049d6a3/bower.json#L12).

Just take that into consideration, because you might skip over it when you implement webpack.
I'm not sure that code is completely jQuery 3 ready.
(although switching `.bind()` with `.on()` was a good start...)

package.json
sharkykh: Don't get me wrong, I can live with it. So I'm just suggesting here;
Seeing how you're already adding an "ava" section in package.json,
I don't see a why you wouldn't also add a single line (fine, 3, lol)
to avoid having to add a new folder to the already-crowded (IMO) root folder.

package.json
OmgImAlexis: Everything **SHOULD** work with jQuery 3+ but we'll keep this in mind. We may have to add the migrate plugin until we can switch it all.

OmgImAlexis:
That last commit should fix all the review issues. @sharkykh I'm not 100% sure on what todo to fix the other commits showing because of the cherry pick so could you or @miigotu fix that?

package.json
sharkykh: Tried that some time ago. Most of the issues were on vendored packages.
I remember `qtip2`, because I tried to find a good alternative but couldn't...

OmgImAlexis: Oh yeah, there is an alternative to qtip2 that doesn't use jquery I just can't remember the name.

sharkykh:
Done.

sharkykh (approved):
Don't forget to squash :stuck_out_tongue:

package.json
```diff
@@ -11,10 +11,13 @@
     },
     "homepage": "https://github.com/SickRage/SickRage#readme",
     "scripts": {
-        "test": "xo && grunt",
+        "test": "xo --verbose && ava tests/js/**/* --verbose",
```
sharkykh: Umm, the `tests/js/**/*` isn't necessary here, because you have it in the config?

OmgImAlexis: Oh whoops. Forgot to remove that.

PR #3944, opened 2017-07-18
Fix UI bugs

sharkykh:
## Fixes bugs caused by #3921 
- [x] **Recent Shows:** Mako statement within a Mako statement
- [x] **Recent Shows:** Missing closing double quotes of `href` attribute.
- [x] **Shutdown link:** Malformed HTML
- [x] **Check for Updates link:** Missing closing double quotes of `href` attribute.
- [x] **Logout link:** Missing closing double quotes of `href` attribute.
- [x] **"+X" (snatched in footer) link:** Missing closing double quotes of `href` attribute.
- [x] **Routes with forms:** Need to have a trailing slash in links (breaks forms if doesn't)
- [x] **Routes with forms:** Force trailing slash on index with `@tornado.web.addslash`


OmgImAlexis:
Is this really the best idea? I don't even know why we're using the static handler at all for non-static files. 😞  Routes aren't static files...

sharkykh:
@OmgImAlexis 
I remember saying so myself...

miigotu:
static_url works for dynamic routes in tornado as well as long as the use_version=False.  We can think about replacing it with a backend implementation another time to simplify it but for now it just works and removes the need for srRoot/sbPort/etc

miigotu:
@sharkykh 
http://www.lexev.org/en/2014/set-url-for-tornado-handlers/
static_url for static, and reverse for reverse, woooo

sharkykh:
@miigotu 
I know. What I don't like about it is the fact you have to define the "name" parameter for each Handler. This might not work with the current setup of handlers.

Doing it properly requires overriding classes, and I don't feel like doing that at the moment.

miigotu:
You only need to name it like 'ui' and call reverse_url('ui', 'get_messages') or something like that. idk, release isnt gonna happen for a bit yet so we will see what we come up with.

PR #3953, opened 2017-08-06
Fix Javascript on Safari

beschoenen:
Fixes Safari errors which made it unable to render the displayShow page.


gui/slick/js/core.js
```diff
@@ -2,7 +2,8 @@ function getMeta(pyVar) {
     return $('meta[data-var="' + pyVar + '"]').data('content');
 }
 
-const srRoot = getMeta('srRoot');
+srRoot = getMeta('srRoot');
```
OmgImAlexis: This essentially makes srRoot a var again. 

Safari supports using `const`?
https://caniuse.com/#feat=const

OmgImAlexis (changes requested):

gui/slick/js/core.js
beschoenen: It does, but the notifications web call in `ajaxNotifications.js` is not working because it's trying to get `http://localhost:8081/home/[object HTMLObject]/ui/get_messages`. This fixed that.

Do you know a better way to fix this?

miigotu: @OmgImAlexis I think he is right here, setting it const makes it not update on subsequent loads like after they are changed when saving config, or a restart, or before the server is fully booted.

OmgImAlexis: No the whole js file is loaded on page load so that's not the case. 

The reason you're getting `[object HTMLObject]` is because of an element on the page overriding the `window['srRoot']` namespace.

Whatever page you're on must have an element with the ID of `srRoot`.

Ref: https://stackoverflow.com/questions/3434278/do-dom-tree-elements-with-ids-become-global-variables

OmgImAlexis: All removing the const is doing is overriding the global var. This is just masking the issue.

beschoenen:
@OmgImAlexis Ok, that worked. Not sure why the `core.min.js` is still "changed."

sharkykh:
@beschoenen 
What node version are you using?
```
$ node --version
v8.1.4
```

I just used `grunt uglify:core` on this branch and the only diff is `displayShow.mako`:
```diff
$ git diff origin/develop
diff --git a/gui/slick/views/displayShow.mako b/gui/slick/views/displayShow.mako
index d33c9bee1..017a8392b 100644
--- a/gui/slick/views/displayShow.mako
+++ b/gui/slick/views/displayShow.mako
@@ -27,7 +27,6 @@

             <div class="row">
                 <div class="col-md-12" style="margin: 5px 0;">
-                    <input type="hidden" id="srRoot" value="${srRoot}" />
                     <div class="form-inline">
                         <label for="pickShow">${_('Change Show')}:</label>
                         <div class="pick-show-group input350">


```

beschoenen:
I'm using node `8.2.1` I guess it's not working as I want it to. Feel free to commit here.

beschoenen:
#3966

Issue #3915, opened 2017-07-09
Add providers YggTorrent & SpaceTorrent


PxReaz:
Branch/Commit: master/96b53dee713dc35ebf295d0f37a67b2a36fb4ea3
OS: Linux-3.16.0-4-amd64-x86_64-with-debian-8.8

Hello!

this is not really an issue but can you add a provider in the next update?
or help me to add it manually i'm not sure how to do that

it's a new provider that is here to replace t411 who's been closed recently.

http://yggtorrent.com

Thanks!

HLFH:
spacetorrent.pw would also be interesting to add as a replacement of t411.

PxReaz:
Where did you get the invitation for this provider?

HLFH:
From https://www.wareziens.net/forum/, I got the email contact that is using ProtonMail. If you have a good profile, you were able to get the invitation code. Now it is more advised to create your own invitation topic, ask for a SPT.pw invitation, and if your profile match, you'll get it.

PxReaz:
Ok thanks! Hope they will add the providers in the next update!

sharkykh:
1. It seems yggtorrent is now under maintenance. I'll check again later.
2. I'm not going to work on spacetorrent.pw because it's a private tracker.

PxReaz:
Okay thanks !

Amonbofils:
ygg is back if you want to try :D

HLFH:
@sharkykh I can invite you on spacetorrent.pw if you want to work on it with some other SickRage users.

antipiot:
@HLFH Hey mate, if you don't bother sharing, i'm interested aswell :-)i was a daily user ot T411 with sickrage

dmsimard:
Relevant ? 
https://github.com/Jackett/Jackett/commit/35103206cfdb382d247ce089209b12abec4c5ada#diff-2eee82756efe0f23652a6a71efa6e983
https://github.com/Jackett/Jackett/commit/14f86107e14e3ec3823479d160cf91ad62365076#diff-2eee82756efe0f23652a6a71efa6e983

jgrossrieder:
I was wanting to implement this provider on SickRage, I did implement a big part of the provider, but I can't find any documentation on how to setup a development environment to test my implementation. Tried to request it through IRC couple of times, but no one was here to answer.

adaur:
I'll take a look at this asap

adaur:
[...]

PR #3956, opened 2017-08-06
Add provider YggTorrent

adaur:
Fixes #

- Add new French provider YggTorrent
- It is not possible to sort torrent during search right now, I'll update later when this is possible


sickbeard/providers/yggtorrent.py
```diff
@@ -0,0 +1,158 @@
...
+
+        return True
+
+    def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-many-locals, too-many-branches
+        results = []
+        if not self.login():
+            return results
+
+        # Units
+        units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB']
```
sharkykh: Default units are identical, so there's no need for this variable.
https://github.com/SickRage/SickRage/blob/4509c8975c452bc139643e1b4cd73d73d079ffee/sickrage/helper/common.py#L209

sickbeard/providers/yggtorrent.py
```diff
@@ -0,0 +1,158 @@
...
+                            size = convert_size(torrent_size, units=units) or -1
+
+                            item = {'title': title, 'link': download_url, 'size': size, 'seeders': seeders, 'leechers': leechers, 'hash': ''}
+                            if mode != 'RSS':
+                                logger.log('Found result: {0} with {1} seeders and {2} leechers'.format
+                                           (title, seeders, leechers), logger.DEBUG)
+
+                            items.append(item)
+                        except StandardError:
+                            continue
```
sharkykh: To keep it in-line with other providers I've seen, this should be below the last scraping variable.
In this case, after the `size` variable.

I guess it's worth noting that the way it's currently done you'll never get notified if the provider suddenly broke (like, when the website's layout changes).
I think you should catch [these exceptions](https://github.com/SickRage/SickRage/blob/4509c8975c452bc139643e1b4cd73d73d079ffee/sickbeard/providers/limetorrents.py) outside the loop (remove the current `try...except` clause).
And instead of continuing, log an error (logging an error automatically adds the traceback info)
```python
logger.log('Failed parsing provider {}.'.format(self.name), logger.ERROR)
```

sickbeard/providers/yggtorrent.py
```diff
@@ -0,0 +1,158 @@
...
+
+            # For each search mode sort all the items by seeders if available
+            items.sort(key=lambda d: try_int(d.get('seeders', 0)), reverse=True)
```
sharkykh: I don't think there's need to `try_int`, isn't seeders already converted to an int?

sickbeard/providers/yggtorrent.py
```diff
@@ -0,0 +1,158 @@
...
+        if not re.search('logout', response):
+            logger.log('Invalid username or password. Check your settings', logger.WARNING)
+            return False
+
+        return True
+
+    def search(self, search_strings, age=0, ep_obj=None):  # pylint: disable=too-many-locals, too-many-branches
+        results = []
+        if not self.login():
+            return results
```
sharkykh, line 85: The site seems to allow searching without logging in.
If you only need to be logged in to download, [`GenericProvider.download_result()`](https://github.com/SickRage/SickRage/blob/4509c8975c452bc139643e1b4cd73d73d079ffee/sickrage/providers/GenericProvider.py#L85-L87) already has the exact same check.

And I recommend adding tests to this provider, if possible.
I'm not sure the dynamic tests for providers are able to test private providers,
[marked with `self.public = False` - default value](https://github.com/SickRage/SickRage/blob/4509c8975c452bc139643e1b4cd73d73d079ffee/sickrage/providers/GenericProvider.py#L68)

sharkykh (commented):
(Didn't test)

These are general comments, code LGTM otherwise.

sickbeard/providers/yggtorrent.py
adaur: I copy/pasted what's in abnormal provider. If this needs to be fixed, abnormal's as well. But yes, you're probably right

sharkykh:
> It is not possible to sort torrent during search right now, I'll update later when this is possible

There doesn't seem to be an option to limit the search to specific categories.

adaur:
Indeed, and it's a pity. We'll have to do without for now.

sickbeard/providers/yggtorrent.py
sharkykh: Nope, scratch that...
https://github.com/SickRage/SickRage/issues/3915#issuecomment-320514394

If you get **a different download url when logged in**, then **don't** change what you did.
If so, tests are also out of the question.

sickbeard/providers/yggtorrent.py
adaur: I won't change this since it is present on several providers

sickbeard/providers/yggtorrent.py
adaur: If I remove this check, then I cannot download the torrent file because I am not logged in...

sharkykh: That's fine, because the download function makes sure the user is logged in.
My question is, does the **download link** you scrape from the site have a special passcode in it when logged in?
I don't have a user to check this.

adaur: No, it does not

sharkykh: Then my original comment still stands :)

sickbeard/providers/yggtorrent.py
```diff
@@ -0,0 +1,159 @@
...
+            'id': self.username,
+            'pass': self.password,
+        }
+
+        response = self.get_url(self.urls['login'], post_data=login_params, returns='text')
+        if not response:
+            logger.log('Unable to connect to provider', logger.WARNING)
+            return False
+
+        if not re.search('logout', response):
```
sharkykh: No need to use regex, use `if 'logout' not in response:`

sickbeard/providers/yggtorrent.py
```diff
@@ -0,0 +1,159 @@
...
+                            logger.log('Data returned from provider does not contain any torrents', logger.DEBUG)
+                            continue
+
+                        # Skip column headers
+                        for result in torrent_rows[1:]:
+                            cells = result('td')
+                            if len(cells) < 5:
+                                continue
+
+                            try:
```
sharkykh: Sorry, that's not what I meant. This way it will still not log errors (because the exceptions are handled here)
Just remove this `try` and its corresponding `except` in the loop (lines 132-133).
Keep the ones outside the loop.

If it fails because of empty rows or something, tweak your code (e.g. `continue` if the row is empty)

sharkykh (dismissed):

sickbeard/providers/yggtorrent.py
adaur: Done

sharkykh (approved):

PR #3973, opened 2017-08-10
Remove torrents from Transmission on file post processing

fireph:
This adds the functionality to remove torrents from Transmission (but can be added to other clients) where the files in the torrent match the file being post processed. It gets the list of torrents, then checks if the torrent is finished, and then checks each file in the torrent for a match to the post processed file. If there is a match and post processing is successful, the torrent will be removed (but files will remain intact).

- [x] Read contribution [guide](https://github.com/SickRage/SickRage/blob/master/.github/CONTRIBUTING.md)


beschoenen:
It might be a good idea to move the UI part to `config -> search settings -> $torrent_client`. What do you think @OmgImAlexis?

Does this also remove torrents that are broken at 99.% (this happens often when you have move files turned on).

fireph:
I don't think it will remove torrents that are broken like that. The way it is designed now, it will only remove the torrent if it has finished seeding when the post processing happens.

I would have to test out how to detect if a torrent is broken like that in the transmission rpc api (if it is possible).

Gelioks:
Is it going to be pushed into Dev/Master anytime soon? Or the project is almost dead?

arantes555:
I'd love to see this too !

fireph:
I'll try to update this again in a bit with the feature request to remove torrents that are in broken states/at 99.9%

Gelioks:
@DungFu Did you have a chance to look at this? Thanks.

PR #3979, opened 2017-08-11
Season_folders was inverted

### LINKED ISSUE #3978: Season_folders via api is inverted
When using the API to add new shows the default `season_folders` is inverted. If you manually specify `flatten_folders=0` the value is still incorrect.

hjone72:
The value was incorrectly inverted, causing the default via the api to be incorrect.

Fixes #3978

- [X] Read [contribution guide]


miigotu:
flatten_folders is opposite of season_folders, so flatten_folders=0 means use season folders.

hjone72:
That's not how it was acting for me and not how the code represents what
was happening.

Flatten_folders=1 would use season folders. I was also running the latest
update. Since making the changes in my pull request the issue was resolved.

PR #3983, opened 2017-08-12
Add custom season names

beschoenen:
### Use case
This is mostly for anime.
Some anime that air with non-absolute numbering can't be downloaded by SickRage.

I have this problem with the show "New Game."
Season 1 is called 'New Game!', but season 2 is called 'New Game!!' and SickRage parses episodes from New Game!! as Season 1 episodes.

But this problem exists for all non-absolute airing anime with season > 1.

The custom names for the seasons have to manually be entered in `Show -> Edit -> Season exceptions`.


OmgImAlexis:
Don't we already have this by editing the [scene exceptions file](https://github.com/SickRage/sickrage.github.io/blob/master/scene_exceptions/scene_exceptions.json)?

beschoenen:
OMG 😫

OmgImAlexis:
Maybe we should add an option to the UI...

beschoenen:
Is the JSON file saved locally? or is it downloaded from GitHub each time?

OmgImAlexis:
I think it's pulled down once every 24 hours from memory.

beschoenen:
So then how do you propose we make a UI for this?

p0psicles:
It's pulled from gh every 24 hours and stored in the db.

OmgImAlexis:
@beschoenen I meant keep the UI you've added but use it to allow users to override the global list locally.

beschoenen:
Is the `cache.db` not emptied at some point? (assuming that's what you mean)

p0psicles:
afaik sr doesn't empty the scene_exceptions table in cache.db.
I would start with at least visualizing the season exceptions that you have. Don't know if this PR already does that though.

beschoenen:
This PR visualizes the data from the new table, but in about the same way as you'd want to display the data from the season exceptions in the `cache.db`, I'd just have to add support for multiple entries per season.

p0psicles:
It's your project. But I wouldn't introduce a new table. As there is already one which can be easily used for custom exceptions. Just have to flag those.

beschoenen:
I already made the new table, and it was a huge waste of time :P

p0psicles:
Or you move everything to the new table?

beschoenen:
@p0psicles Looks like the `scene_exceptions` table *is* emptied. This method is called when you press save in the editShow page.

https://github.com/SickRage/SickRage/blob/master/sickbeard/scene_exceptions.py#L278

p0psicles:
Yes, but that's easily fixable. Just flag the custom exceptions. And only delete the GitHub exceptions. It's deleted to make sure you always have the same as on gh.

PR #4000, opened 2017-08-20
Scene exceptions ui

beschoenen:
- Edit show now displays exceptions for all seasons.
- You can only edit custom exceptions (not remove the ones from GitHub)
- A user can add exceptions for any season.

screen shot 2017-08-20 at 18 55 08


sharkykh:
Looks nice. Apparently I don't use custom scene exceptions in any of my shows.
I think you should update `displayShow.mako` as well, because when you add a custom Show-level exception, [it only shows your entries here](https://user-images.githubusercontent.com/10238474/29671180-83389d98-88f1-11e7-91b6-c840d639b327.png).

beschoenen:
>because when you add a custom Show-level exception, it only shows your entries here.

What do you mean?

sharkykh:
When you only have the scene exceptions from GitHub, it shows the show-level (season = -1) ones there by default.
If you add custom show-level scene exceptions, it only shows the custom ones, even though they are all considered, as far as I know.

beschoenen:
Oh, they should still only show the ones for -1 (+ the custom for -1). I'll look into it.

beschoenen:
@sharkykh Should be fixed now. The problem was that I only send custom exceptions back the server so I didn't have the non-custom exceptions in memory when building up the cache.

OmgImAlexis:
Not sure if this is just a visual bug. 

screen shot 2017-08-27 at 2 02 29 pm

OmgImAlexis (approved):

beschoenen:
@OmgImAlexis Did you fix the bug with your commit?

OmgImAlexis:
Yep looks like it's working fine. Could you test again just incase I missed something.

beschoenen:
It's working again now, but I think it will fail when a entry has a `:`, `|` or `,`.

gui/slick/js/core.js
```diff
@@ -2871,62 +2869,66 @@ var SICKRAGE = {
...
+                        const season = $(el).data('season');
+                        const exception = $(el).val();
+
+                        if (index === 0) {
+                            return [season, exception].join(':');
+                        }
 
-                $('#exceptions_list option').each(function() {
-                    allExceptions.push($(this).val());
+                        return exception;
```
OmgImAlexis: URL encode this if you want to allow `|` and `:` in the exception.

beschoenen:
Ok, that should be it.

gui/slick/js/core.js
```diff
@@ -2872,15 +2872,15 @@ var SICKRAGE = {
                 const allExceptions = $('#exceptions_list').find('optgroup').get().map(function(group) {
                     return $(group).find('option:enabled').get().map(function(el, index) {
                         const season = $(el).data('season');
-                        const exception = $(el).val();
+                        const exception = encodeURIComponent($(el).val());
```
OmgImAlexis: Don't do it here...

beschoenen: Why not? You want the user's input to be encoded so that the python part doesn't crash on unexpected reserved characters, the `|` from our JS code doesn't matter.

beschoenen: The parser on the python side wouldn't be able to differentiate our `|` from the user's `|`.

OmgImAlexis: I meant you should have done it at the return as vars should always stay unmutated as much as possible. This way if we ever change it later on and want to add something to the value we can and it'll be url encoded with the rest of the exception string.

beschoenen: But then it would also encode the colon at the first loop.

gui/slick/js/core.js
```diff
@@ -2870,16 +2870,14 @@ var SICKRAGE = {
...
+                    const season = $(group).data('season');
+                    const exceptions = $(group).find('option:enabled').get().map(function(option) {
+                        const exception = $(option).val();
 
-                        if (index === 0) {
-                            return [season, exception].join(':');
-                        }
-
-                        return exception;
+                        return encodeURIComponent(exception);
```
beschoenen: @OmgImAlexis Is it better like this?

PR #4034, opened 2017-09-11
Fix connection method + leechers / seeders values

theobarrague:
Fixing connection method + seeders / leechers values ( wrong array index )
@adaur please check, i've tested and it's works for me

### DIFF
--- a/sickbeard/providers/yggtorrent.py
+++ b/sickbeard/providers/yggtorrent.py
@@ -69,11 +69,7 @@ class YggTorrentProvider(TorrentProvider):  # pylint: disable=too-many-instance-
         }
 
         response = self.get_url(self.urls['login'], post_data=login_params, returns='text')
-        if not response:
-            logger.log('Unable to connect to provider', logger.WARNING)
-            return False
-
-        if 'logout' not in response:
+        if response: # Yggtorrent return empty response if user is logged, so ...
             logger.log('Invalid username or password. Check your settings', logger.WARNING)
             return False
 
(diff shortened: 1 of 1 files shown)


adaur:
It works but we should not be patching a bug coming from their side...
ygg
By the way I don't understand how your login code is supposed to work? Is the return of the form a blank page now? (wtf then)

theobarrague:
### Why i changed seeders / leechers array index ?
capture du 2017-09-11 18-20-21
**You used 3,4 as indexes but ( update maybe ? ) is 4 and 5 in fact.**
Please note that the site is not bugged ( from my side ), I am on yggtorrent.com but the language used is fr ( but no importance, is sickrage sending fr language code for web crawling ? )


### Why i changed login code ?
capture du 2017-09-11 18-19-50
*Flow for valid credentials*

Useful lines are blue lines. The first request is for the login page (with images, form, ...), the second request is for posting data. When you post correct credentials, this page returns a refresh header like this: *refresh: 0; url = https: //yggtorrent.com/* and **no html/text data** so the browser redirects to home page and this corresponds to the third blue line.

**So, yes, the return of the form is a blank page.** When you post wrong credentials, the return page is the same as the login page ( check, in the first picture, the size is 103B - so, empty page - and on the second, the size is 13,2Kb - the login page with images, form ... - )


capture du 2017-09-11 18-38-58
*Flow for wrong credentials*
  
  
  
  
  
### Resume
*Sending correct credentials*
Response : *(empty)*
Redirection : Home page

*Sending wrong credentials*
Response : ```<!DOCTYPE html><html>...<p>Wrong credentials, try again</p>...</html>```
Redirection : *None*

adaur:
OK, thanks for the clarification. It changed since I first wrote the code, and the blank page with the refresh tag is so fucked up... Strange I can still see the issue. Thanks anyway.

PR #4043, opened 2017-09-14
Fix search in ncore provider

Casterke:
I tried a lot's of different method, but this is the best, find everything and not slow.
With categories it found nothing.

Fixes #

- Be able to search in the backlog.


nledenyi:
I think replacing the incorrect categories with the correct ones would be a better fix, than searching in all categories (which would include movies, apps, ebooks, etc). I tested it on my end, and it works  fine. Therefore my proposal is to replace these lines:
```
        categories = [
            'xvidser_hun', 'xvidser',
            'dvd_hun', 'dvd',
            'dvd9_hun', 'dvd9',
            'hd_hun', 'hd'
        ]
```
with these:
```
        categories = [
            'xvidser_hun', 'xvidser',
            'dvdser_hun', 'dvdser',
            'hdser_hun', 'hdser'
        ]
```
And can leave everything else untouched

Casterke:
Thanks! I didn't notice the wrong categories. I commited with the correct categories.

nledenyi:
Nice one. Not sure if the "nyit_sorozat_resz=true" is required, for me it works even without that. Might not do any harm either though. 

For me the daily search (RSS) doesn't seem to work with this provider, I have an idea what it could be, need to test then I will create an another PR for it based on yours

nledenyi:
Rather than waiting for this to be merged or causing a conflict with an another PR for this file how about if you do the fix @Casterke in your PR? 

So the issue is with the Daily Search is wrong indention around line 125 (in your commit). The 'result =' should **not** be in the 'if mode != "RSS":' statement.

Also, what I noticed is that you can provide the categories as a comma separated list, 
ie: &kivalasztott_tipus=xvidser_hun,xvidser,dvdser_hun,dvdser,hdser_hun,hdser
So the URL can be a bit shorter:
categories = '&kivalasztott_tipus=' + ','.join([x for x in categories])

Hope eventually all these fixes are going to be merged into the master branch

nledenyi (approved):
Looks good, I tested these changes on my end for the past couple of weeks.

Casterke:
Thank you, I hope this is merged soon to the master.

PR #4080, opened 2017-10-04
Fix scene_season not taking

beschoenen:
This PR is probably wrong, but I'm not sure how else I'm gonna get SickRage to accept the custom season exception set via the UI.


sickbeard/name_parser/parser.py
```diff
@@ -287,7 +287,10 @@ def _parse_string(self, name, skip_scene_detection=False):  # pylint: disable=to
 
                     new_absolute_numbers.append(a)
                     new_episode_numbers.extend(e)
-                    new_season_numbers.append(s)
+                    if bestResult.scene_season is not s:
```
miigotu, line 290: I think this should be done in `get_all_episodes_from_absolute_number`, because the e might be different too

Issue #4266, opened 2018-01-24
Sickrage not finding any episode


Redostrike:
Branch/Commit: 4b69468cc1ea330f598364718068352a54b0eb92
OS: Linux-3.2.40-armv7l-with-glibc2.4
What you did: Manual search finds nothing.
What happened: My synology had a power failure the other day so needed a reboot. Since then sickrage cant find any more episodes. Had a old commit (updates where disabled) so i gone ahead an updated as i thought this might resolve the issue. Now auto search and manual search aren't finding any episode of any tv show.
What you expected: Find episodes via auto search & manual search.
Logs:
```
2018-01-24 10:02:17 INFO SEARCHQUEUE-MANUAL-279121 :: Unable to find a download for: [The Flash (2014) - S04E11 - The Elongated Knight Rises]
2018-01-24 10:02:17 DEBUG SEARCHQUEUE-MANUAL-279121 :: formatting pattern: %SN - S%0SE%0E - %EN -> The Flash (2014) - S04E11 - The Elongated Knight Rises
2018-01-24 10:02:17 DEBUG SEARCHQUEUE-MANUAL-279121 :: formatting pattern: %SN - S%0SE%0E - %EN -> The Flash (2014) - S04E11 - The Elongated Knight Rises
2018-01-24 10:02:17 DEBUG SEARCHQUEUE-MANUAL-279121 :: [Nyaa] :: The Flash (2014) is not an anime, skipping
2018-01-24 10:02:17 INFO SEARCHQUEUE-MANUAL-279121 :: [TorrentDay] :: Performing season pack search for The Flash (2014)
2018-01-24 10:02:17 DEBUG SEARCHQUEUE-MANUAL-279121 :: [TorrentDay] :: Fallback season pack search initiate
2018-01-24 10:02:17 DEBUG SEARCHQUEUE-MANUAL-279121 :: [TorrentDay] :: No data returned from provider
2018-01-24 10:02:17 DEBUG SEARCHQUEUE-MANUAL-279121 :: [TorrentDay] :: Requested a json response but response was not json, check the url: https://www.torrentday.com/V3/API/API.php
2018-01-24 10:02:17 DEBUG SEARCHQUEUE-MANUAL-279121 :: [TorrentDay] :: With post data: cata=yes&search=The+Flash+%282014%29+S04E11&c26=1&jxt=8&jxw=b&c24=1&c34=1&c2=1&c7=1
[...]

PR #4298, opened 2018-02-04
Update torrentday.py

kayl669:
Fixes #
https://github.com/SickRage/SickRage/issues/4266
- Fix torrentday search

### DIFF
--- a/sickbeard/providers/torrentday.py
+++ b/sickbeard/providers/torrentday.py
@@ -23,7 +23,7 @@ from __future__ import unicode_literals
 import re
 
 import validators
-from requests.compat import urljoin
+from requests.compat import urljoin, quote_plus
 from requests.utils import dict_from_cookiejar
 
 from sickbeard import logger, tvcache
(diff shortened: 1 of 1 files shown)


AubreyKilian:
I did a manual search for an episode, and the new code did find the episode.  Great work.

AubreyKilian:
Just a question, I see the code does a category lookup first, then when it doesn't find the episode, it does an actual search on the series name + season+episode.  Why do the category lookup first?

AubreyKilian:
Something's not right with the search for series names with brackets in the name.
For example:
GET URL: https://www.torrentday.com/t.json?24=1&
26=1&34=1&q=MacGyver%2B%25282016%2529%2BS02E15&2=1&7=1

The series name is "MacGyver (2016)"
The above URL has %2B%2528 which urldecodes to "+%28", there seems to be a double encoding happening on the series name?
(I'm unsure if this problem is related to this fix though...  If not, please ignore, I'll create a new issue then...)

PR #4410, opened 2018-03-19
Update pynma.py

p00chie:
New API Server, otherwise API calls per hour are very limited.

Fixes #
Could not send notification to NotifyMyAndroid: Your IP exceeded the maximum number of API calls per hour allowed.


p00chie:
More info:
https://www.domoticz.com/forum/viewtopic.php?t=21124#p171497

PR #4439, opened 2018-03-24
Added Amazon as search tag and fixed itunes tag

p00chie:
Fixes

added the amazon tag for releases
changed itunes tag to itunes hd/uhd
changed common.py so avc tag is not needed (avc = re.compile(r'([xh].?26[45])', re.I))
The main reason is that some releases were found with unknown quality.
In common.py you want to know if a release is WEBDL. Itunes/Amazon/Netflix can only be a WEBDL release, so there is no point to look for the codec. In some cases the AVC tag was not found and the result is unknown quality.

On the other hand you could add 'AVC' to the avc tag but in the end it stays the same. The result will be webdl


miigotu:
I agree with most of this PR except the exclusion of the need for the codec. Scene release names require a codec to be a proper release name, and SickRage only trusts real scene releases because anyone can put a random tag on a file and upload it, but real release groups always understand the rules for scene names. This is one of the small nuances we use to be accurate and limit the chances of getting fakes, and it works fairly well to do that.

miigotu:
This is also why your tests are failing btw. I would also like you ask you to add some tests for these new changes. We already test some netflix and itunes release names for these methods.

miigotu:
I'll take your commits and add them to my branch. I have to make some changes to this and add some tests. Thanks for the help.

Also, could you pass me any release names to my email I should add to the tests?  miigotu at gmail, thanks

p00chie:
I think you are correct. A codec is needed otherwise there could be some errors if netfilx/amazon or something like that is in the name. I think AVC is just missing in the AVC tag (lol).

This will fail with the current release --> quality UNKNOWN
Black.Mirror.S02E01.Wiedergaenger.German.DD20.Dubbed.DL.1080p.iTunesHD.**AVC**-TVS

I will sent you an email with more release names.

miigotu:
I added AVC to the tags, but if you notice they are all dubs, which generally are not tagged correctly because they are not usually top tier release groups that follow scene rules. They tend to go by their own rules, and when you extrapolate that to dozens of languages there is no way we really can support all of the frivolous custom naming rules from fringe release groups. We have scene rules for a reason I just wish they would follow them =P

PR #4443, opened 2018-03-25
Better quaity detection for AMZN and Amazon, Itunes, and Netflix releases

miigotu:
Fixes #4439 

- Adds better quaity detection for AMZN and Amazon, Itunes, and Netflix
- Sets lowest quality for those services to HDWEBDL (This might need to change)

- [x] PR is basted on the DEVELOP branch

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -338,7 +338,7 @@ class Quality(object):
             if ep.avc and ep.bluray:
                 result = Quality.UHD_4K_BLURAY if not full_res else Quality.UHD_8K_BLURAY
             # WEB-DL
-            elif (ep.avc and ep.itunes) or ep.web:
+            elif ep.itunes or ep.amazon or ep.netflix or ep.web:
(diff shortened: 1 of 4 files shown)


p00chie:
Wouldn't it be better to introduce a Studio tag? Instead for checking for every single studio it should be a common studio tag. If a new one will emerge in the future you just have to add it there in the tags.py. It would be the only spot to maintain everything.
Also the common.py would be easier to handle with the studio tag because you do not have to check for every single one there too...

miigotu:
Id rather not, because technically it isnt correct how it is right now anyways. There is no `AmazonUHD` for example, Netflix is usually just tagged `-NF` and not all netflix/amazon/itunes are HD because while it may have been downloaded from those services some release groups still reencode it making it a webrip rather than a web-dl and SR treats bot webrip and webdl the same. Per scene rules `Itunes` is supposed to be SD or 720p, `ItunesHD` is supposed to be 1080p, and `ItunesUHD` is supposed to be 4k or 8k, so linking them together as far as they are now is actually not even fine grained enough.

PR #4597, opened 2018-04-11
Don't use torrent cache services as they don't work

felixbrucker:
Currently when using blackhole for torrents, magnet links are "converted" to torrent files using webservices for torrent caching. But this doesn't work.

I'm not very familiar with python and i'd like some pointers on how to achieve the following:

make a checkbox in blackhole config to enable the changes i did (the optional flag `force_magnet`) so magnets are stored in a text file with .magnet extension.

### DIFF
--- a/sickrage/providers/GenericProvider.py
+++ b/sickrage/providers/GenericProvider.py
@@ -87,7 +87,7 @@ class GenericProvider(object):  # pylint: disable=too-many-instance-attributes
         if not self.login():
             return False
 
-        urls, filename = self._make_url(result, true)
+        urls, filename = self._make_url(result, True)
 
         for url in urls:
             if 'NO_DOWNLOAD_NAME' in url:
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
What do you mean by "doesn't work"? This should have been opened as an issue before opening a PR.

felixbrucker:
Well all the enabled sites i checked didn't work, either 404 or similar errors.

I wrote a small app which will handle torrent retrieval via various other services via blackhole but it seems sickrage doesn't have the .torrent files directly, only magnets, and needs to convert them first using some online services.

As my app can work with magnets as well i added this functionality to export magnet links directly instead of converting them to .torrents

This is the list of services currently used: https://github.com/SickRage/SickRage/blob/master/sickrage/providers/GenericProvider.py#L54-L61

OmgImAlexis:
Okay, again this is why you should open an issue before the PR.

Please open an issue and we'll discuss.

OmgImAlexis:
This also should have been open against develop not master.

PR #4668, opened 2018-05-10
YggTorrent : Fix for session expiration (very short)

Arakmar:
The session duration is very short on YggTorrent.
So we need to login again after the first search if we detect an expired session.

Tested for two weeks and it works fine.

- [X] Read [contribution guide](https://github.com/SickRage/SickRage/blob/master/.github/CONTRIBUTING.md)


miigotu:
This will skip the search, since the login happens after the search and data will have no results when parsed.

Arakmar:
With this provider, the search doesn't require the user to be logged in. In fact, it's only used when we retrieve the torrent file. So the data var is still correct after (and before) the call to the login method.
The advantage to do that is to be able to detect if the session has expired or not without another call. With the previous code, we needed to clear manually the session cookie every days.

I tested this code for two weeks without any issues.

PR #4829, opened 2018-07-19
Fix for issue #4721

### LINKED ISSUE #4721: Search providers page errors with 'TypeError: tuple indices must be integers, not unicode'
Branch/Commit: v2018.05.11-1-1-g5a9a634
OS: CentOS 7.5
Browser: Firefox 61
What you did: Clicked 'Search Providers'
What happened: 
```
File /opt/SickRage/gui/slick/views/config_providers.mako:183, in render_pages:
${('', 'checked="checked"')[curNewznabProvider.enable_daily]}
TypeError: tuple indices must be integers, not unicode
```
What you expected: To see the search providers and their settings
Logs: nothing in the log file relating to this at all even with debug logging enabled, the only error is via the browser.

I would have expected some better output about this error, e.g. if it's something in the database causing the issue, where in the database it is.

```sql
sqlite> select * from db_version;
44|0
```

The config file is already ascii with no unicode characters:

```
# iconv --from-code UTF-8 --to-code US-ASCII -c config.ini > config.ascii
# diff config.ini config.ascii
```

ravensorb:
Fixes #4721

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -180,7 +180,7 @@
                                         <div class="col-lg-9 col-md-8 col-sm-7 col-xs-12 component-desc">
                                             <input type="checkbox" name="${curNewznabProvider.get_id("_enable_daily")}"
                                                    id="${curNewznabProvider.get_id("_enable_daily")}"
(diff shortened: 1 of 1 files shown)


miigotu:
Ill do this a little different. Thanks though.

PR #4914, opened 2018-08-16
change yggtorrent url

skillfr:
Or adapter for the code that it recovers the target https://yggtorrent.is

```
import requests
response = requests.get('https://yggtorrent.is')
print(response.url)
```

### DIFF
--- a/sickbeard/providers/yggtorrent.py
+++ b/sickbeard/providers/yggtorrent.py
@@ -48,7 +48,7 @@ class YggTorrentProvider(TorrentProvider):  # pylint: disable=too-many-instance-
         self.minleech = None
 
         # URLs
-        self.url = 'https://ww3.yggtorrent.is/'
+        self.url = 'https://ww4.yggtorrent.is/'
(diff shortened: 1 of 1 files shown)


JoshuaArus (approved):
Worked for me

Dioperso (approved):
work for me

ArthurGarnier:
Changed again to www...

djoole:
Would be better to add a dialog in the torrent settings to let the user choose the url (which is changing to often to hardcode it).
I'll try to do that if I can find the time

ArthurGarnier (changes requested):
Should be "www" now

ArthurGarnier:
"requests" is also a good idea as main domain redirect to the right subdomain.

JoshuaArus:
I agree on the "requests" feature. An input in the providers configuration to specify URL would not be as powerful because user need to notice the domain change, check the new one and update the configuration.

requests.get should work properly at every subdomain change.

PR #5150, opened 2018-10-31
Update yggtorrent.py

ThorlakKilden:
New URL


miigotu:
Why is this necessary? The url changing is automatically corrected after the first search.

Issue #5118, opened 2018-10-24
rTorrent: Unable to send Torrent


JohnMaguire:
### Before submitting your issue:

Enable debug logging in SickChill settings, reproduce the error (be sure to disable after the bug is fixed)

Branch/Commit:
```
Branch: master
Commit: 6e3a1dccd94857fc381c9a246b903904eca424e1
Database Version: 44.0
```
OS: `Linux-4.14.35-1-lts-x86_64-with` (in Docker)
Browser: Google Chrome
What you did: Forced a backlog search
What happened: Torrents found, but failed to send to rTorrent
What you expected: Torrents downloading in rTorrent
Logs:
```
2018-10-24 10:07:39 WARNING SEARCHQUEUE-BACKLOG-194031 :: rTorrent: Unable to send Torrent
2018-10-24 10:07:39 WARNING SEARCHQUEUE-BACKLOG-194031 :: Error while sending torrent:
2018-10-24 10:07:39 DEBUG SEARCHQUEUE-BACKLOG-194031 :: Result Hash is XXX
2018-10-24 10:07:39 DEBUG SEARCHQUEUE-BACKLOG-194031 :: Calling rTorrent Client
2018-10-24 10:07:39 DEBUG SEARCHQUEUE-BACKLOG-194031 :: GET URL: https://XXX [Status: 200]
```

(URL and hash removed.)

From rTorrent xmlrpc.log:

```
---
<?xml version='2.0'?>
<methodCall>
<methodName>load_raw</methodName>
<params>
<param>
<value><base64>
(snip)
</base64></value>
</param>
</params>
</methodCall>

---
Status: 200 OK
Content-Type: text/xml
Content-Length: 309

<?xml version="1.0" encoding="UTF-8"?>
<methodResponse>
<fault>
<value><struct>
<member><name>faultCode</name>
<value><i4>-506</i4></value></member>
<member><name>faultString</name>
<value><string>Method 'load_raw' not defined</string></value></member>
</struct></value>
</fault>
</methodResponse>

---
```

(Removed base64'd value)

JohnMaguire:
FWIW, this looks similar to this issue: https://github.com/pymedusa/Medusa/issues/3223

JohnMaguire:
It looks like `load_raw` was renamed to `load.raw`, along with a bunch of other changes... https://github.com/pymedusa/Medusa/pull/3333/files#diff-33878acf9434aa074045d228eade8661

mr-orange:
[...]

Issue #4865, opened 2018-08-11
Call rtorrent failed but config test is OK


Hydrog3n:
### Before submitting your issue:

Enable debug logging in SickRage settings, reproduce the error (be sure to disable after the bug is fixed)

Branch: master
Commit: 0e727649748ec38c5bba69f3f97c9e7b794d56df
Version: v2018.07.21-1
Database Version: 44.0
OS: Linux
Browser: Chrome
What you did: Try to find episode
What happened: find episode but not send to rtorrent
What you expected: sent to rtorrent for download
Logs:
```
2018-08-11 13:20:52 WARNING SEARCHQUEUE-MANUAL-300472 :: rTorrent: Unable to send Torrent
2018-08-11 13:20:52 WARNING SEARCHQUEUE-MANUAL-300472 :: Error while sending torrent:
2018-08-11 13:20:52 DEBUG SEARCHQUEUE-MANUAL-300472 :: Result Hash is 35a81c06a1f9686e11d7d33e889dd5aa5dbd49af
2018-08-11 13:20:52 DEBUG SEARCHQUEUE-MANUAL-300472 :: Calling rTorrent Client
```

xataz:
Additional information : 
Work with rtorrent 0.9.6, but no with 0.9.7

Try with scgi : scgi://localhost:5000
with rutorrent : http://localhost:8080/torrent/plugins/httprpc/action.php
And scgi_pass uri (nginx) : http://localhost:8080/torrent/RPC

For these tests, config test is Ok

bolhaskutya:
Same here.

Issue #4793, opened 2018-06-18
Failing to send magnets to rTorrent


deusxanime:
Branch/Commit: Branch: master
                           Commit: 93923792e8908427344d6e01de968656cf096626
OS: [Docker/container on unRAID](https://lime-technology.com/forums/topic/46127-support-binhex-rtorrentvpn/)
Browser: Chrome latest
What you did: Manual/Automatic search on a show release
What happened: Failure message, doesn't show up in rTorrent
What you expected: Imported into rTorrent as before
Logs:
```
2018-06-15 23:25:05 WARNING SEARCHQUEUE-MANUAL-284210 :: rTorrent: Unable to send Torrent
2018-06-15 23:25:05 WARNING SEARCHQUEUE-MANUAL-284210 :: Error while sending torrent:
2018-06-15 23:25:05 INFO SEARCHQUEUE-MANUAL-284210 :: Downloading Colony.S03E07.720p.AMZN.WEBRip.DDP5.1.x264-KiNGS[rartv] from Rarbg
2018-06-15 23:25:05 INFO SEARCHQUEUE-MANUAL-284210 :: [Rarbg] :: Release: Colony.S03E07.iNTERNAL.720p.WEB.x264-BAMBOOZLE[rartv] doesn't contain any of AMZN, ignoring it
2018-06-15 23:25:05 INFO SEARCHQUEUE-MANUAL-284210 :: [Rarbg] :: Quality of Colony.S03E07.iNTERNAL.720p.WEB.x264-BAMBOOZLE[rartv] is 720p WEB-DL
2018-06-15 23:25:05 INFO SEARCHQUEUE-MANUAL-284210 :: [Rarbg] :: Quality of Colony.S03E07.720p.AMZN.WEBRip.DDP5.1.x264-KiNGS[rartv] is 720p WEB-DL
2018-06-15 23:25:05 INFO SEARCHQUEUE-MANUAL-284210 :: [Rarbg] :: Release: Colony.S03E07.720p.WEBRip.x264-TBS[rartv] doesn't contain any of AMZN, ignoring it
2018-06-15 23:25:05 INFO SEARCHQUEUE-MANUAL-284210 :: [Rarbg] :: Quality of Colony.S03E07.720p.WEBRip.x264-TBS[rartv] is 720p WEB-DL
2018-06-15 23:25:02 INFO SEARCHQUEUE-MANUAL-284210 :: [Rarbg] :: Performing episode search for Colony
[...]

PR #5216, opened 2018-11-29
rTorrent update (#5118, #4865, #4793)

defaultsecurity:
**Goals:**
- Fix sending to rTorrent
- Make SickChill compatible with rTorrent 0.9

**Tested and Working:**
- Sending torrents via httprpc url: http://0.0.0.0:80/plugins/httprpc/action.php
- Sending torrents as torrent files
- Telling rTorrent to save torrent in a custom folder
- Telling rTorrent to add custom label to torrent
- Telling rTorrent to start download or not

**Not tested:**
- Sending torrent via scgi
- Authentication: Sending torrents to password protected rTorrent/ruTorrent
- Verify SSL certificates for HTTPS requests
- Sending torrents as magnet links

**Notes:**
- Based on develop branch
- Can be tested: (https://github.com/bolhaskutya/SickChill/tree/5118-rtorrent)
       - Git: https://github.com/bolhaskutya/SickChill.git
       - Branch: 5118-rtorrent

**Related Issues:**
- https://github.com/SickChill/SickChill/issues/5118
- https://github.com/SickChill/SickChill/issues/4865
- https://github.com/SickChill/SickChill/issues/4793

This is my first pull request. For example, I'm not sure what is: "continuous-integration/travis-ci/pr — The Travis CI build failed"


miigotu:
First and foremost, thanks for your contribution.

Does this break rtorrent for anyone that may be stuck on an earlier version? The rtorrent docker image I use for testing worked with the before this PR, which tells me that the apt repository for some distros do not provide the newer release of rtorrent. If it does break something, we need to support both the old and the new version like we do with other clients.

Also, please remove those comments at the top of every file. It is one thing to credit the author, but keep it to the format we have set.
We do not need permission to share code, but we do have to give credit to the author. We don't put links to competing forks in our PR's
These comments are what is breaking the build, because `# coding=utf-8` has to be the first line on the files.

defaultsecurity:
@miigotu 
Unfortunately, I can't test it with earlier versions, I could barely get a single ruTorrent release (docker, linuxserver, release 173) working on my synology, sadly.

I changed the header of the files to include the standard SickChill header, but I think we need to credit the Medusa team properly as a whole and include a link to their repo as requested by @medariox here: https://github.com/pymedusa/Medusa/pull/3333#issuecomment-443129582

SickChill isn't my code, Medusa isn't my code. I just want to fix this issue and give credit where it is due. Both teams have done an awesome job, I'm just glad to be here :) Please advise.

miigotu:
@bolhaskutya the lib folder is supposed to stay as it comes from upstream. If the only place this modified librtorrent comes from is medusa, we need to break that out into a separate repo. The licenses in the lib folder stay as how they came from the source.

Putting a link to medusa is unacceptable, and seeing as we have shared code before both ways they know this.

WebSpider:
Also, this breaks my current rutorrent/rtorrent production install, so it's at least not always backwards compatible.

defaultsecurity:
@miigotu This is just sad. Create a new repo? Not credit Medusa with a few links in a few files? I can't fix the rTorrent bug with such conditions. I came here with the best intentions. I thought this repo was about the users, not the authors. Sorry, but I will take no part in this. I'm switching to Medusa. Best wishes fixing the rTorrent bug.

@WebSpider Thanks for the input. I believe the backwards compatibility can be fixed with version checking. The method names are different in the old and new rTorrent, thus two sets of method definitions would be needed.

miigotu:
@bolhaskutya crediting the author is one thing. Linking to a competing fork is something totally different. medariox knows this, as we have shared code back and forth between forks before. In fact, he didnt even write the whole file, they just modified the one that we already had.

miigotu:
And yes, when you modify a library that isnt part of the source code, you contribute back to the upstream librtorrent.

defaultsecurity:
I understand. I guess you have a history. It would be a pity to leave this bug unfixed. Do as you see fit.

PR #5246, opened 2018-12-16
Torrent9 custom url

### LINKED ISSUE #5190: Torrent9 new address
I don t know what happens but i think Torrent9 change again adresse, but i can t dowload on it ?

Did domeone find a way ?

miigotu:
Add custom url ability and change default to torrent9.uno
Fixes #5190


Kubilai12:
I  think the good one is https://www.torrents9.pw cause no strange VPN ads

PR #5257, opened 2018-12-24
Update stale.yml

### LINKED ISSUE #5256: Expand ignores for stale
Stale bot needs to ignore the enhancement and feature request labels as well

miigotu:
Fixes #5256

@WebSpider I'm not sure all of them need to be exempted, because if the bot tags it stale and nobody responds within 7 days it must not be a very important request I think?


miigotu:
Approved

PR #5319, opened 2019-02-07
Add Irib TV1 and Qatar TV networks icons

Farossco:


Farossco:
Is there a way to ignore AppVeyor since I just uploaded pictures ?

miigotu (approved):

PR #5424, opened 2019-04-13
feat: Slack Icon options

chronicrebellion:
**Slack Config**
slack config

**Slack Test Message**
test message

- Default SC png as the icon for slack messages
- Allows custom emoji icons for slack notifications
Reference: https://api.slack.com/custom-integrations/incoming-webhooks

### DIFF
--- a/gui/slick/views/config_notifications.mako
+++ b/gui/slick/views/config_notifications.mako
@@ -3091,15 +3091,6 @@
                                 </div>
                             </div>
 
-                            <div class="field-pair row">
-                                <div class="col-lg-3 col-md-4 col-sm-5 col-xs-12">
-                                    <label for="slack_icon_url" class="component-title">${_('Slack Icon Url')}</label>
-                                </div>
-                                <div class="col-lg-9 col-md-8 col-sm-7 col-xs-12 component-desc">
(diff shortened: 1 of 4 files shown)


chronicrebellion:
@miigotu just wondering if there were any tests I should update or if this is satisfactory?

miigotu:
Why not just send the SC icon like it's supposed to?

chronicrebellion:
I was considering using: https://github.com/SickChill/SickChill/blob/master/gui/slick/images/sickchill-sc.png

I was not sure if this was going to change or not, I did notice that in the sickchill subreddit that there was talk about getting a new logo.

I thought people might want to have a custom icon :)

If we wanted to just allow only custom icon with emoji, I can just set the url to the SC png so it would be the default icon. If no emoji is set, then the url icon will show up.

If you think just using SC is the way to go I can totally do that as well... just let me know!

miigotu:
Yes I think url icon from that url as the default is best, and probably do not need the emoji but you already added it so let that be an override.

chronicrebellion:
> Yes I think url icon from that url as the default is best, and probably do not need the emoji but you already added it so let that be an override.

@miigotu I have updated the PR to default to the SC icon and allowing users to provide a custom emoji icon (screenshots updated above)

PR #5449, opened 2019-05-01
Fix Delete RAR contents issue

mystycs:
Fixes #

- Fixes rar deletion. The old code would look at the subpath where the rar would extract to. Had to change the logic to look at the parent directory. Also many torrents would not return the proper result for some odd reason. This checks if a rar file exists in the folder we want to Delete. Tested very thoroughly and works without issues.

- Fixes regex for detecting rars. Most of the multipart rars filetypes uploaded these days on torrents are formatted as so (file.rar, file.r01, file.r02, file.r03 etc..) Old versions of winrar use the part1.rar, part2.rar style.

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -35,16 +35,8 @@ import requests
 from configobj import ConfigObj
 from tornado.locale import load_gettext_translations
 
-try:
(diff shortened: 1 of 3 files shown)


mystycs:
Meant to compare to merge into develop branch.

PR #5450, opened 2019-05-01
Delete rar contents issue fix

mystycs:
Fixes #

- Fixes rar deletion. The old code would look at the subpath where the rar would extract to. Had to change the logic to look at the parent directory. Also many torrents would not return the proper result for some odd reason. This checks if a rar file exists in the folder we want to Delete. Tested very thoroughly and works without issues.

- Fixes regex for detecting rars. Most of the multipart rars filetypes uploaded these days on torrents are formatted as so (file.rar, file.r01, file.r02, file.r03 etc..) Old versions of winrar use the part1.rar, part2.rar style.

This fix solves multiple open issues (https://github.com/SickChill/SickChill/issues/4076, https://github.com/SickChill/SickChill/issues/4205, https://github.com/SickChill/SickChill/issues/3361)


sickbeard/processTV.py
```diff
@@ -239,8 +239,22 @@ def process_dir(process_path, release_name=None, process_method=None, force=Fals
...
+            try:
+                #The directory_from_rar includes the subfolder that is extracted. We want to cut this from the path string since we want to
+                #delete the rar files in the parent folder.
+                rar_path = str(directory_from_rar).split(os.path.basename(directory_from_rar))[0]
+                result.output += log_helper('Deleting rar files in path: ' + str(rar_path), logger.DEBUG)
+                #Iterate through parent folder and if_rar_file is true add it to the list
+                rar_files = [x for x in os.listdir(rar_path) if helpers.is_rar_file(ek(os.path.join, rar_path, x))]
+                this_rar = [rar_file for rar_file in rar_files if os.path.basename(directory_from_rar) == rar_file.rpartition('.')[0]]
+                result.output += log_helper('Deleting the files: ' + str(this_rar), logger.DEBUG)
+                if any('.rar' in x for x in rar_files):
```
miigotu, line 251:
```python
if any(helpers.is_rar_file(x) for x in rar_files):
```

But I dont understand why you would change `result.result` here. `result.result` is whether processing the extracted file worked, and if it did not we do not want to extract the rar again on next post process. The files are supposed to remain in the extract directory until the issue is handled by the user when there is a failure, rather than eat up CPU cycles and disk bandwidth extracting every 40 minutes. It becomes a huge issue if someone has dozens of unprocessed files that are failing.

miigotu:
Several problems with this. It isnt unicode safe for all localizations. You cannot use `str` calls or call `os`/`sys` calls related to file system or paths without using `ek`

sickbeard/processTV.py
mystycs: For some odd reason when my file is extract even if they are successful it still shows false? Dunno this was just a hack I attempted. I'll check again

MorganLester:
Been having this issue since upgrading to 'SickChill' from rage awhile ago. Extracts fine simply leaves the rar files in root folder of xyz and creates sub dir of unrared content.

sickbeard/processTV.py
miigotu: most likely due to some things not using ek there, like basename, etc

miigotu:
@mystycs Why did you decide to close this instead of finishing it?

PR #5495, opened 2019-05-21
Deluged: add download and completed dirs to config

### LINKED ISSUE #5476: Deluge: download path is not passed
Branch/Commit: 175df89cabf295358b8caa2d96dafadb2347622d 
Deluge: 1.3.15
OS: Synology 

I've set a directory in "Downloaded files location" in Search Settings -> Torrent Search. Torrent gets passed to Deluge without the configured path. The Deluge default download directory is used instead. Expectation is for SickChill to pass the directory specified in  "Downloaded files location" to Deluge, overriding the default directory. 

Nothing relevant in the debug logs about the "Downloaded files location" directory when a torrent is sent from SickChill to Deluge. 

Looks like this was previously filed in #3082 but not fixed. The workaround mentioned does not work for my particular needs. I have SickChill set to download to a particular directory and then a label in Deluge applied which will move it from that directory to a different one when the download is complete. The Deluge default directory is on a different volume, so hardlinking between the two is not possible.

nopoz:
Fixes #5476

- Adds "download directory" and "completed directory" options to "Deluge (via Daemon)". This allows the user to configure the initial download location rather than using the default one set by Deluge. The previous behavior prior to this commit was to only set the "completed directory". Does not effect "Deluge (via WebUI)".


sickbeard/clients/deluged_client.py
```diff
@@ -102,8 +114,8 @@ def _set_torrent_priority(self, result):
         return True
 
     def _set_torrent_path(self, result):
```
miigotu, line 116: This is called right after adding the torrent, not after the torrent is complete. Is this correct to change the path right after adding? I would think it would start downloading at TORRENT_DELUGE_DOWNLOAD_DIR and you would immediately move it to TORRENT_DELUGE_COMPLETE_DIR without waiting for it to finish,

gui/slick/views/config_search.mako
```diff
@@ -1070,6 +1070,56 @@
...
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <label for="torrent_deluge_complete_dir_deluge">${_('where Deluge will move files after downloading is complete (blank for client default)')}</label>
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <label id="path_synology"><b>${_('note')}:</b> ${_('the destination has to be a shared folder for Synology DS</span>')}</label>
```
miigotu: please change the label id to torrent_deluge_complete_dir_deluge (I think)

gui/slick/views/config_search.mako
```diff
@@ -1070,6 +1070,56 @@
...
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <label for="torrent_download_dir_deluge">${_('where Deluge will save downloaded files (blank for client default)')}</label>
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <label id="path_synology"><b>${_('note')}:</b> ${_('the destination has to be a shared folder for Synology DS</span>')}</label>
```
miigotu: Please change the id to torrent_download_dir_deluge (I think)

gui/slick/views/config_search.mako
```diff
@@ -1070,6 +1070,56 @@
                                         </div>
                                     </div>
 
+                                    <div class="field-pair row" id="torrent_deluge_download_dir_option">
```
miigotu: torrent_deluge_download_dir_option and torrent_deluge_complete_dir_option can be in the same div, and hidden/shown as one unit so there is less js

miigotu (changes requested):

sickbeard/clients/deluged_client.py
nopoz: Both options can be set right away when the call is made to Deluge. The TORRENT_DELUGE_COMPLETE_DIR is only used by Deluge when the torrent finishes downloading. 

In the below screenshot from Deluge:
 "Download Location" = TORRENT_DELUGE_DOWNLOAD_DIR
"Move Completed Location" = TORRENT_DELUGE_COMPLETE_DIR

Capture

gui/slick/views/config_search.mako
nopoz: @miigotu Should it be "torrent_complete_dir_deluge" instead to match "torrent_download_dir_deluge"?

gui/slick/views/config_search.mako
nopoz: changed in 79f9e0f7b170ee040badc6189a556838d26750af

gui/slick/views/config_search.mako
nopoz: changed in 79f9e0f7b170ee040badc6189a556838d26750af

gui/slick/views/config_search.mako
nopoz: changed in 79f9e0f7b170ee040badc6189a556838d26750af

gui/slick/views/config_search.mako
```diff
@@ -1070,6 +1070,53 @@
...
+                                                           value="${sickbeard.TORRENT_DELUGE_COMPLETE_DIR}"
+                                                           class="form-control input-sm input350" autocapitalize="off"/>
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <label for="torrent_complete_dir_deluge">${_('where Deluge will move files after downloading is complete (blank for client default)')}</label>
+                                                </div>
+                                            </div>
+                                            <div class="row">
```
miigotu: Just delete this whole div. Its a note under the path for the synology path you copied this from that says the folder has to be a shared folder on synology, which isnt true for deluged

gui/slick/views/config_search.mako
```diff
@@ -1070,6 +1070,53 @@
...
+                                                           value="${sickbeard.TORRENT_DELUGE_DOWNLOAD_DIR}"
+                                                           class="form-control input-sm input350" autocapitalize="off"/>
+                                                </div>
+                                            </div>
+                                            <div class="row">
+                                                <div class="col-md-12">
+                                                    <label for="torrent_download_dir_deluge">${_('where Deluge will save downloaded files (blank for client default)')}</label>
+                                                </div>
+                                            </div>
+                                            <div class="row">
```
miigotu: Just delete this whole row div. Its a note under the path for the synology path you copied this from that says the folder has to be a shared folder on synology, which isnt true for deluged

gui/slick/views/config_search.mako
miigotu: Just delete this whole row div. Its a note under the path for the synology path you copied this from that says the folder has to be a shared folder on synology, which isnt true for deluged

sickbeard/clients/deluged_client.py
miigotu: I understand how you expect it to work, its just not easy to see in the code that it works that way because of the introspection and a lack of deluged rpc docs.

Im just worried that set_torrent_path will just change the download location rather than setting the completed location like intended, or another criss-criss because I cannot verify with docs.

Either way, worst case scenario is it will just need fixed if an issue pops up, Ill just have to trust that you made the calls and checked with differing values to see that it works because Im not interested in installing deluge to check for purpose of review :)

miigotu (changes requested):
I deleted those divs, all good to go!

miigotu:
Thanks

PR #5534, opened 2019-06-12
Backend to allow listing available downloads for a particular episode…

miigotu:
… and force snatching it.


miigotu:
@OmgImAlexis any chance I could entice you to swap out grunt, bower, npm and go entirely into webpack (or Parcel or other?) and yarn only? I don't want to lose my grunt tasks but i feel like 3 different ways of doing things is muddy af  =P

PR #5564, opened 2019-07-01
Add text/plain part as multipart alternative to email notifications.

sickchilluser:
- email notifications are constructed as multipart/alternative messages, but nevertheless they only contain an html part. This PR adds a text/plain part as alternative. This will allow the notification to be read as non-html mail, while the html version (as the last multipart alternative) will still be the preferred version. So this won't change anything for users that allow html mails, but it will allow the notifications to be read for users blocking html mails in favour of text/plain mails.


WebSpider:
I'm merging develop in since it's been a while, and checks against this branch have not ran due to a bug

WebSpider (approved):
Looking good, thanks for your contribution!

PR #5601, opened 2019-08-14
Update cfscrape and dependencies

ghost:
- Update cfscrape to v2.0.8
- Updates dependencies

Notes:
The `lib/README.md` was not updated. It doesn't look like it's been updated in a long time.
The pip command that I used was different from what's mentioned in `lib/README.md` as to match the example set by recently merged PRs.

***These changes to cfscrape have reportedly helped users avoid the reCAPTCHA.***


WebSpider (commented):
Reviewing this I wonder why certifi has been updated and requests requirement has been moved up, can't seem to find this in upstream commit logs.

Do you have any idea?

ghost:
:thinking: Nice catch. I performed a manual cleanup prior to running pip which caused the requests requirement to be updated. The same pip command causes certifi to be updated regardless of said cleanup.

> rm -r lib/{certifi,chardet,idna,urllib3,cfscrape}* lib/requests!(_oauthlib)
> pip install --target=lib -U --no-compile --no-cache-dir cfscrape

I didn't think a post install cleanup would have been as straightforward among the distinfo, egginfo, and other remnants. Why exactly Is the distinfo being included, for license information? The pip command in `lib/README.md` doesn't generate it when using the latest version of pip.

I don't see that it was necessary to update either of those though. *I can fix this up, just lmk.*

Edit: I removed `lib/bin` after the install.

WebSpider:
I actually don't know why we have the lib/ dir. I'm assuming it's a windows installer thing, but don't know for sure.

I would prefer to just have a requirements.txt and be done with it, but this needs further investigation and thus can't be in this PR.

@miigotu any input?

I really don't care that much about certifi, but changing the version of requests is rather big to me, we lean on that a lot across the SC.

ghost:
So cfscrape and certifi have been updated in this PR. If desired, I'll drop certifi as well. Sorry about the complication and thanks for the review @WebSpider.

WebSpider (approved):
Thanks for your contribution, this will be added to develop for further testing. Base branch seems to run fine on my environment

Issue #5207, opened 2018-11-26
RFE: Don't replace existing even if it's preferred


brianjmurrell:
It would be useful to be able to set a preferred quality, but NOT replace an existing episode if it exists even if it's not preferred quality.

As an aside and additionally, it would be also useful to be able to configure a time delay between finding an episode that is not of preferred quality and downloading one from the allowed quality list.  So for example configure to wait for 2 days for a preferred quality download but download any allowed quality if preferred is not found within 2 days.

Both of these could be used to conserve bandwidth by waiting for and preferring an SD download over an HD download but to use an HD download if an SD download doesn't become available within a given time frame.

brianjmurrell:
It's only stale because nobody has even triaged it yet.

brianjmurrell:
How many times are we going to play this game of stale/not stale?

potestus:
This is quite an issue for me.. I download exclusively X265 1080p.. This is not accounted for in sickchill so I have to set custom/Unknown..  I can't set a preferred option as it just looks for 264.. As a result all of my shows that are not wanted are set to allowed (hundreds) .. This causes a huge amount of backlog scan which exceeds my provider limits..

1/ It would be good to have a "any" ie don't care option for allowed

2/ the backlog force could be graduated form. More often  in the first two days then less often gradually over the next 5..
3/ the backlog screen should allow a filter between wanted and allowed.. 
4/ a big button to do all wanted forced. 
Be good to fix the x265 thing though.

tjenvy:
I had an entire series download and almost crashed my server because of this  I had gotten them elsewhere on my own and then in the middle of the night it finally found them and tried downloading them.  It was some backlog of files form ages ago that popped up somewhere.  wiped out 175 gb of space and stopped the server.   

So yes, if a file is downloaded (by checking the directory to see if its there) then yes dont download would be great.  Maybe a checkmark somewhere?

deepseth:
> So yes, if a file is downloaded (by checking the directory to see if its there) then yes dont download would be great. Maybe a checkmark somewhere?

So, the default presets under 'preferred qualities' (Any, SD, HD, HD720p, HD1080p at least) only include 'Allowed' qualities in their list, and no 'Preferred' qualities in their list. If you stick to these, you wont have this occur.

If you use a custom 'Preferred Quality', if you restrict your selection to 'Allowed', then this wont happen. If you make selections from 'Preferred', then this will still happen.

deepseth:
[...]

PR #5977, opened 2020-02-13
Option to restrict backlog searches to 'missing' episodes only

deepseth:
Vaguely relates to #5207
Backlog searches for any missing episodes. It also searches for any episodes that are downloaded that don't match an allowed quality. What if we don't care about what quality existing files are at, and only care about new downloads matching a particular quality matrix?

E.g. We have no objection to episodes from the 1990s being in SD, but we want new episodes from 2020 to come down in HD (and not SD). Backlog doesn't understand this, and continues to hammer backlog approved providers to try and replace the early episodes anyway.

How about we make backlog not care about 'not allowed' quality downloads, and only focus on missing episodes, and make this a not-default-enabled option?

- New configuration variable to track this (default=off)
- Updates to front end to allow this to be toggled
- Updates to backend to honour this toggle


deepseth:
This was remarkably unpainful to try and do, despite having limited python knowledge and no prior experience with this codebase. Kudos to the authors.

miigotu:
Im down to allow another attempt at this, it was implemented several times before in very hacky ways and caused havoc. Could this be an int maybe and only skip episodes older than x number of days or years? 0 would skip all replacements, 1 would skip anything from last year or older, etc.

deepseth:
That'll be possible based on air date I guess, but not based on downloaded date. Whatever the default is, should not change current default behaviour or it'll screw people up who aren't expecting this, so I guess it needs to be two parameters, as a default = 0 is just asking for trouble (0 == False, someone will trip over that later and use == instead of ===). One paramter to enable (what I've already done), and one to provide a time limit.

The limit I guess should be days, months or years. I'm gonna start with it being days for now, based on the original feature request, and set the default value to 7 days (which only matters if the feature is enabled, which by default it is not)

This is going to be based on airdate, as there's no way to be based on downloaded date (that I can see). Any episode that doesn't have an air date defined will fall into the ignored category.

miigotu:
Air date is the best I think except for old shows someone just added =P but yeah enable/disable and a range with a default to disabled.

deepseth:
> except for old shows someone just added

I have no idea how to account for that, hence making the point explicit. If that's a requirement, then the age limit can't be done, and we stop at on or off as it is now.

That said, I don't think that's a problem. It doesn't ignore 'missing', only 'allowed'. If an old show has been added and an unapproved quality download has been found and is there, then it's there. Happy days.

If the goal is to improve the quality of the downloaded episodes, then this feature probably isn't for that person.


Actually, can we just hijack Backlog search day(s) (sickbeard.BACKLOG_DAYS) for this?

miigotu:
This entire feature should already be working if a person has a custom quality with only items selected in the allowed list and none selected in the preferred list btw

miigotu:
The way it works currently, it will allow anything in the allowed list but will continue to try to match a quality in the preferred list if the preferred list has selections, otherwise stick with the first found allowed found,

deepseth:
Look at a show that you have stored in e.g. SD DVD, and specify 720p HDTV as an allowed quality only. No preferred qualities selected.

All episodes of that show that do not match the allowed quality, now get listed in the manage/backlogOverview screen, and the show gets matched and triggered during a backlog search.

This prevents that from occurring.

The way you described is what my expectations were, but it didn't seem to be happening like that, which is why I wrote this. I also *hate* when default behaviour changes out from underneath me, hence making this an option rather than fixing the 'bug'.

miigotu:
Well of course if you change the quality it will start looking for something that matches allowed xD. Anyways, im ok with this being added as-is I guess

deepseth:
Ah sorry, I wasn't sufficiently clear again.

In my case, I batch added a bunch of existing shows to be tracked by Sickchill. They all defaulted to be added as quality HD, but a bunch of old shows are in DVD quality. As a result, they all got added to backlog searches, and continued to be backlog searched for over a week, resulting in rate limit hits.

Anyway, if you're happy as-is, then I'm happy to stop too! :-)

miigotu:
Only worry about the airdate option if you feel you need it, I think its a good PR as it is either way. Thanks.

miigotu:
I get a little nitpicky sometimes and I shouldnt xD

PR #6016, opened 2020-02-25
Remove port from config description

PeteLawler:
See https://github.com/SickChill/SickChill/issues/5981#issuecomment-586648035

Dunno if you wanna add code to fix this for folks in the background, but changing the instructions is a start.

Fixes #

-
-
-


miigotu:
I think it's ok to leave that on there. Port isn't required but some people need to use a different port. I did just push an update that makes it stop erroring if the port is passed in 71e05ad0496d93b9369bd1ead768e013483aa077

Comment back on this if this really needs to be done for some reason. Thanks for the PR, but I'm going to close this one for now unless we find more reason to do this.

PR #6070, opened 2020-03-06
fix login when rev proxy has auth basic

kyuuk:
If the user has a reverse proxy with basic auth enabled & the login/password are different from those on SC the user can never login,
so we must check UI login before Authorization header.

Fixes #

- check UI login before header


miigotu:
This was failing for you because SC and your reverse proxy used a different username/password, if you set the same password for both the proxy would pass the header to SC and it would authenticate you. I will merge this though.

kyuuk:
yup it was reading my proxy Authorization header so i ended on the login page but even when authenticating on SC page i was redirected to it again. Thanks for the merge :)

PR #6081, opened 2020-03-10
Docker permissions

davidfm:
- Set permissions on Docker so `/data` is not owner by root
- Application runs as `sickchill` user
- Build using `docker-compose`
- Make volume mapping configurable on compose using env vars.

Since `/data` is no longer owned by `root` appropriate permissions need to be set on the local machine so that the container can write on the local directory.


docker-compose.yaml
```diff
@@ -3,11 +3,16 @@ version: '2'
...
+        - SICKCHILL_USER=sickchill
     container_name: sickchill
     restart: unless-stopped
     ports:
       - 8081:8081
     volumes:
-      - /docker/sickchill/data:/data
-      - /docker/sickchill/downloads:/downloads
-      - /docker/sickchill/tv:/tv
+      -  ${SICKCHILL_DIR}/data:/data
```
miigotu: `${SICKCHILL_DIR:-/docker/sickchill}`
Dont want to break anything for people already using the docker image

Dockerfile
```diff
@@ -2,7 +2,10 @@ FROM python:2.7-alpine
 LABEL maintainer="miigotu@gmail.com"
 ENV PYTHONIOENCODING="UTF-8"
 
-# TODO: Handle permissions so data/config isnt owned by root
+ARG SICKCHILL_UID
```
miigotu:
```bash
ARG SICKCHILL_UID=317
ARG SICKCHILL_USER=sickchill
```
Please dont break my automatic builds, tests, and deployment lol

docker-compose.yaml
davidfm: That didn't work for me. Reverted to your original mapping

miigotu:
Whats wrong with variable expansion for the volumes again?

davidfm:
> Whats wrong with variable expansion for the volumes again?

```ERROR: Invalid interpolation format for "volumes" option in service "sickchill": "${SICKCHILL_DIR:-/docker/sickchill}/data:/data"```

I know this works on a shell script, maybe compose doesn't support it?

**Comment by miigotu** [comment]
You have to set the variable in environment for expansion to work:

```yaml
services:
  sickchill:
    environment:
      SICKCHILL_DATA_ROOT: ${SICKCHILL_DIR:-/docker/sickchill}
    volumes:
      - $SICKCHILL_DATA_ROOT/data:/data
      - $SICKCHILL_DATA_ROOT/downloads:/downloads
      - $SICKCHILL_DATA_ROOT/tv:/tv
```

Could you try that? I like the idea of making it configurable

Joshfindit:
Since this is on the table, I would like to add:
Ability to specify `UID` *and* `GUI` for the `sickchill` user, so that permissions can be frictionless in mapped volumes.

Especially important since there are options for Sickchill to delete and create files.

miigotu:
@Joshfindit this PR will already do that: https://github.com/SickChill/SickChill/pull/6081/files#diff-3254677a7917c6c01f55212f86c57fbfR5-R6

Joshfindit:
This bit me when doing a docker-compose pull today after a couple months.
@miigotu Looking at that PR, there's no mention of `GID` (though I notice I made a typo above)
`chown -R ${SICKCHILL_USER}:${SICKCHILL_USER}` *seems* to avoid the group permissions problem, but `GID` is mainly important for permissions in media folders.

Joshfindit:
*Of note: [linuxserver.io](https://docs.linuxserver.io/general/understanding-puid-and-pgid) (among others) have chosen `PUID` and `PGID` as the naming convention for mapped `UID` and `GID`s. Not saying it's required, just that it's a handy convention if the SickChill team wants it.*

miigotu:
I actually want to redo the docker file and docker-compose, so it behaves the same way the linuxserver version does. Most users expect their media files to have the same user/group as their user.

PR #6651, opened 2020-08-16
fix() Queue table td fix on try/except

Nyaran:
On try/except block, when a exception occurs, the `<td>` insde try block is printed as the fail is in the evaluated code, and then the `<td>`/`</td>` is also added, so the table is displayed with extra columns

bad_colums_status

- Move the td tags outside try/except block


miigotu:
You'll have to redo this one, I was working on this page and changed it a bunch

miigotu:
I think I complicated that one. I will work on it more tonight, and check if the service has an action or if item has show.

PR #6650, opened 2020-08-16
Feature/customize show

Nyaran:
Proposed changes in this pull request:
- New tab added to Edit Show view, "Customize"
  - Custom name: Allow to set a custom name for the show, to be used on display show, show list, episode name, etc. #5730 
  - Poster/Banner/Fanart image selector: You are able to select an image from any indexer supported by SickChill (currently only TheTVDB) and from Fanart.tv. Also you can upload it from computer. #5943 (is not really a refresh, but it allows to select the image as is also requested)

img0
img_1
img2

- [ ] PR is based on the DEVELOP branch
- [ ] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [ ] Read [contribution guide](https://github.com/SickChill/SickChill/blob/master/.github/CONTRIBUTING.md)

Issue #5943, opened 2020-01-27
[Feature]Refresh Poster


Mickroz:
Since i cannot get some posters to update when refreshing a show(with different browsers, so no cache issue), anyway we can get a refresh poster only button instead of refreshing the whole show?
And with that, maybe choose between a tvdb or fanart poster?

I know that tvdb api has some issues, but some shows refuse to update the posters, even after recreating a show, where some newly added shows do update the poster image.

Nyaran:
I'm working on this right now, I will upload a first version to my fork soon and if is stable enough I will made a PR

@miigotu Anything I need to know? My branch will be based on "develop" branch, to use Python 3

miigotu:
Have at it bud =P

Nyaran:
First version is ready: https://github.com/Nyaran/SickChill/tree/feature/customize_show

Features:

On edit show, a new tab was added, "Customize", where you can:
 - Change Show name: #5730 
 - Edit images (poster/banner/fanart): Get the image from the show indexer or from fanart.tv. Also allows to upload from computer

I will use it for a while to detect issues and also have to write the tests. Meanwhile... some screenshots =)

![img0](https://user-images.githubusercontent.com/1042532/90312371-3f1dc400-df04-11ea-940d-7da200ad9054.png)
![img_1](https://user-images.githubusercontent.com/1042532/90312373-3fb65a80-df04-11ea-95be-c6f8d5151544.png)
![img2](https://user-images.githubusercontent.com/1042532/90312372-3fb65a80-df04-11ea-8668-1b6716caf9a5.png)

Issue #5730, opened 2019-11-27
[Feature Request] changing show name


ahyear:
Hello 
Some time the show name is followed by a sub title. 
It would be nice to be able to remove it from the show name (for list and post process) or edit the name to remove "the" in the beginning of we want to.

defhead:
What about adding a scene exception name: "Edit" (from the show page) => "Advanced" => Enter a scene exception => "Add" => "Save Changes". It will match your show with the search results.

Does it solve your problem?

ahyear:
no I still have the original name in the post process show title.
the issue is note finding episodes but naming them...
![image](https://user-images.githubusercontent.com/8113672/69961255-c90b3580-150b-11ea-8d44-9c5becce707f.png)

nydave69:
Add... So many shows are remakes or copies of shows from other countries,  would be nice to be able to add to the title (US)  or (1969) whatever you want to separate them out.  Don't even suggest going to the TVDB and trying anything.

PR #6655, opened 2020-08-17
Custom show images and name

miigotu:
Replaces #6650 
(same pr, I just couldnt push to your repo ;))

### From #6650 
- New tab added to Edit Show view, "Customize"
  - Custom name: Allow to set a custom name for the show, to be used on display show, show list, episode name, etc. #5730 
  - Poster/Banner/Fanart image selector: You are able to select an image from any indexer supported by SickChill (currently only TheTVDB) and from Fanart.tv. Also you can upload it from computer. #5943 (is not really a refresh, but it allows to select the image as is also requested)

img0
img_1
img2


miigotu:
@Nyaran you agree with my changes?
Also, the tvdb one is not working for me, we would have to download them all locally and display them and I think that's crazy.
image

EDIT: My adblocker blocking it? why.

Nyaran:
I see it was merged after your changes, cool =)

About the adblock... I have it disabled for local network :(

Looking for info, I found this: https://stackoverflow.com/questions/36908023/why-does-adblock-block-my-images

Looks like, as the url contains the word "banner", the image is blocked. I will try to think for a solution, meanwhile... we can add a warning text about it.

miigotu:
I figured we can PR any needed changes xD

miigotu:
I think we might be able to proxy the request through the web server using a redirect

```
def wrap_url(self):
    url = self.get_query_argument('url')
    return self.redirect(url)

In mako:
/wrap_url?url=https://the-tvdb-image
```

miigotu:
I'll add that in a bit =P

miigotu:
https://github.com/SickChill/SickChill/commit/5728849ada8888888767df547002e009a89f3296

miigotu:
@Nyaran can you fix this? 
image

Also, is it possible to move the dropdown into the dialog header so it doesn't scroll with the images? Or pin it to the top?

Nyaran:
> @Nyaran can you fix this?
> image

That is happening also on drive browser, as is the same dialog box, but sure, I can check it
 
> Also, is it possible to move the dropdown into the dialog header so it doesn't scroll with the images? Or pin it to the top?

I will try to pin to the top

Nyaran:
I think I got the problem, on `main.mako` I see this import
```html
        <link rel="stylesheet" type="text/css" href="${static_url('css/lib/jquery-ui-1.10.4.custom.min.css')}" />
```

look the version, it sais "1.10.4", but the jquery-ui version on bower is "~1.12.1", and some changes was made on the dialog classes.

I'm not really sure how to use the css from the bower package, but as a hotfix, I can add the needed CSS to hide the text on all dialogs (drive browser and image selector)

miigotu:
I really really really want to get rid of bower and use webpack instead, but all my attempts fail lol

Issue #6743, opened 2020-09-14
Scene exceptions problem


romainsi:
Branch/Commit:master/20d84f5288ec3e532b6dd46cdbe4b833fbef00fd
OS:Linux-4.19-std-centos-7
Browser:chrome
Version: v2020.09.14-2
Database Version: 44.5
Python Version:3.6.8 (default, Apr 2 2020, 13:34:55) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]

What you did:
Shows with years in parenthesis like : 'Family Business (2019)' was unable to find.
I've try to add a Scene exception at root shows (https://github.com/SickChill/SickChill/wiki/Scene-exceptions-and-numbering) :
![image](https://user-images.githubusercontent.com/30522413/93123174-fe15fc80-f6c7-11ea-89b6-4b2581868236.png)

What happened:
With my actuals Nzb Indexers with manual API query for this season i found correctly episodes without '(YEARS)' :

```
        <item>
            <title>Family.Business.S02E01.Straight.and.Narrow.1080p.NF.WEB-DL.DDP5.1.H.264-NTb</title>
            <guid isPermaLink="true">https://indexer1/details/0a13b0cf52b0067bbe8b0827381c5f25</guid>
            https://indexer2/getnzb/0a13b0cf52b0067bbe8b0827381c5f25.nzb&amp;i=281443&amp;r=****
            <comments>https://indexer1/details/0a13b0cf52b0067bbe8b0827381c5f25#comments</comments>
            <pubDate>Fri, 11 Sep 2020 18:03:05 +0200</pubDate>
            <category>TV &gt; HD</category>
            <description>Family.Business.S02E01.Straight.and.Narrow.1080p.NF.WEB-DL.DDP5.1.H.264-NTb</description>
            <enclosure url="https://indexer1/getnzb/0a13b0cf52b0067bbe8b0827381c5f25.nzb&amp;i=281443&amp;r=****" length="1246999875" type="application/x-nzb" />
            <newznab:attr name="category" value="5000" />
            <newznab:attr name="category" value="5040" />
            <newznab:attr name="size" value="1246999875" />
            <newznab:attr name="guid" value="0a13b0cf52b0067bbe8b0827381c5f25" />
```

What you expected:
That it works ^^ ;)

Logs Sickchill : 
SEARCHQUEUE-MANUAL-364468 :: Unable to find a download for: [Family Business (2019) - S02E01 - Straight and Narrow]
[...]

PR #6752, opened 2020-09-15
Fix rebuild_exception_cache

Nyaran:
Related #6743 

- Fix the rebuild. Now it really replaces the cached data
- Update `exceptionsSeasonCache`, just to ensure is updated after the rebuild.

### DIFF
--- a/sickchill/oldbeard/scene_exceptions.py
+++ b/sickchill/oldbeard/scene_exceptions.py
@@ -324,14 +324,15 @@ def getSceneSeasons(indexer_id):
 
 
 def rebuild_exception_cache(indexer_id):
-    if indexer_id not in exceptionsCache:
-        exceptionsCache[indexer_id] = {}
-
     cache_db_con = db.DBConnection('cache.db')
     results = cache_db_con.action('SELECT show_name, season FROM scene_exceptions WHERE indexer_id=?', [indexer_id])
 
+    exceptions_cache_list = {}
     for result in results:
-        if result["season"] not in exceptionsCache[indexer_id]:
-            exceptionsCache[indexer_id][result["season"]] = []
+        if result["season"] not in exceptions_cache_list:
+            exceptions_cache_list[result["season"]] = []
 
-        exceptionsCache[indexer_id][result["season"]].append(result["show_name"])
+        exceptions_cache_list[result["season"]].append(result["show_name"])
+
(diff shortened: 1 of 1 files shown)


sickchill/oldbeard/scene_exceptions.py
```diff
@@ -324,14 +324,15 @@ def getSceneSeasons(indexer_id):
...
-        if result["season"] not in exceptionsCache[indexer_id]:
-            exceptionsCache[indexer_id][result["season"]] = []
+        if result["season"] not in exceptions_cache_list:
+            exceptions_cache_list[result["season"]] = []
+
+        exceptions_cache_list[result["season"]].append(result["show_name"])
 
-        exceptionsCache[indexer_id][result["season"]].append(result["show_name"])
+    exceptionsCache[indexer_id] = exceptions_cache_list
+    exceptionsSeasonCache[indexer_id] = exceptions_cache_list.keys()
```
miigotu, line 338: Above this `exceptionsSeasonCache[indexer_id]` is a list of integers for each indexer_id, and the exceptionsCache keys are strings.

I think you need to:
`exceptionsSeasonCache[indexer_id] = list({int(x) for x in exceptions_cache_list.keys()})`

Nyaran: That is not what I can see debuging

`exceptionsCache` keys are numbers, and that key is the indexer id.
`exceptionsCache[indexer_id]` also have keys, and that keys are numbers too, as the `season` column on `scene_exceptions` table

miigotu: Yes, but are numbers but one is `"1"` (strings) and one is `1` (int)
I mean the object type not the value

miigotu: Line 111-112
```python

if sql_results:
            exceptions_season_list = list({int(x["season"]) for x in sql_results})
```

Not sure which needs to change but the types need to be the same.

miigotu: You might be right that the sql is returning the correct type for result["season"] now on python3 (it didn't used to).
I'll have to check myself later. 

Sent you an invite to the org btw

miigotu: You were correct, sorry lol

miigotu:
I refactored sceneSeason, pushing soon

PR #6834, opened 2020-10-11
New networks logos

Nyaran:
- Add network logos:
  - ATRESPlayer Premium
  - CTV Sci-Fi (Space)
  - Fox España
  - Movistar+
  - Nippon TV
- Update network logos:
  - Nippon Television (Is the same as Nippon TV/NTV (JP)).

@miigotu an improvement proposal: As there are too many logos and a lot of them are duplicated, I think we can change it to use a dictionary instead of duplicate the logo. This will help to reduce package size and allow to update logos easly when the network changes it.


miigotu:
Honestly I think logos should live in sickchill.github.io repo and be pulled down when it is needed, but really the total size of the network images folder is pretty small considering how many there are. We could use a lookup, but I don't think there are that many dupes.

PR #7086, opened 2021-04-12
Add directory mapping for EMBY notifications

kyuuk:
replace directory from show path before sending notification

- Adding a mapping for show directory for Emby notification in case shows for SickChill does not have the same path on Emby server

This will replace the first field value from the show location with the second field value before sending the notification when it's for a specific show

example of config
image

The clear button is to erase sickchill show dir and remove entry from localStorage to prevent confusion 
Sickchill dir is a browser field, but emby dir is a text field because it's potentially in another system


miigotu:
I'm not sure I like this as a solution, at least not just for __one__ endpoint. I might try and get the update working without passing a path, or we can think about making this a more globalized mapping between client and media server so that other endpoints can also benefit. In general, I advocate that people use the same server to download AND host the media, because it is the only way to keep seeding sufficiently.

kyuuk:
i totally agree with you, specially that it should be global, and if there a solution to not use paths it would be great
but even if i'm using the same host for downloads & media, both are inside docker images and that's why i don't have the same directory, i could change the directory for the sickchill container but it means wiping the database and reconfiguring all the shows :/

kyuuk:
Hello @miigotu i made another PR  #7141 for this issue using the tvdbid for the update notification (i hope it will help)
So i'm closing this one

PR #7141, opened 2021-05-18
use tvdbId for emby show update instead of fs path

### LINKED ISSUE #7007: Jellyfin library update not working
Hi,

Jellyfin is an open source fork of Emby, which SickChill has a notification integration for.
Enabling the Emby one and pointing it at Jellyfin succeeds, at least on the test button, but in practice it seems like it does nothing - updating a show using that doesn't trigger an update on the Jellyfin side.
I can't see any errors in either logs.

Is this supposed to work, is the Emby integration compatible with Jellyfin, or is this a known incompatibility ? I know Emby clients usually work fine with Jellyfin so I would have expected this to work, but the issue might be something else entirely with my setup.
Thank you

kyuuk:
Fixes #7007

- Use endpoint /Library/Series/Added with TvdbId param for update notification on emby instead of /Library/Media/Updated

### Warning 
endpoint /Library/Series/Added is deprecated & we should be using /Library/Media/Updated 
http://swagger.emby.media/?staticview=true#/LibraryService/postLibrarySeriesAdded

### DIFF
--- a/sickchill/oldbeard/notifiers/emby.py
+++ b/sickchill/oldbeard/notifiers/emby.py
@@ -68,9 +68,9 @@ class Notifier(object):
 
             try:
                 session = self.__make_session()
-                response = session.post(url,params=params)
+                response = session.post(url, params=params)
                 response.raise_for_status()
-                logger.debug("EMBY: HTTP response: {0}".format(response.status_code))
+                logger.debug("EMBY: HTTP status: {0}, response: {1}".format(response.status_code, response.text.replace("\n", "")))
                 return True
 
             except requests.exceptions.RequestException as error:
--- a/sickchill/oldbeard/notifiers/emby.py
+++ b/sickchill/oldbeard/notifiers/emby.py
@@ -60,16 +60,17 @@ class Notifier(object):
 
             params = {}
             if show:
-                params.update({"Updates": [{"Path": show.location, "UpdateType": "Created"}]})
-                url = urljoin(settings.EMBY_HOST, "emby/Library/Media/Updated")
+                params.update({"TvdbId": show.indexerid})
+                # Endpoint emby/Library/Series/Added is deprecated http://swagger.emby.media/?staticview=true#/LibraryService/postLibrarySeriesAdded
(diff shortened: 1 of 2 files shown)


sickchill/oldbeard/notifiers/emby.py
```diff
@@ -60,16 +60,17 @@ def update_library(self, show=None):
...
+                params.update({"TvdbId": show.indexerid})
+                # Endpoint emby/Library/Series/Added is deprecated http://swagger.emby.media/?staticview=true#/LibraryService/postLibrarySeriesAdded
+                url = urljoin(settings.EMBY_HOST, "emby/Library/Series/Added")
             else:
                 url = urljoin(settings.EMBY_HOST, "emby/Library/Refresh")
 
             try:
                 session = self.__make_session()
-                response = session.post(url, json=params)
+                response = session.post(url,params=params)
```
miigotu: Leave the space after the comma here please

sickchill/oldbeard/notifiers/emby.py
```diff
@@ -60,16 +60,17 @@ def update_library(self, show=None):
...
                 response.raise_for_status()
-                logger.debug("EMBY: HTTP response: {0}".format(response.text.replace("\n", "")))
+                logger.debug("EMBY: HTTP response: {0}".format(response.status_code))
```
miigotu: We also need the response content here, emby API returns helpful messages when there is an error.

miigotu (changes requested):
With the 2 comments resolved, I will approve this PR.

sickchill/oldbeard/notifiers/emby.py
kyuuk: done :)

sickchill/oldbeard/notifiers/emby.py
kyuuk: According to both Emby and jellyfin api doc it seems there is no response body for this endpoint, but i don't know enough C# to tell that for sure so i still added the reponse in the log

PR #7187, opened 2021-06-19
No libs, pip updatable, auto install requirements

miigotu:
Support updating pip installs
Remove bundled libs
Automatically install requirements (I hope this works, startup time will be a bit slower at first)
Use poetry virtual environment in docker


miigotu:
> Can I also suggest in scene_exceptions do an exceptions fetcher for tvdb exceptoins from sickchill.github.io/sb_tvdb_scene_exceptions/exceptions.txt
> 
> This could then help build a much more complete set of exceptions.
> Ed 1: I grep'd for sb_tvdb_scene_exceptions but couldn't find any references to see where it's used

That file has been deprecated for like 5 years, we only use the json for exceptions at https://github.com/SickChill/sickchill.github.io/blob/master/scene_exceptions/scene_exceptions.json

Which also includes all of the exceptions from the old sb_tvdb_scene_exceptions

BKSteve:
> we only use the json for exceptions a

Sorry my misunderstanding as I'd added to the exception.txt file a couple of times in the past weeks and you'd accepted them so I assumed it was active.
I'll put them into the json.

PR #7200, opened 2021-06-26
Empty torrent file avoidance

BKSteve:
Attempted code to capture an empty torrent file in the failed.history.
Not sure this will work but putting it out there for suggestions or further consideration.

This came about as a torrent was put up and then taken down but the link to it remained and I would get multiple retries on the failed file every search cycle until a proper was posted.
As this deleted / empty torrent file would continually be selected and grabbed first any other torrents which met the search criteria were not checked. So this attempts to record the failed torrent file so the search can move on.


sickchill/oldbeard/search.py
```diff
@@ -119,6 +119,12 @@ def snatchEpisode(result: "TorrentSearchResult", endStatus=SNATCHED):
                 dlResult = client.sendTORRENT(result)
             else:
                 logger.warning("Torrent file content is empty")
+                # attempt to log failed
+                try:
+                    failed_history.logFailed(result.name)
```
miigotu, line 124: Should be using `sickbeard,oldbeard.processTV.process_failed` imo

PR #7341, opened 2021-07-22
Fix detection of virtualenv inside a virtualenv created by Debian 10

nocrad349:
Fix detection of virtualenv inside a virtualenv created by Debian 10's version of virtualenv

Fixes the problem with sys.base_prefix in Debian 10's virtualenv as reported in https://github.com/SickChill/SickChill/issues/7249#issuecomment-884752925

I've check this on Debian 10 without a virtualenv, inside a virtualenv created by Debian 10's virtualenv, inside a virtualenv created by a modern virtualen (20+) and inside a virtualenv created by python's venv. It works for all, but I can't test on other platforms.

Proposed changes in this pull request:
- prefer sys.real_prefix over sys.base_prefix

- [x] PR is based on the DEVELOP branch
- [x] Don't send big changes all at once. Split up big PRs into multiple smaller PRs that are easier to manage and review
- [x] Read [contribution guide](https://github.com/SickChill/SickChill/blob/master/.github/CONTRIBUTING.md)

PR #7343, opened 2021-07-22
Try and use venv module before looking for virtualenv package

miigotu:
Replace #7341


nocrad349:
I checked and this fixes the points I mentioned in https://github.com/SickChill/SickChill/issues/7249#issuecomment-884752925.
Thanks!

Issue #7470, opened 2021-08-29
Update url morethantv


Amnyz:
Hello can you update the url of provider MoreThanTV
from www.morethan.tv to www.morethantv.me
thank you

BKSteve:
amendment in develop branch

PR #7473, opened 2021-08-30
Update morethantv.py

BKSteve:
amends url per #7470


miigotu:
This alone will not fix the issue, there is also a token that needs grabbed from the login form, and some other changes to make this work.

PR #7606, opened 2021-10-31
Reduce image size using docker multistage builds

pakraticus:
Fixes #

I noticed the docker image included build-essential and rust.
Using a multi-stage build image size is reduced to 0.77GB from 1.93GB

- Separate build requirements from run requirements
- Create /data /downloads /tv directories with 0777 perms before declaring them as volumes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
-FROM python:3.9-slim
+FROM python:3.9-slim as build
 
-LABEL org.opencontainers.image.source="https://github.com/sickchill/sickchill"
-LABEL maintainer="miigotu@gmail.com"
 ENV PYTHONIOENCODING="UTF-8"
 ENV PIP_FIND_LINKS=https://wheel-index.linuxserver.io/ubuntu/
 ENV POETRY_INSTALLER_PARALLEL=false
@@ -16,7 +14,6 @@ ENV POETRY_VIRTUALENVS_IN_PROJECT=true
 # -v /etc/localtime:/etc/localtime:ro
 # -p 8080:8081 sickchill/sickchill
 
-VOLUME /data /downloads /tv
 RUN mkdir -p /app/sickchill /var/run/sickchill $HOME/.cache/pip
 RUN chmod -R 777 /app/sickchill $HOME/.cache
 WORKDIR /app/sickchill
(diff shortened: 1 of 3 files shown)


miigotu:
Problem is, the next time we update with a new dependency that requires a c/rust compiler sickchill won't be updatable inside the container.

Maybe this could be a sickchill:slim build and have an ENV variable that disables the sickchill updater completely and forces the user to pull the new image (manually, watchtower, etc)

miigotu:
Or maybe we should disable the updater completely for all docker images

kyuuk:
> 
> 
> Or maybe we should disable the updater completely for all docker images

I think it would be better to disable it, it's a docker best practice to not update app inside the image, and a new version mean a new docker image, so watchtower should do the trick

pakraticus:
I left git and the updater alone because it was the easiest way to see if there was a new version upstream.
I think the updates get reset because it "restarts" by exiting and letting Docker handle it.

I'm poking at using poetry to install to a specific venv and copy that venv to the next build stage.
That'll also trim out the ~250MB of git repository.
I have a kludge to add commit to the config file from environment variable.
However, generating a file with commit, branch, and remotes would be cleaner.

miigotu:
Currently SC updates itself inside the docker container, without pulling a new docker container. If SC happens to have a new version of a dependency with c-extensions or rust extensions it has to build them when it updates it's own virtualenv. That would fail and end in a broken container if the build dependencies are not there.

I'm not sure 1.2GB savings of disk space is a huge win vs being able to update the container rather than requiring a container redeployment. A docker pull every time the image is updated would likely pull a fairly small layer, but it forces the updater to not work.

Possibly we could put this in contrib/docker as Dockerfile.slim and start making slim builds for awhile and see how it works out (with the expected failure of SC's built-in updater possibly occuring and breaking the cointainer requiring a newer deployment)

miigotu:
> > Or maybe we should disable the updater completely for all docker images
> 
> I think it would be better to disable it, it's a docker best practice to not update app inside the image, and a new version mean a new docker image, so watchtower should do the trick

That's just not correct. That "best practice" is what synocommunity and some other embedded devs say, but that isn't best practice in the greater docker community. Everything in that docker container belongs to the application being run in the container, which is explicitly the purpose of docker in the first place.

pakraticus:
> > > Or maybe we should disable the updater completely for all docker images
> > 
> > 
> > I think it would be better to disable it, it's a docker best practice to not update app inside the image, and a new version mean a new docker image, so watchtower should do the trick
> 
> That's just not correct. That "best practice" is what synocommunity and some other embedded devs say, but that isn't best practice in the greater docker community. Everything in that docker container belongs to the application being run in the container, which is explicitly the purpose of docker in the first place.

I'll move this to a separate Dockerfile for a slim image.  I'll need some guidance to allow it to become part of the standard builds.  Can you help?

Yesterday's experiments had it trimmed down to 187MB for the image.

My perspective is production deployments in Rancher v1 and K8S sometimes in the "medical devices" industry and sometimes deployed in aws-cn.  Medical devices includes "That web site you used to get check your latest lab results."
I've had these guidelines at three companies:
Don't change the application running in the container from the application provided in the image.  Because it's a regulatory compliance nightmare when you don't.
Minimize startup time.
Minimize what an attacker may use as a shell within the container.
Minimize what the static security analysis tool has to scan.

And the combination of those four makes my job suck some days.

PR #7700, opened 2021-11-21
Make the saveShowNotifyList section actually function

### LINKED ISSUE #7686: Can't change email from Show notification list & email not sent for these addresses
Hi,

Branch: master
Commit: 01020f3636d01535f60b83464d8127ea0efabfc7
OS: Synology DSM
Browser: Chrome
What you did: change/delete show email list from http://config/notifications/#social (+"save for this show")
What happened: 
- page refreshed, but show email list was the same after refresh; only for shows with blank show email list the email are saved
- email is not sent to addresses in this field 
What you expected: 
- to change the old show email list with the new one or to be able to delete email list for some shows
- to send emails for this list
Logs: this operation don't appear in logs, not even with debug
```

```

Please advise.
Thank you

BKSteve:
Modify the `saveShowNotifyList `to save Null in DB when blanking an email address.

Removed the pointless collection of existing data as that's already done and displayed with `loadShowNotifyLists` and is in the variables when save button pressed, but if changed didn't actually take the new input email and save in the DB.

Fixes #7686 

I also couldn't work out how to pass the `Null` to the database so I put it in the string directly to the DB.

Edit: I think this must not have worked for a long time.
Add: I didn't check the sending functions but made sure the DB would be correctly updated.


BKSteve:
As a newbie I probably have done a long work around to find out which, and make changes depending on the 2 (emails or prowlAPIs) that called the `saveShowNotifyList` and then make the necessary updates or clear the field.

miigotu:
Thanks for looking into this! I'm going to go over this and clean it up a bit before merging when I get a chance.

BKSteve:
@miigotu The edits change the way the `def saveShowNotifyList` functions and you cannot have both `email` and `prowl` per show. This was also part of the initial problem too I think.

Editing a show in one deletes the other as the `saveShowNotifyList` is used by both `Prowl` and `Email` sections of the `notifications` settings.
Thus the code (562-574) to know which of the 2 called the `saveShowNotifyList' and only update that item and not the other.

Add: I couldn't work out how to know whether prowl or email called the saveShowNotifyList so that's why the many lines and ifs

Issue #7739, opened 2021-12-09
Better provider warning logs


gits68:
Branch/Commit: 01020f3636d01535f60b83464d8127ea0efabfc7
OS: docker
What you did: [View Warnings](https://localhost:8081/errorlogs/?level=30)
What happened: 2021-12-09 04:22:43 Unable to connect to provider
What you expected: 2021-12-09 04:22:43 **YggTorrent:** Unable to connect to provider

logs from the web interface are incomplete, this require to go to sickchill.log to find the right information.

Proposed change:

```
sed -i 's/"Failed to download any results"/"{0}: Failed to download any results".format(self.name)/' \
    /app/sickchill/sickchill/providers/GenericProvider.py
sed -i 's/"Unable to connect to provider"/"{0}: Unable to connect to provider".format(self.name)/' \
    /app/sickchill/sickchill/oldbeard/providers/*.py
```

well, I suppose this change have to be done to every **logger.warning()** call.
would be better to adapt the **/errorlogs/?level=30** log viewer...

Logs before:
```
2021-12-09 04:07:38 WARNING :: SEARCHQUEUE-DAILY-SEARCH :: [YggTorrent] :: Unable to connect to provider
```

Logs after:
```
2021-12-09 04:07:38 WARNING :: SEARCHQUEUE-DAILY-SEARCH :: [YggTorrent] :: YggTorrent: Unable to connect to provider
```

BKSteve:
Rather than doing all this `sed` stuff you could just branch and do your own pull requests.
The level of effort on your part wouldn't be much different.

On the logs you show, what's the real difference in actual information? It seems to be duplicate `YggTorrent`

> Logs before:
2021-12-09 04:07:38 WARNING :: SEARCHQUEUE-DAILY-SEARCH :: [YggTorrent] :: Unable to connect to provider
Logs after:
2021-12-09 04:07:38 WARNING :: SEARCHQUEUE-DAILY-SEARCH :: [YggTorrent] :: YggTorrent: Unable to connect to provider

Issue #7738, opened 2021-12-09
YggTorrent search URL has changed


gits68:
Branch/Commit: 01020f3636d01535f60b83464d8127ea0efabfc7
OS: docker

yggtorrent went from .nz to .re
would be better to use **yggtorrent.to** which point to the last good url

```
sed -i s/www.yggtorrent.nz/yggtorrent.to/ \
    /app/sickchill/sickchill/oldbeard/providers/yggtorrent.py
```

PR #7741, opened 2021-12-09
Providers update - Generic, torrent9, yggtorrent

### LINKED ISSUE #7737: Torrent9 search URL have changed
Branch/Commit: 01020f3636d01535f60b83464d8127ea0efabfc7
OS: docker

torrent9 have switched from .pw to .si.
better to backout change from .pw to .one which always point to the right url [patch](https://github.com/pakraticus/SickChill/commit/d0e59cd3636b6fbe65fbbd876030031d96363808)
in fact, you miss my late comment in [#7654](https://github.com/SickChill/SickChill/issues/7654)

BKSteve:
Fixes #7737 #7738 changing the url for these. `.one` and `.to`

#7739 only GenericProviders.py logger.warning only.   No change for all providers.


sickchill/providers/GenericProvider.py
```diff
@@ -109,7 +109,7 @@ def download_result(self, result):
                 remove_file_failed(downloaded_filename)
 
         if urls:
-            logger.warning("Failed to download any results")
+            logger.warning("{0} : Failed to download any results".format(self.name))
```
miigotu: I have been working on utilizing what is called "f-strings", and this should be in this format:
```
logger.warning(f"{self.name} : Failed to download any results")
```

BKSteve: On future file edits I shall follow this directive for 'f' strings

PR #7757, opened 2021-12-18
Poetry Packages updater to improve load speed

### LINKED ISSUE #7747: Python packages reinstalled at each launch
Having some issues on my Synology NAS after last update, I have uninstalled and reinstalled Sickchill to start from scratch.
I now have a new error message on the top bar: `No NZB/Torrent providers found or enabled for daily searches and backlog searches. Please check your settings.`
When I go to the `Search Provider` settings, I see the msg `no providers available to configure. `.

I started debugging the issue reading the code and, importing some modules in the Python console, I noticed that the [__init__.py](https://github.com/SickChill/SickChill/blob/master/sickchill/__init__.py#L4) file is installing the Python packages at each lunch. This is because the function [check_installed()](https://github.com/SickChill/SickChill/blob/master/sickchill/init_helpers.py#L378) always return `False` on my NAS.

This also explains why, when I start Sickchill from the Synology web interface, it takes a long time (many minutes) to have the Sickchill's webserver up and running and display the web interface.

I tried to comment out that line `poetry_install()`, and the start-up is much faster, less than a minute.

Is this behaviour wanted for some reasons or a bug?

BTW, I still have to figure out why I get the error `No NZB/Torrent providers found or enabled for daily searches and backlog searches.`.

Branch: master
Commit: 01020f3636d01535f60b83464d8127ea0efabfc7
Database Version: 44.5
OS:  Linux-3.2.40-i686-with-glibc2.0 (Synology DSM 6)
Python: 3.8.12
Browser: Any
Logs: No errors in the logs

BKSteve:
Fixes #7747 

- change the poetry string sent to pip, 
- compare installed against poetry string, 
- cut out the items that are the same.
- send short list to pip to attempt to install only those.


BKSteve:
contemplating some ideas on why the DSM is slow on this and it's probably that it actually tries all the `python38` installed packages. So this is the first try but will test if just pulling the python38 packages from the list resolves the overall time issue. 
But that said this does speed things up a lot.

BKSteve:
In the search to find why loading is so slow, mismatch of package parameters is noted.
In pyproject.toml the packaging allows up to 22.0 but poetry 1.1.12 is >20.4,<21.0 so changing it reduced restart time to less than 30 seconds.

miigotu:
Im sketchy about adding the stdlib argument for everyone. If it is only an issue on DSM, that could be added as an environment variable directly to the package or the OS

BKSteve:
Yeah, it was a little odd on the original #7747 as my DSM had no issues and would reload and just take longer to get functional when run from the package centre. When run from command line it would show the error, so if you were doing a new install it would fail and not start but when doing a restart it would continue on.

I tried to find a way to know whether on a DSM by looking at the other code but didn't find something easy.
If you can point me to something that I can use to know if on DSM then I can tweak the code and do the minimisation too just for DSM and just process the poetry string through for other OS.

And yes the old 
`187     packages = re.sub(r'Warning.*', "", packages`
may also no longer be required, so will look at this with the above.

miigotu:
Honestly, I really want to plan a path forward that removes the poetry code completely out of SC and just expect the correct packages to be installed correctly. SC should always be installed, and at most we should have to do "pip install -U sickchill" when the updater verifies they aren't on the newest SC. Problem is, not everyone has a computer and we need to completely disable the updater on their platform.

I forget the method we used before to determine if someone is on dsm, but it's something like /etc/platform or lsb_release

BKSteve:
Most definitely a simple install where the pip line could do it all would be great. Simpler package...

I'll fish around some more to see if I can find in the code how to know if DSM.

BKSteve:
Modified the code so DSM get the `SETUPTOOLS_USE_DISTUTILS=stdlib` and not all installs.
Cleaned up coding warnings from pycharm.
Added log to start and complete poetry packages check. This takes 30 seconds.
The start time on my DSM412+ reduces from 1:55 to 1:05 so 50 seconds faster.

Add: missed the RPi and Opensuse needs setuptools string too.

BKSteve:
> Im sketchy about adding the stdlib argument for everyone. If it is only an issue on DSM, that could be added as an environment variable directly to the package or the OS

@miigotu my idea of using the OS to choose to run the poetry call with or without `SETUPTOOLS_USE_DISTUTILS=stdlib` knowing that this issue is with RPi and OpenSuse too, and not knowing if other OS are having the same issue.
I'm wondering if going back to the earlier filtering and taking the error out of the returned poetry string before pip of each item as the easiest way to resolve per https://github.com/SickChill/SickChill/pull/7757/commits/295a406c0caffc9f9790f396e0ea95da28061338
And abandon doing by OS. As 2 lines of `re.sub` saves more lines of if and variable definitions.

miigotu:
Yeah I rather that I think, because even though it's a patchy type thing, I want to remove peotry calls altogether soon.

BKSteve:
Was already working the code. So `Remove all warnings from poetry string` takes out the strings if they occur.
This latest is much better than the first time, probably more code experience.

Add: The Python38 package on syno DSM probably loads too much. `importlib-metadata` one of their packages they have dependencies `typing-extensions` and although SC loads latest `zipp` poetry always has `zipp==3.6.0` so kill that too with the py38pkg manual list.

Add2: The `poetry` call takes 30 seconds to process on my DSM 412+ that's a long time to achieve it's task.

PR #7981, opened 2022-05-26
Anime Quality - Parser Regexes

### LINKED ISSUE #7978: Post Processing Error Anime

Branch/Commit: v2021.07.14-1 
OS: Linux-5.10.28-Unraid-x86_64-with-glibc2.33 
Browser: Firefox
What you did:
Run Post Processing
What happened:
Episode was not parsed correctly
What you expected:
Logs:
```
2022-05-24 07:22:56 DEBUG :: POSTPROCESSOR :: Starting new thread: POSTPROCESSOR
2022-05-24 07:22:56 INFO :: POSTPROCESSOR :: Auto post processing task for /media/Downloads/Bittorrents/_sickchill was added to the queue
2022-05-24 07:22:57 INFO :: POSTPROCESSOR-AUTO :: Beginning auto post processing task: /media/Downloads/Bittorrents/_sickchill
2022-05-24 07:22:57 DEBUG :: POSTPROCESSOR-AUTO :: Processing in folder /media/Downloads/Bittorrents/_sickchill
2022-05-24 07:22:57 INFO :: POSTPROCESSOR-AUTO :: Processing /media/Downloads/Bittorrents/_sickchill
2022-05-24 07:22:57 DEBUG :: POSTPROCESSOR-AUTO :: Processing folder /media/Downloads/Bittorrents/_sickchill
2022-05-24 07:22:57 DEBUG :: POSTPROCESSOR-AUTO :: Could not properly parse a show and episode from [[Judas] One Piece - 1017.mkv]: Unable to match [Judas] One Piece - 1017.mkv to a show in your database. Parser result:  - S10E17 [GROUP: [Judas] One Piece ] [ABD: False] [ANIME: False] [whichReg: ['stupid']] Score: -4
2022-05-24 07:22:57 DEBUG :: POSTPROCESSOR-AUTO :: /media/Downloads/Bittorrents/_sickchill : No processable items found in folder
2022-05-24 07:22:57 INFO :: POSTPROCESSOR-AUTO :: Successfully processed
```

I fixed this locally, but the fix seems so odd that I can't believe that it hasn't been encountered before... so I feel like I must be missing something. I updated the regex for anime_standard to make the "Source_Quality_Etc-" optional (see my updated regex below).
`sickchill\oldbeard\name_parser\regexes.py`
```
(
        "anime_standard",
        # [Group Name] Show Name.13-14
        # [Group Name] Show Name - 13-14
        # Show Name 13-14
        # [Group Name] Show Name.13
        # [Group Name] Show Name - 13
        # Show Name 13
        r"""
     ^(\[(?P<release_group>.+?)\][ ._-]*)?                        # Release Group and separator
     (?P<series_name>.+?)[ ._-]+                                 # Show_Name and separator
     (?P<ep_ab_num>((?!(1080|720|480)[pi])|(?![hx].?26[45]))\d{1,4})                                       # E01
     (-(?P<extra_ab_ep_num>((?!(1080|720|480)[pi])|(?![hx].?26[45]))\d{1,4}))?                             # E02
     (v(?P<version>[0-9]))?                                       # version
     ([ ._-]+\[(?P<extra_info>\d{3,4}[xp]?\d{0,4}[\.\w\s-]*)\])?       # Source_Quality_Etc-
     (\[(?P<crc>\w{8})\])?                                        # CRC
     .*?                                                          # Separator and EOL
     """,
    )
```

BKSteve:
Fixes #7978 

- set the quality requirement in the anime name to optional per the naming samples.
- align quality requirement with that of SxxExx format.


miigotu:
Here's the problem with changing it this simply:

`Show Name - Episode 01 - Ep Name.avi` now matches that regex, which is not an absolute number file name. This is a `Show Name/Season x/Show Name - Episode 01 - Ep Name.avi` file name. 

So instead of editing that existing regex, you need to add a new regex that copies the one you edited, except it should only match when there are 3+ numbers for the absolute number part if there is no source. That way the old anime regex still is strict, but the new one will match your bare absolute number regex with no source/quality. I'd much rather the new regex be optional also, with a setting like "strict_anime_matching" which defaults to true and would disable that anime regex unless a user did the setting.  Some people care about strict matching more than others, for some it is paramount. For other people they just want to match everything and deal with the mismatches manually later.

BKSteve:
Points noted, didn't fully consider the regexes usage and being used for both collecting and processing.
Being able to post process with the modified version of anime_standard in addition to existing strings is something to look into.
I think processing was what @Linden-Ryuujin was also doing in #7978 too.

Will review regexes further.

Add: Not sure then why the last regex `anime_bare` didn't process as it's effectively `anime_standard` without quality wildcard end?

BKSteve:
More investigation shows that if processing show name `[Judas] One Piece - 1017` the `NameParser` in `sickchill/oldbeard/name_parser/parser.py` starts without checking the `shows` anime selection and process it through the `ALL_REGEX` which then does both `regexes.normal_regexes, regexes.anime_regexes` 
With the `normal` giving matches with 9 (stupid) and 13 (bare) and `anime` with 16 (anime_bare) which results in the winner being the lesser 9 being `stupid` which is stupid.

`sickchill/oldbeard/name_parser/parser.py`
Added logger lines to debug in latest commit 
> 100 logger.debug(f"result {result}, {result.which_regex}; {result.score} result.score")
106 logger.debug(f"result.series_name {result.series_name}")

result
> DEBUG :: POSTPROCESSOR-AUTO :: Processing folder /home/stephenr/TVShow
DEBUG :: POSTPROCESSOR-AUTO :: result  [ABD: False] [ANIME: False] [whichReg: ['stupid']] Score: -9, ['stupid']; -9 result.score
DEBUG :: POSTPROCESSOR-AUTO :: result.series_name 
DEBUG :: POSTPROCESSOR-AUTO :: result  [ABD: False] [ANIME: False] [whichReg: ['bare']] Score: -13, ['bare']; -13 result.score
DEBUG :: POSTPROCESSOR-AUTO :: result.series_name [Judas] One Piece
DEBUG :: POSTPROCESSOR-AUTO :: result  [ABD: False] [ANIME: False] [whichReg: ['anime_bare']] Score: -16, ['anime_bare']; -16 result.score
DEBUG :: POSTPROCESSOR-AUTO :: result.series_name One Piece
DEBUG :: POSTPROCESSOR-AUTO :: Could not properly parse a show and episode from [[Judas] One Piece - 1017.mkv]: Unable to match

What surprised me is that the `stupid` returned a blank result but a match in `series.name`


So resolutions?
1. adjust the `stupid` regex of `(?P<release_group>.+?)(?<!WEB)-(?P<series_name>\w*)(?<!\d)[\. ]?(?!26[45])(?P<season_num>\d{1,2})(?P<ep_num>\d{2})$` in particular the \w* so it matches only the type stupid is looking for. `# tpz-abc102` and adjust `bare` regex similarly.
4. work the shows anime yes/no into the `parser.py` so it doesn't run through the `ALL_REGEX` because we've selected `ANIME`
5. Add some more checking of `series_name` in parser.py to check if blank or contains a `[` and set the result score += -100

Linden-Ryuujin:
I have to admit I'm a bit confused by the name parsing loop, and I didn't spot anime_bare (as I stopped when I saw a code example that matched the file I had). If there are multiple matches should it not just drop through? If bare fails to find a match then it just tries the next pattern? To be honest I thought this is what was happening already, but it is a bit of the deep stack to track and I guess at some point it stops moving onto the next one.

BKSteve:
It's a logic and numbering calculation. As every regex under normal is assigned a number as it's iterated through, then same for anime regex. 
That is then numbered. 
The problem is `stupid` at 9 even though no output is considered a better match than `bare` at 13 or in anime regex `anime_bare` at 16.
So it got late for me when I posted last message and will need to examine the regex in normal further to understand why both 9 and 13 are accepted. This will also need some review of parser.py.

BKSteve:
Think these 2 little tweaks have resolved the problems on processing anime and reverted the regex so no new search issues are created, which adjusting the regexes caused.

Linden-Ryuujin:
Looks good to me, not sure if I'm eligable to review though (given I'm not even sure what the regex issues were). @miigotu are you able to review?

sickchill/oldbeard/name_parser/parser.py
```diff
@@ -97,14 +96,19 @@ def _parse_string(self, name, skip_scene_detection=False):
...
             named_groups = list(match.groupdict())
 
             if "series_name" in named_groups:
                 result.series_name = match.group("series_name")
+                logger.debug(f"result.series_name {result.series_name}")
                 if result.series_name:
                     result.series_name = self.clean_series_name(result.series_name)
                     result.score += 1
+                else:
+                    result.score += -20
```
miigotu: result.score -= 20

miigotu (commented):
I'll have to test this, but please use -= rather than += -

BKSteve:
> I'll have to test this, but please use -= rather than += -

Apologies. Rookie programmer mistake.

@Linden-Ryuujin 
It wasn't a regex issue really, it's in how the code was processing the results. Sent me down a rabbit hole and showed 2 items. First was a no name situation having a higher score than a real result (causing error) and second when cross checking if the show was an anime, not parsing the full information and not resulting in using the anime regex instead of a normal regex during post processing (another error).

sickchill/oldbeard/name_parser/parser.py
BKSteve: noted

BKSteve:
bump

miigotu:
Can you make this so it only includes the parsers and regex changes, without the workflows, lock file, and package,json changes? It is conflicting,

BKSteve:
You're challenging my GitHub skills to kick these changes out.
As I did this one using smartgit which has now locked me out and I've learnt more how to do in PyCharm.
~~Might be easier if I create a new pull request.~~
GitHub desktop to the revert squash rescue.

miigotu:
I'm still worried about what this one will do, because we use the same parser for files when we don't have the show name and they are valid files, like on show refresh with existing files that were renamed by SC.

More testing needs done (or tests added) before we can take this one in.

BKSteve:
Allow me to explain what I was seeing when testing and why I went for a -20 if the show had no name.
The check cycles through twice once with normal regexes and then anime regexes. 
The regexes effectively get a minus number based on where they are in the sequence of regexes.
Normal -1 to -10 and Anime -1 to -16 (as far as I recall)

So a no name `stupid` show starts at -9 and gets are few things in the maths down to -2 so +7 points.
The show I was testing was a named anime so it started at -16 ended at -6 (+10) but the  `stupid` normal effectively was the better score so the processing of the file would fail.
Therefore I put in the 20 point penalty if it didn't have a name, it could still be the sole winner but wouldn't usurp more valid regexes.

Then I saw that if I have the show registered as anime in my database this wasn't properly being taken into account by using it in the logic when examining the show name.
`best_result.show = show_anime` 
thus if the show is recognized and it's tagged as anime in my database it can actually sets the best result show to the show name and process too.

miigotu:
Yes, that makes sense when post processing. But the code there is also used for your shows library.

Downloads/xyzabc123.mkv this makes sense
But Shows/House/S01E01.mkv not so much so I'm unsure there. It _should_ know what regex type to use when doing a show refresh, which would mitigate the issue I'm worried about. Since Shows/House is in the database as House, and House is not marked as an anime, the anime regexes are not (at least should not) be compiled and used at all. If the show is anime, it should compile only the anime regexes. For the post processor, we actually can look back at snatch history and determine which regexes to use, by checking which show it belongs to (I'm not sure we do that currently). And as a last resort use regex_all for files that are in the post processing folder but aren't in our snatch history. 

I'm curious if we could replace this entirely with guess - it.guess and stop maintaining these complicated regexes. In any event we really should rework this in a much more intensive way, and absolutely get some better testing in place. I'm down to help over the next week or two.

BKSteve:
That looks interesting - https://github.com/guessit-io/guessit
the documentation https://guessit.readthedocs.io/en/1.0.2/ also has some `other featuers`

>If you have the ‘enzyme’ python package installed, GuessIt can also detect the properties from the actual video file metadata.

miigotu:
Yeah we already have guessit and enzyme installed for subliminal/subtitles

BKSteve:
Did some grunt work on episode names pulled from Regexes and a couple of others. Guessit gets most things.

[GuessIt-RegexNames.ods](https://github.com/SickChill/SickChill/files/9175922/GuessIt-RegexNames.ods)

PR #8015, opened 2022-07-14
Add Show Year and bracket of year features

BKSteve:
Fixes #7995
Adds feature 

- add new Show Name type with have Show Start Year, different to all other Year variables.
- add No Bracket variable to be able to remove or have brackets around Show Year.
- lists the additional parameters in Naming table.


sickchill/tv.py
```diff
@@ -2151,11 +2160,25 @@ def release_group(show, name):
 
         epStatus_, epQual = Quality.splitCompositeStatus(self.status)
 
+        # set the differnet show name variables and only 4 digit years
+        show_name = self.show.name
+        show_name_no = re.sub(r"\(\d{4}\)$", "", show_name).strip()
```
miigotu: `show_name_no` doesn't really tell an unsuspecting future developer what should be in this variable. It should be fully descriptive, even if it seems long. Maybe `show_name_only` or `show_name_no_year` or `show_name_without_year`

sickchill/tv.py
```diff
@@ -2151,11 +2160,25 @@ def release_group(show, name):
...
+        # if the start year is less than 1 character there isn't one so revert to show name (safety check)
+        if len(self.startyear) > 1:
```
miigotu: It is sufficient to just check if, because an empty string in python is false. 

```python
if self.startyear:
```

miigotu: I'd like to make a few changes here in this logic section, using f-strings and some cleaner structure that branches less. I will try to do it soon.

miigotu:
This one needs formatting fixed.
I have many of the tasks defined in pyproject.toml and use poethepoet to execute those tasks. The rest of the tasks are called using grunt, but that's for another time to explain, because I want to move those tasks out of grunt.

```bash
# Switch to this branch
git checkout SeriesYearB
# Install/update poetry and poethepoet if they aren't already 
pip install --update poetry poethepoet
# Get the dev dependencies installed, so poe works correctly 
poetry install --no-root
# Call the poe task
poe format
# Add the changed files to the change set and sign the commit, with a message. You should look into setting up signed commits, because chain of trust is always nice. 
git commit -asm "Fix formatting"
# Check the changes 
git diff origin/SeriesYearB
# Push the changes to this branch
git push origin SeriesYearB
```

Next time, you should just be able to call `poe format` or `poe lint` or `poe depup` or any of the defined tasks.

BKSteve:
Some more learning to do.
This whole branch I did in PyCharm, does it have such formatting abilities or need to use `poe` for this?

sickchill/tv.py
BKSteve: My coding logic is from a long time ago. Pascal and Fortran were the languages taught.
If you have a link to f-string guidance please share. The average search doesn't give much relevant.

miigotu:
Pycharm does have formatting, but it misses some stuff. I do everything in pycharm too but I use the terminal in pycharm and have a virtual environment for the project set up.

BKSteve:
`poe format` made no changes.
I ran it in the PyCharm terminal and then reloaded the project to see if anything changed and nothing??

sickchill/tv.py
BKSteve: `show_name_no_year`

sickchill/tv.py
BKSteve: changed and did the f-string for `show_start_year` much neater

miigotu:
> `poe format` made no changes. I ran it in the PyCharm terminal and then reloaded the project to see if anything changed and nothing??

I'll look at it once I get the lxml branch build working, because I had some formatting errors in my js also that is breaking your build 😂

miigotu:
This one should build once I merge the lxml PR and we update this branch with those commits.

sickchill/tv.py
```diff
@@ -2120,7 +2125,11 @@ def _replace_map(self):
 
         def dot(name):
             # assert isinstance(name, unicode), f'{name} is not unicode'
-            return helpers.sanitizeSceneName(name)
+            # re.sub str . was any char (messy) and sanitizer wouldn't allow brackets to remain without other issues.
```
miigotu, line 2128: Was the `.` in brackets? Because `[.]` or `\.` is a literal `.`

BKSteve: No the `.` was outside the brackets. As in the name of the show was already `Show Name (2021) S02E03' and when it to put the `.` went through the sanitizer to put the name to `Show.Name.(2021).S02E03` it would come back without brackets and missing characters like 'how.ame.2021.S02E03)`. It was weird but brackets would always be taken out.
I escaped the `.` with `\.` with a regex too but it didn't resolve properly. 

As the names have already been sanitized when show is added I made the assumption that we only need to change those characters in the name already. I left `-` in as it was in one regex below and wouldn't hurt. This bit of code came from one of the sanitizers too.

BKSteve:
Better prepare some text and screen shots to update the wiki as well as the other margins one.
Replace the images with SR too.

Issue #8042, opened 2022-07-28
Anime shows produce "Couldn't find release in snatch history" warning when postprocessed


Smokey23:
Branch/Commit: PIP
OS: Windows 10
What you did: Nothing

What happened: Starting on 7-22, I began getting the warnings below. I have no records of that warning before the 22nd.  It appears that the warnings are showing up for Anime, but not for regular tv shows.

What you expected:  No warning?

Logs:
```
2022-07-26 08:48:40 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history
2022-07-27 08:49:32 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history
2022-07-27 08:49:49 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history
2022-07-25 12:47:58 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history
2022-07-23 08:46:06 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history
2022-07-24 04:46:50 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history
2022-07-24 08:46:58 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history
2022-07-22 10:15:18 WARNING :: POSTPROCESSOR-AUTO :: Couldn't find release in snatch history

2022-07-27 08:49:49 DEBUG :: POSTPROCESSOR-AUTO :: Retrieving episode object for 2x4
2022-07-27 08:49:49 DEBUG :: POSTPROCESSOR-AUTO :: Snatch history had a quality in it, using that: 720p HDTV
2022-07-27 08:49:49 DEBUG :: POSTPROCESSOR-AUTO :: Quality of the episode we're processing: 720p HDTV
2022-07-27 08:49:49 DEBUG :: POSTPROCESSOR-AUTO :: SC snatched this episode and it is not processed before
2022-07-27 08:49:49 DEBUG :: POSTPROCESSOR-AUTO :: Is ep a priority download: True
2022-07-27 08:49:49 DEBUG :: POSTPROCESSOR-AUTO :: There is no existing file so there's no worries about replacing it
2022-07-27 08:49:49 INFO :: POSTPROCESSOR-AUTO :: This download is marked a priority download so I'm going to replace an existing file if I find one
[...]

PR #8043, opened 2022-07-28
Anime Warning if not in snatch history.

BKSteve:
Fixes nothing. Start a discussion.

### DIFF
--- a/sickchill/oldbeard/postProcessor.py
+++ b/sickchill/oldbeard/postProcessor.py
@@ -880,6 +880,7 @@ class PostProcessor(object):
         """
 
         self._log("Processing " + self.directory + " (" + str(self.release_name) + ")")
+        logger.debug(f"DEF PROCESS self.release_name {str(self.release_name) }")
 
         if os.path.isdir(self.directory):
             self._log("File {0} seems to be a directory".format(self.directory))
--- a/sickchill/oldbeard/show_name_helpers.py
+++ b/sickchill/oldbeard/show_name_helpers.py
@@ -143,8 +143,12 @@ def determine_release_name(directory=None, release_name=None):
             return release_name
 
         logger.info(_("Using release for release name."))
-        hell = release_name.rpartition(".")
-        logger.debug(f"{hell} return rpartition")
+        release_name_rpart = release_name.rpartition(".")
+        if release_name_rpart[0]:
+            release_name_rpart_return = release_name_rpart[0]
+        else:
+            release_name_rpart_return = release_name_rpart[2]
+        logger.debug(f"{release_name_rpart_return} return rpartition")
         return release_name.rpartition(".")[0]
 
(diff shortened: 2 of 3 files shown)


BKSteve:
@miigotu I've played with this and if the filename has no `.` in it then it will kick a Warning #8042 but if it has then there is no warning.
Put some log lines in the 2 files so they are in this pull request.

I was trying to understand the `show_name_helpers.py` line `return release_name.rpartition(".")[0]` and what it was to achieve.

Tested examples with tuple result `hell` 
`[GM-Team] Swallowed Star. 30.mkv` processes without Warning. `('[GM-Team] Swallowed Star', '.', ' 30') return rpartition`
`[GM-Team] Swallowed Star 45.mkv` processes with Warning `('', '', '[GM-Team] Swallowed Star 45') return rpartition`

BKSteve:
Is this a valid answer?
Just playing around. If there are no `.` then use the full string [2] else use [0]
Not that it makes any difference as the next log line uses the full string anyway.

miigotu:
The release name can either be an actual release name, or a file name in the format release_nam.mkv for example. The file name will not match the release name with the extension, and we also want to match release names that have the extension, with the same release from another tracker that happens to not have the extension on it, so we don't download the same release in a few situations, such as if that particular is marked as failed.

miigotu:
`release_name.rpartition(".")[0]` will always contain the full string of the string does not contain an extension. But it becomes a problem if the release name has no extension but it has dots. We need to check if it's a media file name first, and only repartition if it is to drop the extension. helpers.remove_extension also does this.

I don't have internet right now, unscheduled isp maintenance bs. So I can't search really easily on my phone. I was working on improving the gruntfile for releases when they did this to me smh.

BKSteve:
I looked up str.rpartition it doesn't  give full string in [0] if not found 
(https://docs.python.org/3/library/stdtypes.html#str.rpartition)
Split the string at the last occurrence of sep, and return a 3-tuple containing the part before the separator, the separator itself, and the part after the separator. If the separator is not found, return a 3-tuple containing `two empty strings, followed by the string itself.`

Add: The values passed to the `determine_release_name` already has the extension removed. Thus I ask the question.

miigotu:
Wow, you are right. You might have just found a major oversight by me and this might solve a big problem. This is actually to drop nzb/nzb (or any media file extension). I'll need to look at the code also on my laptop, but if you want to you can make a PR for it, and see if I made the same mistake on rpartition anywhere else. 

I want to do a git blame and follow it back and see if I was the first one to do this lol.

BKSteve:
I think a simple change from rpartition to rsplit will give the desired [0] with string.
`return release_name.rsplit('.', 1)[0]`

BKSteve:
Crashed and burned. One line so created new one.

PR #8079, opened 2022-08-16
Fix file.extension double dot

### LINKED ISSUE #8074: Post processing issue.  When creating episode nfo, extra period is added before file extension
I have post processing set to move file with KODI metadata.  When it moves the file and makes a new nfo file, it puts an extra period before the file extension.  so episode would be XXX.mk and nfo will be XXX..nfo.  Not sure when this started but it seems to be with one of the newer updates in the past couple of days.

its not making any entries in the log as far as i can see to add here.

edit:
forgot to add the post processing is also set to rename the file.  But not sure that is important as it renames the video file correctly, just not the nfo. None of the videos i  had processed yet have downloaded a jpg so not sure if its also doing it on jpgs or just nfos

BKSteve:
Fixes #8074 
Database getting `..` instead of `.` stored so rename and other extensions being corrupted.

### DIFF
--- a/sickchill/oldbeard/postProcessor.py
+++ b/sickchill/oldbeard/postProcessor.py
@@ -1058,7 +1058,7 @@ class PostProcessor(object):
             old_path = Path(self.filename)
             orig_extension = old_path.suffix
             new_base_name = os.path.basename(proper_path)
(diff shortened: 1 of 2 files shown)


BKSteve:
revert to fixing the double dot issue only.

PR #8087, opened 2022-08-17
chg: Fix case sensitivity supportedParams property for newznab

### LINKED ISSUE #8055: not snatching episodes since 2022.07.20
SickChill is not finding any episodes when searching against NZBHydra via API. I can go into NZBHydra manually, see the API search returned results and download manually. Currently running on Ubuntu/Python 3.9 but I've encountered same issues on MacOS and Python 3.7, 3.8 and 3.9.

Version info:

Branch: [master](https://blankrefer.com/?https://github.com/SickChill/SickChill/tree/master)
Commit: [d3d9e0cd9ffe9343563ccbecd5a6c5a55878d0d5](https://blankrefer.com/?https://github.com/SickChill/SickChill/commit/d3d9e0cd9ffe9343563ccbecd5a6c5a55878d0d5)
Version: [2022.07.20](https://blankrefer.com/?https://github.com/SickChill/SickChill/releases/tag/2022.07.20)

This is the error now, that is likely related to the issue, when performing searches against NZBHydra:

Performing episode search for *******************
/opt/SickChill/.venv/lib/python3.9/site-packages/bs4/builder/__init__.py:545: XMLParsedAsHTMLWarning: It looks like you're parsing an XML document using an HTML parser. If this really is an HTML document (maybe it's XHTML?), you can ignore or filter this warning. If it's XML, you should know that using an XML parser will be more reliable. To parse this document as XML, make sure you have the lxml package installed, and pass the keyword argument `features="xml"` into the BeautifulSoup constructor.
  warnings.warn(

tking2:
Potentially fixes #8055

- Update the `supportedparams` property for newznab provider to be the correct case sensitivty `supportedParams` following language change to XML in the BS4 processor

Unfortunately I am unable to test whether this resolves all issues as I do not have a configurable instance of SickChill, but it should be a step in the right direction.

### DIFF
--- a/sickchill/oldbeard/providers/newnewznab.py
+++ b/sickchill/oldbeard/providers/newnewznab.py
@@ -121,7 +121,7 @@ class NewznabProvider(NZBProvider):
         elm = data.find("tv-search")
         self.use_tv_search = elm and elm.get("available") == "yes"
         if self.use_tv_search:
-            self.cap_tv_search = elm.get("supportedparams", "tvdbid,season,ep")
+            self.cap_tv_search = elm.get("supportedParams", "tvdbid,season,ep")
 
         self._caps = any([self.cap_tv_search])
 


miigotu:
Thanks for this. Definitely overlooked that we needed to respect case (idk why I didn't write it that way to begin with).

brianjmurrell:
How long will this take to move from develop to master so that those following mainline will get it?

PR #8092, opened 2022-08-19
Regex fixes

BKSteve:
Make regexes work with both upper and lowercase sS and eE regexes.
Except one which specifically identifies lowercase characters

### DIFF
--- a/sickchill/oldbeard/name_parser/regexes.py
+++ b/sickchill/oldbeard/name_parser/regexes.py
@@ -8,10 +8,10 @@ normal_regexes = [
         # Show Name - S01E02 - S01E03 - S01E04 - Ep Name
         r"""
      ^(?P<series_name>.+?)[. _-]+                # Show_Name and separator
-     s(?P<season_num>\d+)[. _-]*                 # S01 and optional separator
-     e(?P<ep_num>\d+)                            # E02 and separator
-     ([. _-]+s(?P=season_num)[. _-]*             # S01 and optional separator
-     e(?P<extra_ep_num>\d+))+                    # E03/etc and separator
+     [sS](?P<season_num>\d+)[. _-]*                 # S01 and optional separator
+     [eE](?P<ep_num>\d+)                            # E02 and separator
+     ([. _-]+[sS](?P=season_num)[. _-]*             # S01 and optional separator
+     [eE](?P<extra_ep_num>\d+))+                    # E03/etc and separator
      [. _-]*((?P<extra_info>.+?)                 # Source_Quality_Etc-
      ((?<![. _-])(?<!WEB)                        # Make sure this is really the release group
      -(?P<release_group>[^ -]+([. _-]\[.*\])?))?)?$              # Group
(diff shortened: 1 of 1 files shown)


BKSteve:
Trying to work through why already post processed files that have been renamed are no longer recognised when rescanned.
Their abs numbers have been converted to SxxExx and with this naming convention they aren't recognised.

miigotu:
This is unnecessary, we already compile them as case insensitive. (re.I is re.IGNORECASE)

https://github.com/SickChill/SickChill/blob/master/sickchill/oldbeard/name_parser/parser.py#L78

https://docs.python.org/3/library/re.html#re.I

miigotu:
You aren't supposed to use the post processor to add files that are already renamed by SickChill. You use "add existing show"

BKSteve:
Inside the show when `re-scan files` it runs through the post processor and gives `Unable to match` but it should right?

```
2022-08-21 14:43:40 INFO :: SHOWQUEUE-REFRESH :: Performing refresh on A Record of a Mortal's Journey to Immortality
2022-08-21 14:43:40 DEBUG :: SHOWQUEUE-REFRESH :: 384640: Loading all episodes from the show directory /home/stephenr/Videos/A Record of a Mortal's Journey to Immortality
2022-08-21 14:43:40 DEBUG :: SHOWQUEUE-REFRESH :: 384640: Creating episode from A.Record.of.a.Mortal's.Journey.to.Immortality.S03E09.Episode.55.Unknown.mkv
2022-08-21 14:43:40 DEBUG :: SHOWQUEUE-REFRESH :: 384640: Creating episode object from /home/stephenr/Videos/A Record of a Mortal's Journey to Immortality/A.Record.of.a.Mortal's.Journey.to.Immortality.S03E09.Episode.55.Unknown.mkv
2022-08-21 14:43:40 DEBUG :: SHOWQUEUE-REFRESH :: 384640: Unable to match /home/stephenr/Videos/A Record of a Mortal's Journey to Immortality/A.Record.of.a.Mortal's.Journey.to.Immortality.S03E09.Episode.55.Unknown.mkv to a show in your database. Parser result: A Record of a Mortal's Journey to Immortality - S03E09 [ABS: [0]] [ANIME VER: 1] [ABD: False] [ANIME: True] [whichReg: ['anime_and_normal']] Score: -4
```

miigotu:
That has nothing to do with case sensitivity though, it hasn't found that exact name in your show list. 

It might be the apostrophe. But nevertheless, that is merely a debugging line, so unless it isn't being matched and scanned in there shouldn't be an issue. Check your scene exceptions for that show.

BKSteve:
It isn't being matched and then sets the show to `archived` but the file is still there

miigotu:
Probably because the show is set to anime and then you don't have an absolute number there in the file name. (Episode 5 doesn't match as an absolute number, you need %AB in your naming rule iirc. Both SxxExx and absolute number can be there in the name) If you disable the anime option, save the show, and then rescan I bet it finds it.

It shouldn't be archiving that episode if it already knows that file belongs to an episode, but if it doesn't know it (you dropped it in there or renamed it yourself) then it's expected.


Anime needs a lot of work, but it needs done in a proper way. It has always been a non-priority for me, because I don't watch or care about anime at all personally to be honest. Always up for someone else to make it better though lol.

BKSteve:
I only have a few anime and note your comments on naming and will attempt to include absolute number in naming.
If I get somewhere I'll raise a new pull request.

Add: Might be more of a `guessit` fix.

PR #8117, opened 2022-08-27
SickChill with more rage removal

BKSteve:
Fixes the one last icon with the shark in Help & Info.

Removed some more sr and SR and SickRage items changing to sc SC and SickRage.

Probably over stepped the wildcard replace of SR with SC as changed init.

Hidden in there is alas a `Regex` change which i had done and left in by mistake, it's `s` or `e` replaced by `[sS]` or `[eE]`


BKSteve:
Left the .pot changes in so the `SR` is removed.
Removed some SickRage translations using crowdin

Add: A bunch of messages.json / .po files don't exist in crowdin. hr_Croatian, id_Indonesian, is_Icelandic and others.

contrib/runscripts/init.debian
```diff
@@ -56,7 +56,7 @@ RUN_AS=${SR_USER-sickchill}
 # Run as group
 RUN_GROUP=${SR_GROUP-sickchill}
 
-# Path to app SR_HOME=path_to_app_SickChill.py
+# Path to app SC_HOME=path_to_app_SickChill.py
 APP_PATH=${SR_HOME-/opt/sickchill}
```
miigotu: The ones inside the variable expansion are all missed in runscripts.

You'd have to check both variable names, by doing something like:
```bash
sc_home=${SC_HOME:-$SR_HOME}
APP_PATH=${sc_home:-/opt/sickchill}
```

miigotu:
If the messages.json isn't on crowdin, they can be deleted. Some languages never had any translation done so they were removed from our target languages.

The regex change has no effect, so I'd rather that taken out. We always check the regexes case insensitive, otherwise they would be even more convoluted.
`git revert 407c23731a47f5b5dedb95290ea13fd719610c39`

The contrib folder is the biggest concern here, because if someone has for example SR_OPTS in `/etc/systemd/system/sickchill.service.d/override.conf`
and then you change the variable name to SC_OPTS their install will not start correctly unless they change their override variable name manually.


You missed some spots in the contrib/runscripts also, see the inline comment.

contrib/runscripts/init.debian
miigotu: Possibly, we could also set some checks in SC that looks for variables like SR_HOME in the environment and logs a warning if they are found, with the notice that in X amount of time we will be removing support for SR_ variables (maybe 3 months?), With a link to a wiki page about how to fix the issue and/or locations of the files they are in, to give them time to correct the issue before we break their setup with arbitrary code changes there.

miigotu: Possibly even a command they can run on Linux or windows to fix it for them automatically.

Log:
```bash
sudo sed -Ri 's/SR_/SC_/gm' /etc/systemd/system/sickchill*
```

or whatever it is.

BKSteve: Yes, I've temporarily reverted as I'd only changed the comments and not the actual code.
Will work on this so we don't crash people out.

sickchill/oldbeard/databases/main.py
```diff
@@ -228,7 +228,7 @@ def execute(self):
...
+                "CREATE TABLE tv_shows(show_id INTEGER PRIMARY KEY, indexer_id NUMERIC, indexer NUMERIC, show_name TEXT, location TEXT, network TEXT, genre TEXT, classification TEXT, runtime NUMERIC, quality NUMERIC, airs TEXT, status TEXT, flatten_folders NUMERIC, paused NUMERIC, startyear NUMERIC, air_by_date NUMERIC, lang TEXT, subtitles NUMERIC, notify_list TEXT, imdb_id TEXT, last_update_indexer NUMERIC, dvdorder NUMERIC, archive_firstmatch NUMERIC, rls_require_words TEXT, rls_ignore_words TEXT, sports NUMERIC, anime NUMERIC, scene NUMERIC, default_ep_status NUMERIC DEFAULT -1, sub_use_sc_metadata NUMERIC DEFAULT 0);",
```
miigotu: This is why some of these changes have not been made, we can't just change the name of database columns without a migration.

miigotu (commented):
I'm going to have to go over this PR pretty hard, just a fair warning. Some things are still sr named in code because they will cause an issue by renaming them.

sickchill/oldbeard/databases/main.py
BKSteve: Got it. Thanks for teaching me. Best to pull this last one back.
I'll do a review of the others

BKSteve:
Don't need to go breaking things unnecessarily.

PR #8156, opened 2022-09-11
subprocess.run poetry

BKSteve:
Fixes a number of Synology DSM items.

With a sequence of steps to manually get the `SynoComm` package working at `develop`

Tested on Pycharm (debian 11) and my DSM 7 VM.


miigotu:
And breaks GitHub actions 😂

miigotu:
This code has been removed, so it is unnecessary to fix it now =P

PR #8205, opened 2022-09-24
Docker multistage build

miigotu:
Make sure sickchill is ran from the installed version rather than ran directly from the source, reduce size and build time of docker build, and add back arm/v6 and arm/v7 targets to the manifest

Export built wheels into build artifacts for easy download by users
Combine Dockerfile and Dockerfile.rpi

Fix groups in pyproject.toml


miigotu:
@pigsyn let's see how this one works. I did thos in a browser on my phone so hopefully I didn't have any typos. This way we have the shared libraries in the base image, but not the build essentials or dev headers or build tools. Then we copy out the venv into the final + base so we should have everything we need. We need to disable the updater in docker builds also, and require them to just pull a new image to update.

ghost:
Forgot to set `ENV POETRY_VIRTUALENVS_CREATE=true`
Problem remains the same, if you let `poetry install` create the venv, dependencies are installed in /.venv
```
 => [builder 5/6] RUN poetry install --no-interaction --no-ansi                                                                                                     62.4s
 => => # Creating virtualenv sickchill-Rxg8AEJs-py3.10 in /.venv
 => => # Updating dependencies
 => => # Resolving dependencies...
```

But the package is considered to be installed in /sickchill, we need to copy some files.

root@f066b754cce4:/# cat .venv/sickchill-Rxg8AEJs-py3.10/lib/python3.10/site-packages/sickchill.pth
/sickchill

libffi8 doesn't seem to be available with apt

`POETRY_VIRTUALENVS_IN_PROJECT=false` generates a random venv name i.e .venv/sickchill-`Rxg8AEJs`-py3.10/bin so it makes it harder to set PATH. 
This is to prevent overwriting of existing virtual environments

Everything can be solved like this:
```diff
diff --git a/Dockerfile b/Dockerfile
index da55adea3..e792597de 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,11 +9,12 @@ ENV PIP_NO_CACHE_DIR 1
 ENV PIP_FIND_LINKS=https://wheel-index.linuxserver.io/ubuntu/

 ENV POETRY_INSTALLER_PARALLEL=false
-ENV POETRY_VIRTUALENVS_CREATE=false
-ENV POETRY_VIRTUALENVS_IN_PROJECT=false
-ENV POETRY_VIRTUALENVS_PATH=$HOME/.venv
+ENV POETRY_VIRTUALENVS_CREATE=true
+ENV POETRY_VIRTUALENVS_IN_PROJECT=true
+ENV POETRY_VIRTUALENVS_PATH=/app/.venv
 ENV POETRY_CACHE_DIR=$HOME/.cache/pypoetry
 ENV POETRY_HOME=$HOME/.poetry
+ENV PATH=$POETRY_VIRTUALENVS_PATH/bin:$PATH

 # docker run -dit --user 1000:1000 --name sickchill --restart=always \
 # -v mount_point:/mount_point \
@@ -21,27 +22,25 @@ ENV POETRY_HOME=$HOME/.poetry
 # -v /etc/localtime:/etc/localtime:ro
 # -p 8080:8081 sickchill/sickchill

-RUN mkdir -m 777 -p /sickchill $HOME/.cache
-
 RUN sed -i -e's/ main/ main contrib non-free/gm' /etc/apt/sources.list
-RUN apt-get update -qq && apt-get install -yq libxml2 libxslt1.1 libffi7 libffi8 libssl1.1 libmediainfo0v5  mediainfo unrar && \
+RUN apt-get update -qq && apt-get install -yq curl libxml2 libxslt1.1 libffi7 libssl1.1 libmediainfo0v5  mediainfo unrar && \
 apt-get clean -yqq && rm -rf /var/lib/apt/lists/*

-FROM base as builder
+FROM base as builder
 RUN apt-get update -qq && apt-get install -yq build-essential cargo libxml2-dev libxslt1-dev libffi-dev libssl-dev libmediainfo-dev python3-dev && \
 apt-get clean -yqq && rm -rf /var/lib/apt/lists/*

 RUN pip3 install --upgrade --prefer-binary poetry pip wheel setuptools

-WORKDIR /sickchill
+WORKDIR /app
 COPY . .

 RUN poetry install --no-interaction --no-ansi
-WORKDIR $POETRY_VIRTUALENVS_PATH/local/bin
-RUN rm -rf /sickchill

 from base as final
 COPY --from=builder $POETRY_VIRTUALENVS_PATH $POETRY_VIRTUALENVS_PATH
+COPY --from=builder /app/sickchill /app/sickchill
+COPY --from=builder /app/SickChill.py /app/SickChill.py

 VOLUME /data /downloads /tv
 ```

**Comment by miigotu** [comment]
> Forgot to set `ENV POETRY_VIRTUALENVS_CREATE=true`
> Problem remains the same, if you let `poetry install` create the venv, dependencies are installed in /.venv
> ```
>  => [builder 5/6] RUN poetry install --no-interaction --no-ansi                                                                                                     62.4s
>  => => # Creating virtualenv sickchill-Rxg8AEJs-py3.10 in /.venv
>  => => # Updating dependencies
>  => => # Resolving dependencies...
> ```
> 
> But the package is considered to be installed in /sickchill, we need to copy some files.
> 
> root@f066b754cce4:/# cat .venv/sickchill-Rxg8AEJs-py3.10/lib/python3.10/site-packages/sickchill.pth
> /sickchill
> 
> libffi8 doesn't seem to be available with apt
> 
> `POETRY_VIRTUALENVS_IN_PROJECT`=false generate a random venv name i.e .venv/sickchill-`Rxg8AEJs`-py3.10/bin so it makes it harder to set PATH. 
> This is to prevent overwriting of existing virtual environments
> 
> Everything can be solved like this:
> ```diff
> diff --git a/Dockerfile b/Dockerfile
> index da55adea3..e792597de 100644
> --- a/Dockerfile
> +++ b/Dockerfile
> @@ -9,11 +9,12 @@ ENV PIP_NO_CACHE_DIR 1
>  ENV PIP_FIND_LINKS=https://wheel-index.linuxserver.io/ubuntu/
> 
>  ENV POETRY_INSTALLER_PARALLEL=false
> -ENV POETRY_VIRTUALENVS_CREATE=false
> -ENV POETRY_VIRTUALENVS_IN_PROJECT=false
> -ENV POETRY_VIRTUALENVS_PATH=$HOME/.venv
> +ENV POETRY_VIRTUALENVS_CREATE=true
> +ENV POETRY_VIRTUALENVS_IN_PROJECT=true
> +ENV POETRY_VIRTUALENVS_PATH=/app/.venv
>  ENV POETRY_CACHE_DIR=$HOME/.cache/pypoetry
>  ENV POETRY_HOME=$HOME/.poetry
> +ENV PATH=$POETRY_VIRTUALENVS_PATH/bin:$PATH
> 
>  # docker run -dit --user 1000:1000 --name sickchill --restart=always \
>  # -v mount_point:/mount_point \
> @@ -21,27 +22,25 @@ ENV POETRY_HOME=$HOME/.poetry
>  # -v /etc/localtime:/etc/localtime:ro
>  # -p 8080:8081 sickchill/sickchill
> 
> -RUN mkdir -m 777 -p /sickchill $HOME/.cache
> -
>  RUN sed -i -e's/ main/ main contrib non-free/gm' /etc/apt/sources.list
> -RUN apt-get update -qq && apt-get install -yq libxml2 libxslt1.1 libffi7 libffi8 libssl1.1 libmediainfo0v5  mediainfo unrar && \
> +RUN apt-get update -qq && apt-get install -yq curl libxml2 libxslt1.1 libffi7 libssl1.1 libmediainfo0v5  mediainfo unrar && \
>  apt-get clean -yqq && rm -rf /var/lib/apt/lists/*
> 
> -FROM base as builder
> +FROM base as builder
>  RUN apt-get update -qq && apt-get install -yq build-essential cargo libxml2-dev libxslt1-dev libffi-dev libssl-dev libmediainfo-dev python3-dev && \
>  apt-get clean -yqq && rm -rf /var/lib/apt/lists/*
> 
>  RUN pip3 install --upgrade --prefer-binary poetry pip wheel setuptools
> 
> -WORKDIR /sickchill
> +WORKDIR /app
>  COPY . .
> 
>  RUN poetry install --no-interaction --no-ansi
> -WORKDIR $POETRY_VIRTUALENVS_PATH/local/bin
> -RUN rm -rf /sickchill
> 
>  from base as final
>  COPY --from=builder $POETRY_VIRTUALENVS_PATH $POETRY_VIRTUALENVS_PATH
> +COPY --from=builder /app/sickchill /app/sickchill
> +COPY --from=builder /app/SickChill.py /app/SickChill.py
> 
>  VOLUME /data /downloads /tv
>  ```

The problem now is libffi8, I'll figure it out.

**Comment by ghost** [comment]
python3-dev installs python3.9 on builder, even if you remove it, libpython3.9-dev is installed for other pkg dependencies

**Comment by miigotu** [comment]
I have a commit here on my PC that I forgot to push haha, I was sidetracked trying to find the libffi8 issue. I think we need to use the 3.9-slim, or move back to 3.8-slim. I'll make that quick change for now and rebase my commit later (Renaissance fair today so I am a little busy for a few hours, was a demolition derby yesterday too much happening)

**Comment by miigotu** [comment]
If we can't get it to build, we can switch the base to debian sid or Ubuntu lts which has libffi8 and python3.10, probably with a newer version of cargo/rust by default.

That's probably going to be the best thing to do anyway, so we can use python3.10 by default and have a more transparent build base.

File: Dockerfile [hunk 1]
```diff
@@ -1,49 +1,48 @@
-FROM --platform=$BUILDPLATFORM python:3.10-slim
+FROM --platform=$TARGETPLATFORM python:3.9-slim as base
 
 LABEL org.opencontainers.image.source="https://github.com/sickchill/sickchill"
 LABEL maintainer="miigotu@gmail.com"
 ENV PYTHONIOENCODING="UTF-8"
+ENV PYTHONBUFFERED 1
+
+ENV PIP_NO_CACHE_DIR 1
```
miigotu: This might actually force us to have to rebuild the wheels in extract-wheels, where if we keep the pip cache they should already be there from being built in build.

miigotu: Nevermind, this folder is lost in the build stage anyway, unless we copy it to final.

ghost: Is it possible to only use rustup script for the builder stage ? They say toolchain is not required to run cryptography.
https://cryptography.io/en/latest/installation/#rust
Then there's openssl, all those dependencies are a bit 😩

miigotu: The rustup script isn't in there anymore, I installed rust and cargo from the package manager in the builder stage instead.
Problem is, I don't think rustc works for x32

Dockerfile
```diff
@@ -1,49 +1,50 @@
...
 
-RUN . $HOME/.cargo/env && poetry install --no-interaction --no-ansi
+from base as final
+COPY --from=builder $POETRY_VIRTUALENVS_PATH $POETRY_VIRTUALENVS_PATH
+COPY --from=builder /app/sickchill /app/sickchill
+COPY --from=builder /app/SickChill.py /app/SickChill.py
 
-RUN rm -rf /app
-WORKDIR /
+VOLUME /data /downloads /tv
```
miigotu, line 58: I don't want to keep the sickchill dir, we don't need the source code.

miigotu:
Going to merge this for now, we can open a different PR with more changes if needed.

Dockerfile
ghost: SickChill is not installed, only dependencies and entrypoint:
```
root@581b3f14ad0d:~# cat /.venv/sickchill-Rxg8AEJs-py3.9/lib/python3.9/site-packages/sickchill.pth
/sickchill
root@581b3f14ad0d:~# ls -lah /sickchill/
total 8.0K
drwxrwxrwx 2 root root 4.0K Sep 26 12:28 .
drwxr-xr-x 1 root root 4.0K Sep 26 12:30 ..

root@5292b6ac91dc:~# ls -lah /.venv/local/bin
total 8.0K
drwxr-xr-x 2 root root 4.0K Sep 26 12:30 .
drwxr-xr-x 3 root root 4.0K Sep 26 12:30 ..
```

https://github.com/python-poetry/poetry/issues/34#issuecomment-1193365526

PR #8211, opened 2022-09-26
Fix dockerfile

ghost:
Fixes # docker images are broken (master, develop)

-
-
-

### For the main Dockerfile:
`poetry install --with root` installs sickchill as editable https://github.com/python-poetry/poetry/issues/34#issuecomment-1193365526
So you **need** to copy sickhill and SIckChill.py to have a working image, else you end up with dependencies, the SickChill entrypoint and this path file:
```
cat /.venv/`sickchill-Rxg8AEJs-py3.9`/lib/python3.9/site-packages/sickchill.pth
/sickchill
```
Note that I only copy /sickchill/sickchill module and /sickchill/SickChill.py, not the whole git repo
### On build variables
```bash
ENV POETRY_VIRTUALENVS_CREATE=true
ENV POETRY_VIRTUALENVS_IN_PROJECT=true
ENV POETRY_VIRTUALENVS_PATH=/sickchill/.venv
```
`POETRY_VIRTUALENVS_IN_PROJECT=true` makes POETRY_VIRTUALENVS_PATH irrelevant, it's always installed inside the project, still using it to copy and set PATH, could use a better name.

**Why do I do this ?**

When it's false and `POETRY_VIRTUALENVS_PATH` is set, poetry will **always** create a random named virtual environment:
/.venv/`sickchill-Rxg8AEJs-py3.9`/bin/
This is very inconvenient to update PATH.

All $HOME variables are non functional and resolved to nothing when building an image while not using USER instruction.
`$POETRY_VIRTUALENVS_PATH/local/bin`  Always empty so I removed it, not sure what you are trying to achieve 🤔 

### On docker pip installs
As you don't like the sickchill source copy I have included Dockerfile.with_pip as an example of what needs to be done do install dependencies with poetry and sickchill as a package. 
(venv creation, poetry install, poetry build and pip install dist/sickchill-*-py3-none-any.whl)
Version could be retrieved from `poetry version -s`
But Dockerfile.pip from contrib is almost the same.

Also fixed Dockerfile.pip from contrib as there was no sourcing of virtual environment and packages were installed in /usr/local/lib I believe

### TODO: Add a user and drop privileges, preferablty from --user argument
I could slowly update Dockerfiles for this, it may have implications in workflows, or user running docker as root , or not using user: "UID:GUID" in their docker compose files

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,8 @@ ENV PYTHONBUFFERED 1
 
 ENV POETRY_INSTALLER_PARALLEL=false
 ENV POETRY_VIRTUALENVS_CREATE=true
-ENV POETRY_VIRTUALENVS_IN_PROJECT=false
-ENV POETRY_VIRTUALENVS_PATH=$HOME/.venv
+ENV POETRY_VIRTUALENVS_IN_PROJECT=true
+ENV POETRY_VIRTUALENVS_PATH=/sickchill/.venv
 ENV POETRY_CACHE_DIR=$HOME/.cache/pypoetry
 ENV POETRY_HOME=$HOME/.poetry
 
@@ -43,15 +43,17 @@ RUN pip3 install --upgrade --prefer-binary poetry pip wheel setuptools
 WORKDIR /sickchill
 COPY . /sickchill/
 
-RUN poetry run python3 -m pip  install --upgrade cffi pynacl greenlet pyopenssl cryptography --force-reinstall
+RUN poetry run python3 -m pip install --upgrade cffi pynacl greenlet pyopenssl cryptography --force-reinstall
 RUN poetry install --no-interaction --no-ansi --without dev --with speedups --with root --without experimental --without types
 
 from base as sickchill-final
 
 COPY --from=builder $POETRY_VIRTUALENVS_PATH $POETRY_VIRTUALENVS_PATH
+COPY --from=builder /sickchill/sickchill /sickchill/sickchill
+COPY --from=builder /sickchill/SickChill.py /sickchill/SickChill.py
 
-ENV PATH=$POETRY_VIRTUALENVS_PATH/bin:$POETRY_VIRTUALENVS_PATH/local/bin:$PATH
-WORKDIR $POETRY_VIRTUALENVS_PATH/local/bin
+ENV PATH=$POETRY_VIRTUALENVS_PATH/bin:$PATH
+WORKDIR $POETRY_VIRTUALENVS_PATH/bin
 
 VOLUME /data /downloads /tv
 
--- a/Dockerfile.with_pip
+++ b/Dockerfile.with_pip
@@ -1,0 +1,65 @@
+FROM --platform=$TARGETPLATFORM python:3.9-slim as base
+
+LABEL org.opencontainers.image.source="https://github.com/sickchill/sickchill"
+LABEL maintainer="miigotu@gmail.com"
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PYTHONIOENCODING="UTF-8"
+ENV PYTHONBUFFERED 1
+
+#ENV PIP_EXTRA_INDEX_URL=https://www.piwheels.org/simple
+#ENV PIP_FIND_LINKS=https://wheel-index.linuxserver.io/ubuntu/
+
+ENV POETRY_INSTALLER_PARALLEL=false
+ENV POETRY_VIRTUALENVS_CREATE=false
+ENV POETRY_VIRTUALENVS_IN_PROJECT=false
+ENV POETRY_CACHE_DIR=$HOME/.cache/pypoetry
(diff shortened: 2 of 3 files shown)


miigotu:
The source and SickChill.py is not used at all once it's installed.

We don't need poetry to create the venv at all. I'll make the update later, I don't want to keep going backwards when I'm trying to accomplish something with it.

ghost:
There is still an issue if you are running docker as non root or using user "UID:GUID" in docker compose

ghost:
> The source and SickChill.py is not used at all once it's installed.

It is by the entrypoint

```
import sys
from SickChill import main

if __name__ == '__main__':
    sys.exit(main())
```

miigotu:
In python 3.10 the binary is in . venv/local/bin. In everything else it's in . venv/bin. Leave it that way.

$HOME is empty, when you aren't using --user or useradd, but it is not when you use --user.

poetry build makes more sense than copying the source files directly.

The literal only change that makes sense is using virtualenv_in_project. It's not about making it how you would do it on your system, it's about it working on all of them.

Gotta stop changing what I'm changing. I'm doing it for a reason and I wasn't done, I just didn't want the PR to keep being so big.

miigotu:
> > The source and SickChill.py is not used at all once it's installed.
> 
> It is by the entrypoint
> 
> ```
> import sys
> from SickChill import main
> 
> if __name__ == '__main__':
>     sys.exit(main())
> ```

Yes, from site-packages in the virtualenv. Not the local dir.

I didn't get it finished yet how I want it, because I'm trying to get arm/v6 and arm/v7 to build first.

ghost:
> In python 3.10 the binary is in . venv/local/bin. In everything else it's in . venv/bin. Leave it that way.

I didn't noticed it in python:3.10-slim or ubuntu 22.04 with py3.10, but yeah it could be different on other systems. 
Sometimes it's hard to read what is the intention, but want to help ❤️ 

> I didn't get it finished yet how I want it, because I'm trying to get arm/v6 and arm/v7 to build first.

So sad you pushed master image 😢, I can always test or better could be done with some workflow with a curl test.
Maybe  version tags can be good for people that want to roll back !?

This PR was just to provide working images until everything can be sorted out. 
Current master still doesn't work because of wrong `POETRY_VIRTIALENVS_PATH` and editable poetry install.

Good thing is I have a solution for running as non root user with docker --user ID:GID or with docker compose, will wait for a clean image setup to propose it.

miigotu:
> > In python 3.10 the binary is in . venv/local/bin. In everything else it's in . venv/bin. Leave it that way.
> 
> I didn't noticed it in python:3.10-slim or ubuntu 22.04 with py3.10, but yeah it could be different on other systems. Sometimes it's hard to read what is the intention, but want to help ❤️
> 
> > I didn't get it finished yet how I want it, because I'm trying to get arm/v6 and arm/v7 to build first.
> 
> So sad you pushed master image 😢, I can always test or better could be done with some workflow with a curl test. Maybe version tags can be good for people that want to roll back !?
> 
> This PR was just to provide working images until everything can be sorted out. Current master still doesn't work because of wrong `POETRY_VIRTIALENVS_PATH` and editable poetry install.
> 
> Good thing is I have a solution for running as non root user with docker --user ID:GID or with docker compose, will wait for a clean image setup to propose it.

--user and --group already works, as does `user: 1001:1001` in compose or setting UID and GID environment variables in docker-compose env for the service.

The only thing that doesn't work right now is if you start it without those variables it is created by the root user.

miigotu:
> > In python 3.10 the binary is in . venv/local/bin. In everything else it's in . venv/bin. Leave it that way.
> 
> I didn't noticed it in python:3.10-slim or ubuntu 22.04 with py3.10, but yeah it could be different on other systems. Sometimes it's hard to read what is the intention, but want to help ❤️
> 
> > I didn't get it finished yet how I want it, because I'm trying to get arm/v6 and arm/v7 to build first.
> 
> So sad you pushed master image 😢, I can always test or better could be done with some workflow with a curl test. Maybe version tags can be good for people that want to roll back !?
> 
> This PR was just to provide working images until everything can be sorted out. Current master still doesn't work because of wrong `POETRY_VIRTIALENVS_PATH` and editable poetry install.
> 
> Good thing is I have a solution for running as non root user with docker --user ID:GID or with docker compose, will wait for a clean image setup to propose it.

With a docker image, switching to the previous tag is literally changing one 6 letter string and pulling. It's not that big of a deal.

miigotu:
And actually, it seems the versioned tags stopped getting tagged 5 months ago, I'll fix that later. I have family shit to do right now.

miigotu:
And the reason it got merged to master is because there were fixes that needed to be put out because some people couldn't start SC. The number of people using a pip install is like 20x how many people use docker for it, so it's either leave it broken for them until I get docker completely ironed out, or break docker for a day (when most people aren't using pull or watchtowerr on docker anyway, they usually use the updater even though that won't be an option soon.)

PR #8212, opened 2022-09-26
Combine Dockerfile, Dockerfile.wheels, and Dockerfile.pip, create our own venv

miigotu:


kyuuk:
Hey @miigotu,

i'm commenting here to avoid opening a new issue,

i saw that the docker docker images build fails for linux/arm64, and i found this issue https://github.com/python-poetry/poetry/issues/5131 which seems to be related, should we go back to installing rust ?

miigotu:
> Hey @miigotu,
> 
> i'm commenting here to avoid opening a new issue,
> 
> i saw that the docker docker images build fails for linux/arm64, and i found this issue https://github.com/python-poetry/poetry/issues/5131 which seems to be related, should we go back to installing rust ?

It doesn't fail for arm64. It fails for 32bit architectures (armv6 and armv7), because they don't pre build wheels for them and there is an upstream qemu big that prevents them from building on GitHub actions (docker in docker).

I have built the wheels locally and then turn around and push the same code and environment to GitHub and it fails. It's been doing it for 3 years now, we just were able to work around it for a long time. This is the same reason people on Synology have no wheels currently and cant run SC on non-amd64/arm64. 

Rust doesn't even release an armv6 compiler. 

I've been working on getting something together for it, but since it works locally and not remotely I have to let the build happen each time in between changes to see if it fixes anything. Its taking over 2 hours per build right now with my latest tests, because I'm building python from source to see if it's something they did in the GitHub python packages. 

https://github.com/miigotu/python-docker-images/actions/runs/3164329961/jobs/5152551055

I've been doing it a ton of different ways and nobody anywhere has gotten it to work on GitHub yet, or I would have copied what they are doing already.

https://github.com/miigotu/python-docker-images/tree/main/.github/workflows

There's some of the workflows, I even cross compiled on x64 and then copied the cargo files over to the target image and that fails too.

miigotu:
But by all means, if someone can get proper wheels for cffi, pynacl, cryptography, msgfmt, wrapt, lxml, and sqlalchemy to build for arm32, ppc64le, and s390x on GitHub actions please make a pull request.

I suggest making a fork or another repo and getting it working on your fork first though, because you won't see it failing here.

kyuuk:
yes i saw that armv6/7 does not want to work despite all you tried, i also tried looking for that but i still don't know enough python to to really understand all this,

but what i'm refering to is that develop build are not passing right now and always fail here https://github.com/SickChill/sickchill/actions/runs/3163965294/jobs/5151980987#step:11:3279 when installing poetry

Also, i just thought of something, would using a self-hosted  runner on a raspberrypi for example be a workaround for arm build ?

miigotu:
The poetry error is a different issues, that has nothing to do with architecture. There's a bug in the 1.2.1 poetry installer.

miigotu:
It's being worked out

miigotu:
https://github.com/python-poetry/install.python-poetry.org/issues/37

PR #8218, opened 2022-09-30
Fix rss feeds

### LINKED ISSUE #8217: RSS Feeds are broken since tvcache.py was refactored

Version: 2022.9.28
OS: Linux
Browser: Chrome
What you did: Upgraded to new version of SickChill.  It had been a while.
What happened: RSS Feeds stopped working.
What you expected: RSS Feeds should continue working.
Logs:
```
2022-09-29 14:53:44 DEBUG :: SEARCHQUEUE-DAILY-SEARCH :: [custom] :: Did not find size|seeders|leechers: None, 0, 0
```

The above message shows up for pretty much everything in the feed.... I went digging and found that https://github.com/SickChill/sickchill/pull/8124 had made a subtle change to how item_size gets computed.  The old code essentially always defaulted to -1 (even though that's not obvious).

wyrickre:
Fixes #8217 

- Default item_size to -1


wyrickre:
Whoops... my mistake.

PR #8219, opened 2022-09-30
Default item_size so torrent RSS feeds work again

### LINKED ISSUE #8217: RSS Feeds are broken since tvcache.py was refactored

wyrickre:
Fixes #8217

- Make item_size default to -1

### DIFF
--- a/sickchill/oldbeard/tvcache.py
+++ b/sickchill/oldbeard/tvcache.py
@@ -60,7 +60,7 @@ class RSSTorrentMixin:
             logger.debug(f"{item}")
             return
 
-        item_size = None
+        item_size = -1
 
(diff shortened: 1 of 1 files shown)


miigotu:
Does your feed not have size information at all? because that is not technically a valid feed.

miigotu:
Nevermind, I see showrss doesnt have them. It's not commonly used but this is fine for now. I want to actually make this a setting someday.

PR #8221, opened 2022-10-01
Enable 32-bit ARM v7

offbyone:
This will let me pull the image on my Raspberry Pi

- enable arm/v7 in the docker build

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -210,7 +210,7 @@ jobs:
         context: .
         # disable 32 bit arm builds for now, just get a working image out. 
         # platforms: linux/amd64,linux/arm64,linux/arm/v6,linux/arm/v7
-        platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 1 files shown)


miigotu:
This won't work. I've been working on arm6 and arm7 images for the past 2 weeks (and many times over the past 3 years).

Cryptography is not built for 32 bit arm, which is required by poetry and other tools, along with SickChill.

The docker build will fail every time with armv7, because of an upstream bug in qemu. I found a hacky way around it last night, but it's going to take more time. When that's done, arm6, arm7, s390x, mips64le, and ppc64le images will be added back.

Closing this, because it absolutely will fault to build (check the actions results)

miigotu:
For example, this build and compilation works locally, but not on GitHub

https://github.com/miigotu/python-docker-images/actions/runs/3157390999/jobs/5138207414

You can check out GitHub source and build it locally with no issues for armv7, because you won't be docker in docker in docker like the actions are.

Issue #7694, opened 2021-11-17
synodsmnotify not working in sickchill with DSM7


spock99:
Describe the bug
No notification sent to host system when download shows

To Reproduce
Download any show after turning synology notifier on

Expected behavior
I would expect it to sent notification anytime it grab file none seen
Work fine in DSM6.x Know change to synodsmnotify in DSM7

Screenshots
Nothing to show

SickChill Info:
Branch: master
Commit: 01020f3636d01535f60b83464d8127ea0efabfc7
Database Version: 44.5

  Python Version:
3.8.12 (default, Oct 22 2021, 21:05:38) [GCC 7.5.0]

  SSL Version:
OpenSSL 1.1.1l 24 Aug 2021

  OS:
Linux-4.4.180+-x86_64-with-glibc2.2.5

Additional context
Is there any way to test and trace this?
DS918+
synodsmnotify was changed in DSM7

Below is example from another project that was looking at this.
https://community.synology.com/enu/forum/36/post/140600

Failure in Sickchill is same issue the related to synodsmnotify and i18n under DSM-7
This command can also be run from command-line 

Hope this helps

spock99:
Be aware this not related to mail. It is Synology notification this sent to the web interface. Reproduction can be done on cli on 6.x but 7.x changed text format so need to support new format.

BKSteve:
This is a pain in the A.. i18n no ability to seemingly input text directly. Everything pre assembled.
[https://global.download.synology.com/download/Document/Software/DeveloperGuide/Firmware/DSM/7.0/enu/DSM_Developer_Guide_7_0_Beta.pdf](https://global.download.synology.com/download/Document/Software/DeveloperGuide/Firmware/DSM/7.0/enu/DSM_Developer_Guide_7_0_Beta.pdf)
```
Application InternationalizationThe desktop application can have i18n text referenced by config, help, etc.
ui (specified by dsmuidir in INFO)
└── texts   
├── enu   
│    └── strings
└── cht
     └── strings
You have to create directories according to supported languages then create a file named strings inside each language directory.
[dsmuidir]/texts/enu/strings
    [app_tree]
    index_title="This is a title"  
    node_1="This is node1"  
    [app_tab]  
    tab1="This is tab1"  
    tab2="This is tab2"

[dsmuidir] /texts/cht/strings
    [app_tree]  
    index_title=""  
    node_1="1"  
    [app_tab]  
    tab1="1"  
    tab2="2"
    
[...]

PR #8271, opened 2022-10-22
Feature: notify on logged errors

### LINKED ISSUE #8270: [feature] Enable push notifications on logged errors
I would like to receive push notifications when there is an error logged in SickChill, at least the first time the error appears in the log. I generally try not to look at the system too much when it's working, but if I experience authentication failures to a search provider, I definitely want early notification of that.

How much work would it be to wire in the notification system to the error log filter? I'm happy to look into doing this myself, too, if that's the best way to achieve it.

offbyone:
Fixes #8270 and #8034

* Added a setting to the advanced settings tab: "Notify on Errors"
* Added a call to `notifiers.notify_logged_error` when seeing an error message in the log.
* Implemented `notify_logged_error` for Synology and for Discord, as bare bones initial implementations


offbyone:
I didn't do the I18N part; I'm not actually sure how to update the `.po` files, but I'm happy to add that commit in here as well.

I'm actually also not even a bit sure how to test this; I don't know how to intentionally trigger an error from my development instance.

miigotu:
Yeah don't worry about regenerating the po/mo, all you need to to is mark the strings with _()

BKSteve:
The DSM 7 Synology notifications system is fubar. They killed all external access to their internal notification system. 
#7694 

ADD: Maybe Syno Notifier should be dropped?

offbyone:
I more or less just grabbed that system out of a hat, because I don't use it either; I think if it's dead, sure, it can be pulled in a separate PR.

offbyone:
I tested it by generating some errors; I think the formatting needs some work, but the plumbing is there.

offbyone:
Is this merged? I can't tell from the force push at the end.

miigotu:
No, I just merged the develop branch back into this so that is doesn't get out of sync and create a conflict. Are you ready for it to be merged yet?

sickchill/oldbeard/notifiers/discord.py
```diff
@@ -27,6 +28,13 @@ def notify_login(self, ipaddress=""):
             title = common.notifyStrings[common.NOTIFY_LOGIN]
             self._notify_discord(title + " - " + update_text.format(ipaddress))
 
+    def notify_logged_error(self, ui_error: UIError):
+        if settings.USE_DISCORD:
+            update_text = ui_error.message
+            title = ui_error.title
+
+            self._notify_discord(f"{title} - {update_text}")
```
miigotu, line 36: It would be cool if we had a method to pastebin the error log or the debug log when there is an error, and send it to a special channel on our discord server. We could have this as an optional feature, defaulting to disabled, where the options are to automatically send them on errors, or on a manual click on a button in the error log viewer. There should be throttling and duplicate/spam prevention perhaps with storing the last date a paste was created and sent and not sending errors from before that date again, and somehow preventing the same error for each of a thousand shows  from being sent one after the other, and limit each installation to only sending errors logs once a day to discord. In the issue_submitter (or in logger.py) there is a method that sanitizes and gathers the log lines to be sent already, which we use for automatically submitting big reports, but having them all go to one channel on discord would make it easy to see bugs quicker, even before someone writes an issue on GitHub.

Preferably the user should have their GitHub credentials stored in SC, and a private gist could be created, and otherwise use a public pastebin site that has a private option and ability to paste anonymously would be a fallback. 
What do you think?

offbyone: I think it's a fantastic idea, although if I understand you correctly, that's a separate discord notification; the one I've got here is for notifying the _owner's_ discord, whereas you want a pipeline to notify your discord as a rapid bug report.

I probably wouldn't tie them together, tbh.

offbyone:
> No, I just merged the develop branch back into this so that is doesn't get out of sync and create a conflict. Are you ready for it to be merged yet?

Yep, I'm good to merge this.

miigotu:
Thanks for this PR. I'm going to merge it into develop now.

PR #8281, opened 2022-10-30
Poster sort Next Episode

### LINKED ISSUE #8275: Poster order does not match sort criteria
Version: 2022.10.13 (docker)
OS: Linux
Browser: Firefox
What you did: Sort posters by 'next episode'
What happened: Dates from the tiles do not match the order
What you expected: Posters sorted by dates

image
image
image
I just noticed that ironically this paste, it actually was correct, but more then not this is not the case, see the second (which is my second) line as well.

I also realize, that half comes from bad data, but some shows really are weird. Some shows use the final expected episode date for sorting, some the last aired time, it's really quite messy ...

BKSteve:
Fixes #8275

- add `Prev Ep:` to shows without future `Next Episode` dates, which goes in front of `Previous Episode` date.
- not a fix as not needing a fix, just needing clarity on what's displayed.

image
Those with `Prev Ep` are sorted alphabetically by default.


BKSteve:
Added a few more icons to Readme.md

miigotu:
Have to set the width and height on all of those logos to the same size. 

I figure 50x50 is a good size for them all like the single one was before, but they definitely need to all be the same size.

BKSteve:
The main jetbrains is so small otherwise. gone with 60x60

miigotu:
Maybe we could do this a little differently, by putting the prev episode date in a different div, that way if there is no next ep the next ep div is empty instead of sorting picking up the div with the prev date?

I'd like to check the branch out and see the problem myself and see if we can change how we lay it out to make it even clearer than this.

By separating them into different divs, we could sort all by previous date and then by next date and it would probably look correct/logical in sorting? What about options to hide one with no next episode? 

Really cool you finally found the cause of this issue report, it has come up many times and I have never been able to reproduce it.

BKSteve:
I didn't want to paste images with all the shows I'm pulling so cut it to just one line on my test machine.

The code combines `next` with `prev` for display but really this could just be `next` only dates. I just mod it and inserted text `Prev Ep:` to clarify the date, the other choice was to dump `prev` ~~totally but that will probably throw an error. If it doesn't great, if it does  then put `Unknown` or `Next Unknown` ??~~

BKSteve:
image
sorts continuing with next ep, then continuing no next ep, then ended.
I like this better than earlier, cleaner, simple, easy to explain in wiki. :smirk: 

Old way
image

PR #8302, opened 2022-11-26
265 and mass update

BKSteve:
Always allow 265 hevc to be collected.

Add word functions to mass update - edit selected (pending)

#7333


sickchill/oldbeard/providers/newznab.py
```diff
@@ -17,7 +17,7 @@ class NewznabProvider(NZBProvider, tvcache.RSSTorrentMixin):
     Tested with: newznab, nzedb, spotweb, torznab
     """
 
-    def __init__(self, name, url, key="0", catIDs="5030,5040", search_mode="eponly", search_fallback=False, enable_daily=True, enable_backlog=False):
+    def __init__(self, name, url, key="0", catIDs="5030,5040,5090", search_mode="eponly", search_fallback=False, enable_daily=True, enable_backlog=False):
```
miigotu: This can't be changed globally like this, as many indexers do not know this category. It will cause some to fail finding anything at all. https://nzbplanet.net/api?t=caps

sickchill/oldbeard/providers/newznab.py
```diff
@@ -88,9 +88,6 @@ def providers_list(data):
                     ",".join([x for x in providers_dict[default.name].catIDs.split(",") if 5000 <= try_int(x) <= 5999]) or default.catIDs
                 )
 
-                if settings.QUALITY_ALLOW_HEVC and "5090" not in providers_dict[default.name].catIDs:
```
miigotu, line 91: This probably wasn't a great idea when I did this. When we check caps in the initialization of the provider, we can check if the provider defined 5090 and add it there if necessary.

miigotu:
This looks good for the first part of the change, maybe we should merge this as is since it looks done for removing the setting. Then another PR to add the feature on mass update for required/ignored words.

BKSteve:
> This looks good for the first part of the change, maybe we should merge this as is since it looks done for removing the setting. Then another PR to add the feature on mass update for required/ignored words.

Fine by me, I'll create a new one for the words

sickchill/oldbeard/providers/newznab.py
BKSteve: so 5045 ?
`<subcat id="5045" name="UHD"/>`

BKSteve: removed 5090

sickchill/oldbeard/providers/newznab.py
BKSteve: Never used NZBs so with the 2 items did some reading and checking. Users can select 5090 or 5045 depending on their `Provider settings`

miigotu: Yes, I know that. Which is why we shouldn't arbitrarily add 5090. We should check for 5045 or 5090 in the categories response when we get server caps and add the right one to the provider automatically.

PR #8323, opened 2022-12-29
Add ability to save fansub Black/White list as defaults

kyuuk:
Fixes #7992

- add ability to save fansub groups (Black/White list) to default settings when adding a new show

---


sickchill/gui/slick/views/inc_addShowOptions.mako
```diff
@@ -106,10 +114,3 @@
             <label>${_('Use current values as the defaults')}</label>
```
miigotu, line 114: This $ has to stay to call the python _ function, right? Or is it using the js function instead?  I'll look over it more later

sickchill/gui/slick/js/core.js
```diff
@@ -286,8 +286,8 @@ const SICKCHILL = {
             },
         },
         updateBlackWhiteList(showName) {
-            $('#white').children().remove();
-            $('#black').children().remove();
+            //$('#white').children().remove();
```
miigotu: Have to put a space after the //, or use `poe format` to automatically fix it.

sickchill/gui/slick/js/core.js
```diff
@@ -4077,7 +4077,7 @@ const SICKCHILL = {
                 $('#bestQualities option:selected').each((i, d) => {
                     bestQualArray.push($(d).val());
                 });
-
+                generateBlackWhiteList();
```
miigotu: This function is undefined in testing.

sickchill/gui/slick/views/inc_addShowOptions.mako
kyuuk: weird enough it's still there (i just moved that block after B/W list), but i don't know why it is in red https://github.com/SickChill/sickchill/blob/0642fac6d74b8fa0421dcc24ee22bea35568b99d/sickchill/gui/slick/views/inc_addShowOptions.mako#L114

but from what i saw the call is made in JS

sickchill/gui/slick/js/core.js
kyuuk: i totally removed these 2 lines

sickchill/gui/slick/js/core.js
kyuuk: should be fixed with my last (force) push 
the function is defined in [sickchill/gui/slick/js/blackwhite.js](https://github.com/SickChill/sickchill/pull/8323/files#diff-aa1b2c93e31861be74f6641df2308beb98c368856b8d8aeb88d1405525bbd82f) i added an es-lint comment like in other usage of the function in core.js

sickchill/gui/slick/views/inc_addShowOptions.mako
miigotu: When the $ is there {} is a function to call python code. When it is {} without it it would show that whole call in the html, including the {}

sickchill/gui/slick/js/core.js
```diff
@@ -286,8 +286,6 @@ const SICKCHILL = {
             },
         },
         updateBlackWhiteList(showName) {
-            $('#white').children().remove();
-            $('#black').children().remove();
```
miigotu, line 290: This is comical, apparently the code was removing all of the white and black children, apparently mostly leaving the Asian and Native children.

sickchill/gui/slick/views/inc_addShowOptions.mako
```diff
@@ -96,6 +96,14 @@
         </div>
     </div>
     <br>
+
+    % if enable_anime_options:
+        <% import sickchill.oldbeard.blackandwhitelist %>
```
miigotu, line 101: I would prefer to always import the python package, and place the imports at the top in the imports block. This way, if there is an error in the python for blackandwhitelist it should throw for me even in my testing since I almost never have anime active. The include can stay down here obviously.

sickchill/views/config/general.py
```diff
@@ -34,7 +34,16 @@ def saveRootDirs(rootDirString=None):
 
     @staticmethod
     def saveAddShowDefaults(
-        defaultStatus, anyQualities, bestQualities, defaultSeasonFolders, subtitles=False, anime=False, scene=False, defaultStatusAfter=WANTED
```
miigotu, line 37: This doesn't need done in this PR, but I just had the thought that all of the web route methods should only have *args, **kwargs as parameters, and when we need the parameter self.get_arguments, self.get_post_arguments, and self.get_query_arguments should be used, and in the template render function at the end we should always just pass **locals() instead of passing all of the local variables verbosely. This way I could work towards fixing the router, which would let us use mako's router instead of tornado eventually. That means people could use wsgi or all kinds of different web servers to serve sickchill directly instead of through a python server.

miigotu:
! approve

sickchill/gui/slick/views/inc_addShowOptions.mako
miigotu, line 101: I would prefer to always import the python package, and place the imports at the top in the imports block. This way, if there is an error in the python for blackandwhitelist it should throw for me even in my testing since I almost never have anime active. The include can stay down here obviously.

miigotu (approved):

sickchill/views/config/general.py
kyuuk: i agree with this, it would lessen the burden of adding/removing params, i could try doing that once i'm move familiar with all the dev environment (and that my little girl sleeps all night 🤣 )

sickchill/gui/slick/views/inc_addShowOptions.mako
kyuuk: Don't worry, you can count on me to test it, i'm running on develop and i use SC mainly for animes, so if something is broken, i will know 😆

kyuuk:
@miigotu  I let you merge if everything is good for you

sickchill/gui/slick/views/inc_addShowOptions.mako
miigotu: I mean, for it to be future proof, that it is easy for future developers to find and fix an issue for example 25 years from now ;)

PR #8328, opened 2023-01-02
XML headers now write to NFO

ethoscraig:
Fixes #8325 

- Added xml.write parameter "xml_declaration=True", this ensures an XML header is written to NFO files.


miigotu:
Awesome, thank you.

Issue #8365, opened 2023-02-19
NZB splitting of season packs does not work


Rubio9:
### Before submitting your issue:

Enable debug logging in SickChill settings, reproduce the error (be sure to disable after the bug is fixed)

Version: 2023.1.2
OS: Linux / Synology DSM7
Browser: Any
What you did: Force backlog search for a show where there is a season pack with missing episodes.
What happened: Logs show it finds the season pack and decides it will use it, but never does. Debug logs show it errors.
What you expected: Season pack's nzb to be split so the wanted episodes out of it are downloaded.
Logs:
```
2023-02-17 13:39:51 DEBUG :: SEARCHQUEUE-BACKLOG-xxxxx :: [nzb.su] :: Episode already exists with quality SDTV but the found result quality 1080p WEB-DL is wanted more, getting found result for xxxxx S01E06
2023-02-17 13:39:51 DEBUG :: SEARCHQUEUE-BACKLOG-xxxxx :: [nzb.su] :: Breaking apart the NZB and adding the individual ones to our results
2023-02-17 13:39:52 DEBUG :: SEARCHQUEUE-BACKLOG-xxxxx :: [nzb.su] :: Using cached parse result for: xxxxx.S01.1080p.WEB-DL.DD+.2.0.x264-xpost
2023-02-17 13:39:52 DEBUG :: SEARCHQUEUE-BACKLOG-xxxxx :: [nzb.su] :: Split out xxxxx.S01E06.xxxxx.1080p.WEB-DL from xxxxx.S01.1080p.WEB-DL.DD+.2.0.x264-xpost
2023-02-17 13:39:52 DEBUG :: SEARCHQUEUE-BACKLOG-xxxxx :: [nzb.su] :: Parsed xxxxx.S01E06.xxxxx.1080p.WEB-DL into xxxxx - S01E06 [ABD: False] [ANIME: False] [whichReg: ['standard']] Score: 5
2023-02-17 13:39:52 DEBUG :: SEARCHQUEUE-BACKLOG-xxxxx :: [nzb.su] :: Any,Best = [ 720p HDTV, 1080p HDTV, 720p WEB-DL, 1080p WEB-DL, 720p BluRay, 1080p BluRay ] [ 1080p WEB-DL ] Found = [ 1080p WEB-DL ]
[...]

PR #8366, opened 2023-02-19
Fix nzb splitting

Rubio9:
Fixes # 8365

- Fix splitting NZB to grab individual episodes out of season packs

This code does not look like it's been touched in some time, and there were a few issues preventing splitting a season pack's NZB into individual files for downloading. The initial error (in #8365) is fixed in nzbSplitter but there were then issues getting the split NZB files actually sent to (in my case) sabnzbd because of issues with getURL() and how it was being called; these fixes are in the other files here.

After these changes, things worked correctly for me.

Thanks for SickChill - it's a wonderful project, and I'd be happy to be able to give back to it!

### DIFF
--- a/sickchill/oldbeard/nzbSplitter.py
+++ b/sickchill/oldbeard/nzbSplitter.py
@@ -108,8 +108,8 @@ def strip_xmlns(element, xmlns):
     :param xmlns: xml namespace to be removed
     :return: processed element
     """
-    for e in element.iter():
-        e.tag = e.tag.replace("{" + xmlns + "}", "")
+    for subelement in element.iter():
+        subelement.tag = subelement.tag.replace("{" + xmlns + "}", "")
 
     return element
 
--- a/sickchill/oldbeard/nzbSplitter.py
+++ b/sickchill/oldbeard/nzbSplitter.py
@@ -108,9 +108,8 @@ def strip_xmlns(element, xmlns):
     :param xmlns: xml namespace to be removed
     :return: processed element
     """
-    element.tag = element.tag.replace("{" + xmlns + "}", "")
-    for cur_child in element.getchildren():
-        strip_xmlns(cur_child, xmlns)
+    for e in element.iter():
+        e.tag = e.tag.replace("{" + xmlns + "}", "")
 
     return element
 
@@ -167,7 +166,7 @@ def split_result(obj):
 
         want_ep = True
         for ep_num in parsed_obj.episode_numbers:
-            if not obj.extraInfo[0].wantEpisode(season, ep_num, obj.quality):
+            if not obj.show.wantEpisode(season, ep_num, obj.quality):
                 logger.debug("Ignoring result: " + new_nzb)
                 want_ep = False
                 break
@@ -175,7 +174,7 @@ def split_result(obj):
             continue
 
         # get all the associated episode objects
-        ep_obj_list = [obj.extraInfo[0].getEpisode(season, ep) for ep in parsed_obj.episode_numbers]
+        ep_obj_list = [obj.show.getEpisode(season, ep) for ep in parsed_obj.episode_numbers]
 
         # make a result
         cur_obj = classes.NZBDataSearchResult(ep_obj_list)
--- a/sickchill/providers/GenericProvider.py
+++ b/sickchill/providers/GenericProvider.py
@@ -213,8 +213,7 @@ class GenericProvider(object):
                             [
                                 ep
                                 for ep in episodes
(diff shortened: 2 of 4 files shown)


sickchill/oldbeard/nzbSplitter.py
```diff
@@ -108,9 +108,8 @@ def strip_xmlns(element, xmlns):
     :param xmlns: xml namespace to be removed
     :return: processed element
     """
-    element.tag = element.tag.replace("{" + xmlns + "}", "")
-    for cur_child in element.getchildren():
-        strip_xmlns(cur_child, xmlns)
+    for e in element.iter():
```
miigotu: Please rename e to subelement or something, single letter variables are discouraged.

sickchill/oldbeard/nzbSplitter.py
```diff
@@ -108,9 +108,8 @@ def strip_xmlns(element, xmlns):
...
+        e.tag = e.tag.replace("{" + xmlns + "}", "")
```
miigotu: This has to be done at top level, and then call strip_xmlns on each subelement.

sickchill/providers/GenericProvider.py
```diff
@@ -213,8 +213,7 @@ def find_search_results(self, show, episodes, search_mode, manual_search=False,
                             [
                                 ep
                                 for ep in episodes
-                                if (ep.season, ep.scene_season)[ep.show.is_scene] == (parse_result.season_number, parse_result.scene_season)[ep.show.is_scene]
-                                and (ep.episode, ep.scene_episode)[ep.show.is_scene] in parse_result.episode_numbers
+                                if ep.season == parse_result.season_number and ep.episode in parse_result.episode_numbers
```
miigotu, line 216: This change breaks scene numbered shows vs regular numbered shows, right?

miigotu (changes requested):
We have to discuss some of these changes, and if I'm wrong about some of them I need to understand why. Thanks for the PR!

sickchill/oldbeard/nzbSplitter.py
miigotu: iter does do all subelements of the element, and their children, but not the element itself. You only need to add the call on the element itself I think here, and then iter and also do it. Recursion unnecessary for sure.

Rubio9: The iter actually walks the root and all children, unlike what getchildren() used to do.

sickchill/providers/GenericProvider.py
Rubio9: No - this is what makes scene numbered shows actually work, and non-scene-numbered shows are unaffected by it.  I was seeing an issue with a scene-numbered, NON-anime, show, rejecting matches at this code because sickchill/oldbeard/name_parser/parser.py:280 already converts scene-numbering to indexer-numbering when scene numbering is enabled for a show.  (note the analogous code at lines 264-265 for Anime shows.)

Because it's already indexer-numbered, doing another conversion of indexer numbering to scene numbering HERE means we  will _not_ match episodes that are actual matches.

sickchill/oldbeard/nzbSplitter.py
Rubio9: Fixed.

sickchill/oldbeard/nzbSplitter.py
Rubio9:
```
import xml.etree.ElementTree as ET

a = ET.Element('a')
b = ET.SubElement(a, 'b')
c = ET.SubElement(a, 'c')
d = ET.SubElement(a, 'd')
e = ET.SubElement(b, 'e')
f = ET.SubElement(d, 'f')
g = ET.SubElement(d, 'g')

print("iter by itself will walk root and all children:")
print(f"{[elem.tag for elem in a.iter()]}")
# (outputs:) ['a', 'b', 'e', 'c', 'd', 'f', 'g']

print(f"To exclude the root itself I have to do it explicitly as I do here:")
print(f"{[elem.tag for elem in a.iter() if elem is not a]}")
# (outputs:) ['b', 'e', 'c', 'd', 'f', 'g']
```

sickchill/providers/GenericProvider.py
miigotu: I'll trust you on it, because I can't get to a computer right now to track it through the code.

sickchill/oldbeard/nzbSplitter.py
miigotu: Yeah I see that now, and I think the iter method is broken in that way by design which is strange. You can leave it this way, but an iterator in the traditional sense is supposed to work just like a `for x in element:`

Good job on following the docs 😁

Issue #8339, opened 2023-01-10
Client Error posting to discord webhook only from PostProcessor, snatch is fine.


JustSomeGuyNZ:
### Before submitting your issue:

Enable debug logging in SickChill settings, reproduce the error (be sure to disable after the bug is fixed)

Version: 2022.10.13
OS: Windows Server 2019
Browser: N/A
What you did/ happened/expected:
 At some point after NYE, web hook notifications from the post processor seem to have stopped working. I'm still getting notifications on snatch, but not when the post processor imports and attempts to notify on completion. Restarting the service doesn't seem to have changed anything, Debug logging shows a 400 client error.
What you expected:
Both snatch and download notifications to work.
Logs:
```
2023-01-10 20:14:33 INFO :: POSTPROCESSOR-AUTO :: Processing X:\Path\A TV Show\Season 02\A TV Show - S02E14 - Episode Name.1080p.AVC.EAC3.WEB-DL-METCON.mkv (None)
2023-01-10 20:14:33 DEBUG :: POSTPROCESSOR-AUTO :: Analyzing name A TV Show - S02E14 - Episode Name.1080p.AVC.EAC3.WEB-DL-METCON.mkv
2023-01-10 20:14:33 DEBUG :: POSTPROCESSOR-AUTO :: Trying a new way to verify if we can parse this file
2023-01-10 20:14:34 DEBUG :: POSTPROCESSOR-AUTO :: Parsed A TV Show - S02E14 - Episode Name.1080p.AVC.EAC3.WEB-DL-METCON into A TV Show - S02E14 [GROUP: METCON] [ABD: False] [ANIME: False] [whichReg: ['standard']] Score: 5
2023-01-10 20:14:34 DEBUG :: POSTPROCESSOR-AUTO :: Retrieving episode object for S02E14
2023-01-10 20:14:34 DEBUG :: POSTPROCESSOR-AUTO :: Snatch history had a quality in it, using that: 1080p WEB-DL
2023-01-10 20:14:34 DEBUG :: POSTPROCESSOR-AUTO :: Quality of the episode we're processing: 1080p WEB-DL
2023-01-10 20:14:34 DEBUG :: POSTPROCESSOR-AUTO :: This was manually downloaded but it appears to be better quality than what we have so I'm marking it as priority
2023-01-10 20:14:34 DEBUG :: POSTPROCESSOR-AUTO :: Is ep a priority download: True
[...]

Issue #8318, opened 2022-12-13
Kodi/Plex notification connect error kills post processor thread


adrhc:
**Version**: 2022.10.13
**OS**: Ubuntu 22.04.1 LTS
**Browser**: Chrome, any

**What you did**: 
I used Manual Post Processing to scan a directory pertaining to season 1 (same for 2, 3) of Star Wars: The Clone Wars.

**What happened**:
Only 1 movie is picked/detected for each scan.

**What you expected**:
All movies in the movies in the season to be detected such that to not be forced to manually post process the same directory, the same way, 10-20 times, depending on how many movies that season (directory) contains.

The seasons directories:
```
$ll Videos/sickrage-incomplete/Star\ Wars\:\ The\ Clone\ Wars/
total 12K
drwxrwxrwx 2 gigi gigi 4.0K Dec 13 20:31 Star.Wars.The.Clone.Wars.S01.1080p.DSNP.WEB-DL.DDP5.1.H.264-playWEB
drwxrwxrwx 2 gigi gigi 4.0K Dec 13 20:51 Star.Wars.The.Clone.Wars.S02.1080p.DSNP.WEB-DL.DDP5.1.H.264-playWEB
drwxrwxrwx 2 gigi gigi 4.0K Dec 13 20:12 Star.Wars.The.Clone.Wars.S03.1080p.DSNP.WEB-DL.DDP5.1.H.264-playWEB
```

The season 2 files:
```
$ll /home/gigi/Videos/sickrage-incomplete/Star Wars: The Clone Wars/Star.Wars.The.Clone.Wars.S02.1080p.DSNP.WEB-DL.DDP5.1.H.264-playWEB
total 32G
-rw-rw-rw- 2 gigi gigi 1.5G Dec 13 20:15 Clone.Wars.S02E01.Holocron.Heist.1080p.DSNP.WEB-DL.DDP5.1.H.264-playWEB.mkv
-rw-rw-rw- 1 gigi gigi 1.5G Dec 13 20:15 Star.Wars.Clone.Wars.S02E02.Cargo.of.Doom.1080p.DSNP.WEB-DL.DDP5.1.H.264-playWEB.mkv
-rw-rw-rw- 1 gigi gigi 1.5G Dec 13 20:15 Star.Wars.The.Clone.Wars.S02E04.Senate.Spy.1080p.DSNP.WEB-DL.DDP5.1.H.264-playWEB.mkv
[...]

Issue #8258, opened 2022-10-12
Downloaded episode notification not working since 2022.10.8


binbashblog:
### Before submitting your issue:

Enable debug logging in SickChill settings, reproduce the error (be sure to disable after the bug is fixed)

Version: 2022.10.08
OS: Ubuntu 22.04 (Kubernetes)

What happened: Snatched email notifications are being sent, however the downloaded notification no longer works. Test notifications get sent successfully. The 'notify on download' option is selected.
What you expected: Downloaded notifications to be sent to my email.

miigotu:
Yes. Read the first line of the first post in this thread lol.

BKSteve:
Just tested. Got [SN], [DL] and [PP] emails as expected.
The [DL] and [PP] emails come pretty much at the same time as SC only really knows it's downloaded when it's found in the folder when doing its Post Processing task.

binbashblog:
> Yes. Read the first line of the first post in this thread lol.

I already did, thanks. Debug logging is enabled, found no entries about generating emails, do you have an example log entry when an email is generated so I can search for it?

Generating a test email from SickChill works, snatched emails are received, as they have been for the last 6 years.

PostProcessing works as expected, the 'DL' or 'PP' email just doesn't get sent. I receive the snatched email. Has been working for 6 years up until build 2022.10.8 (worked 2022.9.17 and below)

Config.ini shows notifications are enabled.

  email_notify_onsnatch = 1
  email_notify_ondownload = 1
  email_notify_onpostprocess = 1

PR #8397, opened 2023-03-26
Notifications settings Check add Discord fix

BKSteve:
Fixes some notifications still attempting to send when main selector is deselected.

traceback.format_exc() inserted per #8339
and other partial notification issue items #8318  #8258


sickchill/oldbeard/notifications_queue.py
```diff
@@ -145,7 +145,7 @@ def _send_discord(self, webhook: str = None, name: str = None, avatar: str = Non
             return False
         except requests.exceptions.RequestException as error:
             if error.response.status_code != 429 or int(error.response.headers.get("X-RateLimit-Remaining")) != 0:
-                raise error
```
miigotu, line 148: Imo this should be:
```python
                logger.exception(traceback.format_exc())
                raise error
```

sickchill/oldbeard/notifications_queue.py
```diff
@@ -154,7 +154,7 @@ def _send_discord(self, webhook: str = None, name: str = None, avatar: str = Non
             )
             r.raise_for_status()
         except Exception as error:
-            logger.exception(f"Error Sending Discord message: {error}")
```
miigotu, line 157:
```python
            logger.exception(f"Error Sending Discord message: {error}")
            logger.exception(f"Traceback: {traceback.format_exc()}")
```

sickchill/views/config/notifications.py
```diff
@@ -217,22 +217,37 @@ def saveNotifications(
         results = []
 
         settings.USE_KODI = config.checkbox_to_value(use_kodi)
-        settings.KODI_ALWAYS_ON = config.checkbox_to_value(kodi_always_on)
```
miigotu: Every change in here I think is the wrong way to do this. The settings should always be read, or they will be saved wrong (reverted to defaults in config) on every shutdown.

Instead, make sure the methods in each notifier check that the notifier is enabled (they already should be, some are probably just missed)

Pseudo code example:

```python
class DiscordNotifier:
    def notify_subtitles(self, x):
        if settings.DISCORD_NOTIFY_SUBTITLES:
            return self.send_discord(x)

    def send_discord(self, x):
        if settings.USE_DISCORD:
            return discord.send(x)
```

miigotu (changes requested):
I think this approach breaks a few things.

sickchill/views/config/notifications.py
BKSteve: The discord part wasn't changed, the code change was just clearing the settings of sub items when you deselect the main check box and then do a save. 
As there is currently no check for the main item if a notification and the sub item is checked for notification it sends. So if you've set it before and then disable there is nothing that clears the sub items and thus some errors being thrown. 
The other way is to check the main item before the sub item in the queue. I thought this was easier 🪜

sickchill/oldbeard/notifications_queue.py
BKSteve: noted

sickchill/views/config/notifications.py
BKSteve: the `config_notifications.mako` loads all of them so when the save button is clicked they get saved in config.ini just as before.

BKSteve:
#8339 code fix added.

sickchill/views/config/notifications.py
miigotu: The settings shouldn't be cleared though, they just shouldn't be used if it's disabled when the code calls a notifier. My pseudoecode was showing how it should work, it wasn't specific to discord.

BKSteve: Point taken and understood. 
This was the lazy way, if main selector unchecked write uncheck to all child items when main page save is clicked, whilst leaving data. Should it be reenabled then a few check boxes and save.

miigotu: im going to check trhe code and see where the missed if was in the notifiers, and push it to this PR so the ui changes can go back.

BKSteve: I understand but if you unselect the main item then you don't want any of the sub items. You don't then need to chase down all the other code that only refers to the sub items.

Add: it's the Kodi area where the errors were happening.

miigotu: You need it to retain the settings though. You might just be disabling a notifier temporarily.

It's not chasing it down, the notifiers have a base class, and all of the notifier methods check if it's enabled before it tries to send anything.

miigotu: If I disabled something for a test or temporary because a service was down or something, and then when re-enabling I had to go find all my passwords and API keys and urls or whatever settings for that notifier just to re-enable it I'd be pissed.

BKSteve: That's why only clear check boxes all other data stays.

BKSteve: Does the `fix kodi only` pull work better for you? 

Can revert the `notifications.py` and then go through all the other `notifiers\*.py`

BKSteve: Fully revised

BKSteve:
Bump

BKSteve:
> I think this approach breaks a few things.

Totally removed those changes to notification.py and fixed up the others

miigotu:
Yeah I'll go over this tonight and see about merging it. It looks better for sure, I just don't have time to do an actual review right now.

sickchill/gui/slick/views/config_notifications.mako
```diff
@@ -879,7 +879,7 @@
                                 </div>
                                 <div class="row">
                                     <div class="col-md-12">
-                                        <label><b>${_('note')}:</b>&nbsp;${_('requires SickChill to be running on your Synology NAS.')}</label>
+                                        <label><b>${_('note')}:</b>&nbsp;${_('requires SickChill to be running on your Synology NAS (DSM 6 only).')}</label>
```
miigotu, line 882: This is also older DSM versions, not just DSM 6.

BKSteve: DSM 5 pretty much dead and AFAIK doesn't have Python 3

PR #8474, opened 2023-06-21
Leechtime

BKSteve:
Add minimum poll time to torrent leech

### DIFF
--- a/sickchill/oldbeard/providers/torrentday.py
+++ b/sickchill/oldbeard/providers/torrentday.py
@@ -38,7 +38,7 @@ class Provider(TorrentProvider):
         self.enable_cookies = True
 
         # Cache
-        self.cache = tvcache.TVCache(self, min_time=10)  # Only poll IPTorrents every 10 minutes max
+        self.cache = tvcache.TVCache(self, min_time=10)  # Only poll every 10 minutes max
 
     def login(self):
         cookie_dict = dict_from_cookiejar(self.session.cookies)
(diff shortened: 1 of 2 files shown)


miigotu:
The default cache time is 10 minutes: https://github.com/SickChill/sickchill/blob/master/sickchill/oldbeard/tvcache.py#L177

We arent spamming TL, searches are being re-done (cache is being broken) because of the bdecode error they caused with bad cloudfare config and a lack of an api.

Maybe set theirs to 30 minutes. I don't care if their provider is slow if they are talking trash about us rather than communicating with us. They don't want our users then we don't need to help their torrents stay seeded.

godofjesters:
instead of hardcoding a timeout, could a variable be set by the user to help randomize the timing and calm down the spamming? set minimum to 10 minutes and gof rom there

BKSteve:
There's already a user setting this is a backup so it uses the cache results. Ipt and td already have this setting I just duplicated it for tl

miigotu:
> There's already a user setting this is a backup so it uses the cache results. Ipt and td already have this setting I just duplicated it for tl

You didnt change anything though, the default value is 10 minutes (min_time is 10 if it isnt passed in, for ALL providers. The class defines the default value.) We use caching for every provider, and also respect etags and do not re-request a provider before the min_time AND the etags expire for the same search.

Really its most likely pirate admins who know little about how their backend works and probably dont even know what etags are lol

miigotu:
> instead of hardcoding a timeout, could a variable be set by the user to help randomize the timing and calm down the spamming? set minimum to 10 minutes and gof rom there

Just going to start setting it really high for unfriendly providers. Their reputation and traffic will falter when they punish an entire class of users over a handful of excessive users that run multiple instances of SC and thousands of shows. I'm not fond of people who do that myself.

BKSteve:
Especially when the added hits come from their messing up the coding of files to have `\n` at the back of them.

PR #8475, opened 2023-06-21
Leech filter test

BKSteve:
Simple trial of `\n` removal from TL file hash check.
Totally unproven to work.

Failed. Don't know my !r from my elbow.


BKSteve:
Need to process as a byte list not str so the `result.content = result.content.strip("\n")` doesn't work.
```
TypeError: a bytes-like object is required, not 'str'
```
That's why i did the `dirty` check the end as a string and if it was just hacked off the last 2 characters avoiding byte conversion issues. Not the nicest but proof of concept.

miigotu:
You can freely switch between bytes and string though.

PR #8482, opened 2023-06-29
Delta Time #2 Add New Fix

BKSteve:
add new show functionality such that a new show add will conform to the selected Previous Episode status.
add existing will apply new delta time setting

### DIFF
--- a/SickChill.py
+++ b/SickChill.py
@@ -114,7 +114,9 @@ class SickChill:
                 sys.stdout.write(f"{parser_name}: {result}\n")
             sys.exit(int(not any(results)))
 
-        if args.force_update:
+        if args.no_update:
+            settings.DISABLE_UPDATER = True
+        elif args.force_update:
             result = self.force_update()
             sys.exit(int(not result))  # Ok -> 0 , Error -> 1
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,8 +43,6 @@ packages = [
 "Bug Tracker" = "https://github.com/sickchill/sickchill/issues"
 "News" = "https://raw.githubusercontent.com/SickChill/sickchill.github.io/master/sickchill-news/news.md"
 "Discord" = "https://discord.com/invite/U8WPBdf"
-"Slack" = "https://sickchill.slack.com/"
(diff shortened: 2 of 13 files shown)


BKSteve:
One day I'll remember to set base branch to develop before clicking submit.

miigotu:
I'm not so sure about this, almost need to have a date the show was added to the database stored in the database to do this correctly

BKSteve:
This change is for adding new show effectively disable the date delta and follow the chosen settings. 
Existing shows add same as any daily changes would be treated as if already there and date delta applied. 
When testing this change has S1E1 set as wanted others with date delta.

miigotu:
I don't think this or the last PR was the place to make the change. The setting is right, but it should be used where episodes are checked for allowed/preferred if a better version needs downloaded. If the episode is any of downloaded/skipped/ignored/archived and the episode is older than this setting then it should be left alone. If it is wanted/failed, allow it to be wanted and searched for.

We also should set a longer interval for show updates when a show is ended for the show updater.

BKSteve:
Tv.py change from previous and this one interact. 
`1641 elif self.status in [UNAIRED, UNKNOWN]:`
So only these 2 status are affected by the timer everything else remains as before. 
The issue here is when you add a new show as everything is new they are unknown. By setting the timer to 0 temporarily the add new sequence avoids the timer and adds like it used to. Add existing has timer applied for all unknown unaired episodes.
Will need to do some more testing in regards unaired to ensure they do get set to wanted correctly. Or just make a new if for unaired and process the two separately. 
Open to suggestions or an alternate method.

miigotu:
There has to be a better way to do this, it just doesn't feel right.

Maybe adding an argument `force` to `loadEpisodesFromIndexer` and `loadFromIndexer` instead. That way if a user clicks the force update on the UI you can force all episodes to be updated later also.

BKSteve:
Yeah, it needs something that says we're adding either a new or existing show to the database so apply the configured settings of the page, otherwise use time delta on any other `unknown` status at all other times.

miigotu:
I think it works better this way, because the automatic updater will respect the settings, but forcing a show to update in the displayshow or massupdate will force a full update.

PR #8488, opened 2023-07-04
Page Sub Save Button not saving configuration

BKSteve:
Fix the multiplier that calculates the Backlog search minimum value which becomes a variable once daily search is greater than 357 minutes. If not a 60 mult then this can be an issue as if you try to `Save Changes` on the other tab there is no error visible.

Also consistent space before \ with other inner `Save Changes` buttons


BKSteve:
The other option would be to take out the `step 60` and leave it float?? 
The mako wouldn't see it as non compliant to the step factor and just go for it.

PR #8507, opened 2023-07-20
Pause and Ended Shows delay, xhtml_unescape

BKSteve:
- Put configuration of days delay to ended show below the hour to check show updates. Add some detail to notes.
- Include paused with ended in delay usage
- Fixed language identification code by .decode when adding show

* minor logging text fixes to mediabrowser and tivo


sickchill/views/manage/add_shows.py
```diff
@@ -105,7 +105,7 @@ def searchIndexersForShowName(self, search_term, lang=None, indexer=None, exact=
             final_results.sort(key=lambda x: x[4].lower().index(search_term.lower()))
             final_results.sort(key=lambda x: x[4].lower() == search_term.lower(), reverse=True)
 
-        lang_id = sickchill.indexer.lang_dict()[lang]
+        lang_id = sickchill.indexer.lang_dict()[lang.decode()]
```
miigotu: How is this getting encoded?

BKSteve: From what I can tell its the index.py change of #8482 which no longer has 'xhtml_escape' thus not changing b'en' to 'en'
Where it's getting set, good question. 
I didn't trace setting of the values yet just an initial tweak to make it go through, but it also gives b'0' or b'1' for check box value too.
Will look further.

miigotu: xhtml_escape (https://www.tornadoweb.org/en/stable/_modules/tornado/escape.html#xhtml_escape)
and xhtml_unescape
(https://www.tornadoweb.org/en/stable/_modules/tornado/escape.html#xhtml_unescape)

Do actually to_basesteing (`str(value)`) and `unicode(value)` which is crazy to me.

 I think the correct way to do this (I have been gradually doing this because it's the correct way, no method should have arguments directly) is to add before line 42 `lang = self.get_query_argument("lang")` and change line 45 to `search_term = self.get_query_argument("search_term")`

This way, the arguments are decoded properly and also sanitized against attacks with control characters and other injection attacks. There are other places that this needs done, in every web method that has arguments where we should be using get_query_argument (single argument)or get_query_arguments (list of arguments) for methods only used with GET, get_body_argument/get_body_arguments for methods that only use POST, and get_argumebt/get_arguments for methods that are called by both GET and POST

This method is only called with GET, but it could be changed to POST ideally by adding `method: 'POST'` in the $.ajax call for it in core.js and then use get_body_argument here.


Until we get all of the methods using self.get_x_argument, we probably need to change `def async_call` back to something like this:

```python
            for arg, value in kwargs.items():
                if len(value) == 1:
                    kwargs[arg] = unicode(value[0])
                elif isinstance(value, str):
                    kwargs[arg] = unicode(value)
                elif isinstance(value, list):
                    kwargs[arg] = [unicode(v) for v in value]
                else:
                    kwargs[arg] = value[0]
                elif not isinstance(value, [str, list]):
                    raise Exception
```

OR maybe this works for now? This would make every method use the sanitization from tornado correctly, and also would turn the arguments into the correct type. Then line 45 of this file would simply be able to he removed, no need for xhtml_unescape at all. 
```python
            for arg, value in kwargs.items():
                if len(value) == 1:
                    kwargs[arg] = self.get_arguments(arg)[0]
                elif isinstance(value, str):
                    kwargs[arg] = self.get_argument(arg)
                elif isinstance(value, list):
                    kwargs[arg] = self.get_arguments(arg)
                else:
                    kwargs[arg] = self.get_arguments(arg)[0]
                elif not isinstance(value, [str, list]):
                    raise Exception
```

miigotu: There should never be a place we need to .decode or .encode unless we are sending data to a 3rd party library or a quirky outgoing request.

miigotu: We also need to handle it in these other places as well https://github.com/search?q=repo%3ASickChill%2Fsickchill%20xhtml_&type=code

Also, quote/unquote and quote_plus/unquote_plus should be replaced with tornado.escape.url_escape/tornado.escape.url_unescape if they are actually still needed after we switch to get_query_argument/get_body_argument/get_argument (have to test), or removed if they aren't needed.

We want to avoid doing any extra escaping, quoting, encoding, and decoding. Otherwise we will end up with more problems like UnicodeDecodeError and/or values like passwords not being in the correct format and causing login issues with special characters incorrectly double-escaped/unescaped.

BKSteve: Of the parameters passed into sickchill/views/manage/add_shows.py line 39
> def searchIndexersForShowName(self, search_term, lang=None, indexer=None, exact=False)

all are byte encoded with search_term on line 45 `search_term = xhtml_unescape(search_term)` being unescaped.

I'm not skilled enough today to get the $.ajax call and get_argument to work and the unicode no go either.

what if `xhtml_unescape` the other 3 parameters.
lang = xhtml_unescape(lang) etc at the top?

Add: will continue learning and trying later.

miigotu: get_argument does the decode for you. It calls to_unicode on each argument and sanitizes it. I'm going to make the change here.

miigotu:
I got a bunch of it done. I'm going to try and do these other instances, and also unquote/unquote_plus that is used on route arguments (I don't want to go backwards to what was used before, it's what caused password encoding problems, show dir problems, etc and was hard to track down)

Before other routes are updated to use the correct argument handling, these need gone over (I'll try to do them tonight):
https://photos.app.goo.gl/HhRx8XnpoJjKL8KT7

miigotu:
Why did you comment out my changes in add_shows.py?

BKSteve:
They killed the processing. The items with the changes in index were nolonger byte strings see
https://discord.com/channels/502612977271439372/1132179311223373907
where I was journaling the testing.

miigotu:
@BKSteve I know this got way out of hand, I was enjoying it again for once lol. If you have the time could you check this over? We can merge this and continue in another PR for the get_argument and the template improvements.

miigotu:
I have a bunch of change here to push to this, including a change for the async_call (`if len(value) == 0: kwargs[key] = self.getargument(key)` and a bunch of fixes for traktTrending, the images API errors, qdsShowByID, and a bunch more. Don't worry about any of the current errors if you start working on this. Most of my changes are in add_shows, def index methods, and traktTrending.

miigotu:
Ahhhh I'm in too deep :laughing: 

Soon this needs to just be merged and work on fixing the issues I created in a new PR

miigotu:
That redirect at config save has always been there, I just think it was broken at some point. This is part of why people were complaining they couldn't tell that settings were saved. It only redirects to providers from savings provider settings.

BKSteve:
It was redirecting for every single save. No matter where in the config area to the /providers/ config page. Removing let the reload just reload the page

miigotu:
I see what I did wrong. The one you removed leave it removed, and move the `window.location.reload(true)` to where it was. There are no objects using`config_submitter_refreah`, everything uses `config_submitter` when saving config. 

The idea is to reload the same page you are on when saving config, for now, to give a visual indicator if they have desktop/WebKit notifications disabled. I also would like to add jQuery.notifications again like we used to, it was much nicer than. Pontify trying to do everything.

Lol, check how the 404 page is now, maybe that's a good way to mark saving config, and automatically remove the save_config site message when requesting another page or on a timer.

BKSteve:
Hmm. I see the error with `Save Providers`
```
HTTPServerRequest(protocol='http', host='localhost:8081', method='POST', uri='/config/providers/saveProviders', version='HTTP/1.1', remote_ip='127.0.0.1')

TypeError: '>' not supported between instances of 'int' and 'NoneType'
```
which only seems to happen on `http://localhost:8081/config/providers/` page only.

miigotu:
There shih be more to the traceback to tell us what variable

BKSteve:
The IMDbPY isn't being updated so changed to Cinemagoer 2023.5.1 in 
bunch of `_sr_` to `_sc_`
Added `g` to replaceAll in core.js 428 - now `#2` is displayed when adding shows. (steep learning curve).

now with `add existing` a more interesting item.
```
2023-07-29 20:39:44 DEBUG :: WEBSERVER-ADDSHOWS_0 :: Searching for Show with search term(s): ['Around the World in 80 Days'] on Indexer: theTVDB (exact: False)
```

Edit 20230730: 
See `Dateutil Rabbit Hole` resolving shows with names that get interpreted as dates.
Haven't followed further with IMDB adding and went for Trakt.

BKSteve:
When working on the trakt add show issue I note the directory and options settings no longer show at top of screen.

This proves to be an issue as the cur_dir isn't known and thus the add process gets stuck.
Edit: truncated as not needed

miigotu:
Yeah I don't want to add that code back. I'm in the middle of completely rewriting how trending and favorite and popular shows are added. They used to use a different method, I'm changing it to use new show directly and combining those 3 pages into 1.

BKSteve:
OK,  will step back from this area for a couple of days.

miigotu:
I messed up in a few places in the last 3 commits where `if s is not None:` went to `if season:`, the is not None needs to stay on that for season and episode.

miigotu:
I also still have to fix the add shows. If you want, you could check the add existing and normal show adding. I'll worry about the IMDb/trakt/popular lists.

sickchill/gui/slick/views/testRename.mako
```diff
@@ -77,27 +77,25 @@
                             <tbody>
                             % for current_episode in sorted(show.episodes[current_season], reverse=True):
                                 <%
-                                    ep_obj = show.episodes[current_season][current_episode]
-                                    if not (ep_obj and ep_obj._location):
+                                    episode_object = show.episodes[current_season][current_episode]
+                                    if not (episode_object and episode_object._location):
                                         continue
 
-                                    epStr = str(ep_obj.season) + "x" + str(ep_obj.episode)
```
BKSteve, line 84: The removal of this line and the removal of epStr below make the whole renaming function change to return SssEee rather than previous ssxee format. Breaking this function.
The select all items also gets broken.
If this is long term plan to change all to use SssEee rather than ssxee then let me know, otherwise I'd revert this.

sickchill/gui/slick/views/testRename.mako
```diff
@@ -77,27 +77,25 @@
...
-                                    curLoc = ep_obj.location[len(show._location)+1:]
+                                    curLoc = episode_object.location[len(show._location)+1:]
                                     curExt = curLoc.split('.')[-1]
-                                    newLoc = ep_obj.proper_path() + '.' + curExt
+                                    newLoc = episode_object.proper_path() + '.' + curExt
                                 %>
                                 <tr class="season-${current_season} ${('wanted', 'good')[curLoc == newLoc]} seasonstyle">
                                     <td class="col-checkbox">
                                         % if curLoc != newLoc:
-                                            <input type="checkbox" class="epCheck" id="${epStr}" name="${epStr}"  title="Episode check"/>
```
BKSteve, line 100: here's the other part

sickchill/gui/slick/views/testRename.mako
miigotu: Yeah that can be changed back. I'm letting my refactors get too big to catch everything.

sickchill/gui/slick/views/testRename.mako
miigotu: Yep undo that.

miigotu: i'll fix it now

sickchill/gui/slick/views/testRename.mako
miigotu: fixed

BKSteve:
> If you want, you could check the add existing and normal show adding.

Tested. Functional.
All show functions; Add, Add Existing (single and multiple), and all internal show edit items working.

Add: Mass Update not ok (expected)

miigotu:
What is wrong with mass update?

PR #8522, opened 2023-08-14
massEditUpdate locations

BKSteve:
Sanitise only new input from Edit Show page otherwise use already processed location.

Neaten up a log line.


BKSteve:
The qualities drop list is not in the Sort order it should be. `HD` is coming after the other HD items. 

change argument to arguments for qualities

PR #8544, opened 2023-08-20
manage index args

BKSteve:
manage index args cleanup and variables.


sickchill/views/manage/index.py
```diff
@@ -608,106 +610,115 @@ def massEditSubmit(self):
...
+            message_detail += "</li></ul>"
 
         if updates + refreshes + renames + subtitles:
-            ui.notifications.message(_("The following actions were queued") + ":", messageDetail)
+            ui.notifications.message(_("The following actions were queued") + ":", message_detail)
 
         return self.redirect("/manage/")
 
-    def failedDownloads(self, limit=100, toRemove=None):
+    def failedDownloads(self, limit=100):
```
miigotu: Is there a reason to not also get_argument limit?

BKSteve: A little less familiar, so left it was all.

miigotu: If we switch to using get_*_argument on a method, you have to switch all of them or leave them all in the function signature, otherwise the router will throw an exception.

Not sure if this method is called by a link or from js, but if it's a get/link `limit = self.get_query_argument("limit", 100)` and if it's a post `get_body_argument`

miigotu: Strangely, this method is used both ways, post and get. At another time I'll be changing the js form to use multiple fields with the same name to pass things like to_remove instead of joining and splitting with | (so we can use `get_body_arguments` with the `s` to get a list of values directly, so I can go back and change some stuff later too. But for now in this method everything should be `get_argument`, which works with both post and get, `limit = self.get_argument("limit", 100)`

BKSteve: I must admit this https://www.tornadoweb.org/en/stable/web.html isn't very clear on differentiation on their use.

Add: Found this https://stackoverflow.com/questions/34818996/tornado-what-is-the-difference-between-requesthandlers-get-argument-get-qu

The Request.get_query_argument() looks for URL parameters, the RequestHandler.get_body_argument() lets you retrieve parameters set in the POST body. The RequestHandler.get_argument() method retrieves either a body or a URL parameter (in that order).

BKSteve: So to be clear use `get_argument` and `get_arguments` where there are lists that a split with | use `get_arguments`
As in, really only use `get_argument` when we know it's only one parameter.
Don't use `get_*_argument/s`?

BKSteve: have to do `limit = self.get_argument("limit", "100")` as doesn't like int wants a str.

BKSteve: > so we can use `get_body_arguments` with the `s` to get a list of values directly

Adding the `s` gives a list instead of a single item, the `.split` in `to_update = to_update.split("|") if to_update else []` then gives `Unresolved attribute reference 'split' for class 'list' `
So some editing will be required, but as the split line code would be removed adding an `s` at the same time shouldn't be too difficult.

miigotu: > So to be clear use `get_argument` and `get_arguments` where there are lists that a split with | use `get_arguments` As in, really only use `get_argument` when we know it's only one parameter. Don't use `get_*_argument/s`?

No, for now we always use the variant without the s. To use get_*_arguments (with s) we have to make js and template changes to the forms to get rid of passing things with a | joined string. I'll do that in another set.

miigotu: The big thing is this particular route is call post as a get request (query arguments) and a post request (body arguments)

That means to get the arguments in this one route, we have to use get_argument (not get_query_argument or get_body_argument) because for example get_body_argument("limit") won't have a value of it's a get request, since the argument is passed as a query, not a body.

Usually we want to use the specific request type and use the more fine grained get_*_argument methods, for error control and protection from form/url request forging. We just can't in this one route because of mixed calling request methods.

sickchill/views/manage/index.py
miigotu, line 1: Have to revert this one or change the js

BKSteve: The modifications with the current index.py are working with the js.
The two qualities lines with `self.get_body_arguments` do work as they process a list.

the `limit = self.get_argument("limit", "100")` is the only `get_argument` but testing this `def failedDownloads` to ensure it works is the issue.

BKSteve: Could just leave `def failedDownloads` unchanged at this point.

miigotu: Yeah leave all the split("|") ones how they were, after you get_argument (no s), and do `limit = self.get_argument("limit", "100")`

I can adjust it maybe later if you can't get to it.

sickchill/views/manage/index.py
```diff
@@ -608,106 +610,116 @@ def massEditSubmit(self):
...
 
         if updates + refreshes + renames + subtitles:
-            ui.notifications.message(_("The following actions were queued") + ":", messageDetail)
+            ui.notifications.message(_("The following actions were queued") + ":", message_detail)
 
         return self.redirect("/manage/")
 
-    def failedDownloads(self, limit=100, toRemove=None):
+    def failedDownloads(self):
+        to_remove = self.get_body_argument("toRemove", None)
```
miigotu: If you change this to 

```diff
---       to_remove = self.get_body_argument("toRemove", None)
+++       to_remove = self.get_body_argument("toRemove", "")
```
(default being a string "")

Later you don't need to check if to_remove, just split it. 

```diff
---        to_remove = to_remove.split("|") if to_remove else []
---
---        for release in to_remove:
+++        for release in to_remove.split("|"):
```

We can also change the argument names in the template and js, so it's passed as `remove` or `to_remove` instead of `toRemove` and its consistent, not important right this instant though.

sickchill/views/manage/index.py
```diff
@@ -608,106 +610,116 @@ def massEditSubmit(self):
...
-    def massUpdate(self, toUpdate=None, toRefresh=None, toRename=None, toDelete=None, toRemove=None, toMetadata=None, toSubtitle=None):
-        toUpdate = toUpdate.split("|") if toUpdate else []
-        toRefresh = toRefresh.split("|") if toRefresh else []
-        toRename = toRename.split("|") if toRename else []
-        toSubtitle = toSubtitle.split("|") if toSubtitle else []
-        toDelete = toDelete.split("|") if toDelete else []
-        toRemove = toRemove.split("|") if toRemove else []
-        toMetadata = toMetadata.split("|") if toMetadata else []
+    def massUpdate(self):
+        to_update = self.get_body_argument("toUpdate", None)
```
miigotu: Same up here, make the default values "" and then just split them. 

For all the arguments,  that need split, default to `"" `
```diff
---        to_metadata = self.get_body_argument("toMetadata", None)
+++        to_metadata = self.get_body_argument("toMetadata", "")
```
```diff
---        to_metadata = to_metadata.split("|") if to_metadata else []
+++        to_metadata = to_metadata.split("|")
```

The `if x else []` was a protection from the variable being None, and trying to split NoneType. But since we can control the argument now and sanitize it, and make sure the default is the correct type it's just using unnecessary CPU cycles and making the code longer and uglier 😂

miigotu:
For certain things, you can change the js so it doesn't join the array, and then you use get_arguments in the python. Example is anyQualities and best qualities for saveAddShowDefaults, https://github.com/SickChill/sickchill/blob/7715e0e51c9d8bd72807ef5c1b0de1f3e1b5c841/sickchill/gui/slick/js/core.js#L4081

You could change those two lines in core.js:
```diff
---                   anyQualities: anyQualArray.join(','),
---                    bestQualities: bestQualArray.join(',')
+++                   anyQualities: anyQualArray,
+++                    bestQualities: bestQualArray
```
And then change the get_argument in saveAddShowDefaults, https://github.com/SickChill/sickchill/blob/7715e0e51c9d8bd72807ef5c1b0de1f3e1b5c841/sickchill/views/config/general.py#L48-L56

```diff
---        if anyQualities:
---            anyQualities = anyQualities.split(",")
---        else:
---            anyQualities = []
---
---        if bestQualities:
---            bestQualities = bestQualities.split(",")
---        else:
---            bestQualities = []
+++        any_qualities = self.get_body_arguments("anyQualities") # since it's post and an array
+++        best_qualities = self.get_body_arguments("bestQualities")
```

In the one that is using get_arguments in manage/index for any/bestQualities, I think I already had changed the js for it, and needed to test that it was done correctly. Im glad it's working.

sickchill/views/manage/index.py
BKSteve: > For all the arguments, that need split, default to `""`

That's where I messed up. I left None and then tried the `.split` and it would fail, using `""` resolves it and cleans up the code.

sickchill/views/config/general.py
```diff
@@ -45,15 +45,8 @@ def saveAddShowDefaults(
...
-        if anyQualities:
-            anyQualities = anyQualities.split(",")
-        else:
-            anyQualities = []
-
-        if bestQualities:
-            bestQualities = bestQualities.split(",")
-        else:
-            bestQualities = []
+        anyQualities = self.get_body_arguments("anyQualities")
```
BKSteve: This throws a 
`NameError: name 'self' is not defined`
after reverting the lines in core.js
```
anyQualities: anyQualArray.join(','),
bestQualities: bestQualArray.join(','),
```

without reverting `join` in core.js it throws missing 6 parameters error.

miigotu: The name error is a different issue. It's because this route is decorated as a staticmethod (since it didn't use self before)

You have to remove `@staticmethod `on https://github.com/SickChill/sickchill/pull/8544/files/6a64051ccb7b3062afcfb984d83b1406869aea3a#diff-b70a50d70ca28d16fa38da024c378eab3b03a2c4c3e6b68018e1d1814a4da5fdR35
And set the first argument to the method as self. I'll make the change now.

sickchill/views/config/general.py
```diff
@@ -28,12 +28,11 @@ def index(self):
     def generateApiKey():
         return helpers.generateApiKey()
 
-    @staticmethod
-    def saveRootDirs(rootDirString=None):
-        settings.ROOT_DIRS = rootDirString
+    def saveRootDirs(self):
+        settings.ROOT_DIRS = self.get_query_argument("rootDirString")
 
-    @staticmethod
```
miigotu, line 35: Now that it has the decorator removed and self added as the argument from the class, you can do the other arguments as well in the proper way.

BKSteve: done other arguments.

sickchill/views/config/general.py
```diff
@@ -31,36 +31,30 @@ def generateApiKey():
...
-        blacklist,
-        subtitles=False,
-        anime=False,
-        scene=False,
-        defaultStatusAfter=WANTED,
-    ):
-        anyQualities = self.get_body_arguments("anyQualities")
-        bestualities = self.get_body_arguments("bestQualities")
+    def saveAddShowDefaults(self):
+        anyQualities = self.get_body_argument("anyQualities", "").split(",")
```
BKSteve: back to arguments and split

sickchill/views/config/general.py
```diff
@@ -31,36 +31,30 @@ def generateApiKey():
...
+        subtitles = self.get_body_argument("subtitles", "False")
+        anime = self.get_body_argument("anime", "False")
+        scene = self.get_body_argument("scene", "False")
 
-        newQuality = Quality.combineQualities([int(quality) for quality in anyQualities], [int(quality) for quality in bestQualities])
+        settings.STATUS_DEFAULT = int(self.get_body_argument("defaultStatus"))
 
-        settings.WHITELIST_DEFAULT = whitelist
-        settings.BLACKLIST_DEFAULT = blacklist
+        if '' in anyQualities:
```
BKSteve: Need to check it's clean, this might not be the nicest way.

sickchill/views/config/general.py
BKSteve: did some reordering too

sickchill/views/config/general.py
miigotu: Why would we need to do that? If there's no quality selected, it passes an empty array `[]`, or at least it should

sickchill/views/config/general.py
```diff
@@ -28,47 +28,33 @@ def index(self):
...
-        if anyQualities:
-            anyQualities = anyQualities.split(",")
-        else:
-            anyQualities = []
+    def saveAddShowDefaults(self):
+        anyQualities = self.get_body_argument("anyQualities", "").split(",")
+        bestQualities = self.get_body_argument("bestQualities", "").split(",")
+        defaultSeasonFolders = self.get_body_argument("defaultSeasonFolders")
+        subtitles = self.get_body_argument("subtitles", "False")
+        anime = self.get_body_argument("anime", "False")
```
miigotu: `False` not `"False" `

sickchill/views/config/general.py
miigotu: This was before my commits in the other pr, I'm adding code changes for removing the splits in
setStatus
massEdit
massUpdate
failedDownloads

sickchill/views/config/general.py
miigotu: That was part of the other PR, it just needed merged (I changed python here and js there)

miigotu:
Aside from the manage/index stuff, we also need. To make sure to test these: (some are manage/index)
setStatus
massEdit
massUpdate
failedDownloads

I don't want to join and split any arrays, both the web server and python and js handles them natively without conversion.

Hopefully this PR is near complete, before it gets too big again.

sickchill/views/config/general.py
```diff
@@ -28,48 +28,26 @@ def index(self):
...
-        if bestQualities:
-            bestQualities = bestQualities.split(",")
-        else:
-            bestQualities = []
+    def saveRootDirs(self):
+        settings.ROOT_DIRS = self.get_query_argument("rootDirString")
 
-        newQuality = Quality.combineQualities([int(quality) for quality in anyQualities], [int(quality) for quality in bestQualities])
+    def saveAddShowDefaults(self):
+        anyQualities = [int(quality) for quality in self.get_body_arguments("anyQualities")]
```
BKSteve: This throws an error as `'anyQualities': [b'4,16,32,64,128,256']` is a single item string list in a list, so needs splitting to iterate through as it comes from `anyQualities: anyQualArray.join(),`.
If per previous core.js doesn't have the join and only `anyQualArray` then the return values don't get processed by `body_arguments` as anyQualities comes as `anyQualities%5B%5D` ([] characters). Which I couldn't work out how to filter for with `body_arguments`
What's needed is `'anyQualities': [b'4',b'16',b'32',b'64',b'128',b'256']`

BKSteve: Let me clarify a little more.
_core.js_ with `anyQualities: anyQualArray,` and 
_general.py_ with `anyQualities = [int(quality) for quality in self.get_body_arguments("anyQualities[]")]`

first _save_ gets `'anyQualities[]' = [b'4', b'16', b'32', b'64', b'128', b'256']`
if quality changed and _save_ again without refreshing page `'anyQualities[]' = []` 

So if `inc_addShowOptions.mako` could refresh after the save and then next save would iterate `def async_call` `self.request.method == "POST"` it should all work.

miigotu: > This throws an error as `'anyQualities': [b'4,16,32,64,128,256']` is a single item string list in a list, so needs splitting to iterate through as it comes from `anyQualities: anyQualArray.join(),`.
> If per previous core.js doesn't have the join and only `anyQualArray` then the return values don't get processed by `body_arguments` as anyQualities comes as `anyQualities%5B%5D` ([] characters). Which I couldn't work out how to filter for with `body_arguments`
> What's needed is `'anyQualities': [b'4',b'16',b'32',b'64',b'128',b'256']`

I'll fix that. Don't worry 😂

miigotu: > This throws an error as `'anyQualities': [b'4,16,32,64,128,256']` is a single item string list in a list, so needs splitting to iterate through as it comes from `anyQualities: anyQualArray.join(),`.
> If per previous core.js doesn't have the join and only `anyQualArray` then the return values don't get processed by `body_arguments` as anyQualities comes as `anyQualities%5B%5D` ([] characters). Which I couldn't work out how to filter for with `body_arguments`
> What's needed is `'anyQualities': [b'4',b'16',b'32',b'64',b'128',b'256']`

anyQualities[] is correct. It needs passed differently in the js, I'll make the change.

miigotu: > > This throws an error as `'anyQualities': [b'4,16,32,64,128,256']` is a single item string list in a list, so needs splitting to iterate through as it comes from `anyQualities: anyQualArray.join(),`.
> > If per previous core.js doesn't have the join and only `anyQualArray` then the return values don't get processed by `body_arguments` as anyQualities comes as `anyQualities%5B%5D` ([] characters). Which I couldn't work out how to filter for with `body_arguments`
> > What's needed is `'anyQualities': [b'4',b'16',b'32',b'64',b'128',b'256']`
> 
> anyQualities[] is correct. It needs passed differently in the js, I'll make the change. 

And anyQualities is not joined in the js, it's an array. It needs to stay the array in js not a string, and it needs to be get_arguments (get the array, not a string) in python. I removed the join in the js already, you added it back and I removed it again. get_arguments("anyQualities") calls the anyQualities[] from the request body. You might need to clear your request cache in your browser or enable developer mode if you are still getting a joined string, or pull the latest changes. 

Edit: I guess that commit where I removed the .join in core.js somehow didn't make it in the PR?

BKSteve: > Edit: I guess that commit where I removed the .join in core.js somehow didn't make it in the PR?

No, I added _join_ back then worked the problem some more. 
Learnt the issue was the refreshing of the page and the `POST`.
It would always work once then not, so I'd go change stuff. 🪛 💻 🚒

sickchill/views/config/general.py
BKSteve: I put `"False"` as if using `False` Pycharm warning `Expected type 'None | str | _ArgDefaultMarker', got 'bool' instead `
eg
[`force = self.get_query_argument("force", default=False)`](https://github.com/SickChill/SickChill/blob/0f1933eb63226a67fb942c850614198c8a111e05/sickchill/views/home.py#L1319)

BKSteve:
Did some array arguments [] work.

_Aside from the manage/index stuff, we also need. To make sure to test these: (some are manage/index)_

- setStatus - tested individual show set multiple episode status change, working
- massUpdate  
  - _delete_ fixed [core.js#L3443](https://github.com/SickChill/SickChill/blob/96c89c7ad090e4e0cab5161081334a29ad0410bc/sickchill/gui/slick/js/core.js#L3443).
  - _massEdit_ is not, the `Edit Selected` button sends to `def massEdit(self):` but doesn't render the `manage_massEdit.mako` page. [core.js#L3366](https://github.com/SickChill/SickChill/blob/96c89c7ad090e4e0cab5161081334a29ad0410bc/sickchill/gui/slick/js/core.js#L3366)
- massEdit - unable to access per above.
  - setDefaultStatus - unable to access
- failedDownloads - **don't know how to test**

sickchill/views/config/general.py
miigotu: Yeah I think it was saved as a comma separated list of strings in your config or something because of the changes and the original join, without the join it should work fine, we might need to add the brackets in the js side though.

BKSteve: Without the join it works fine but only once, if you change the quality and click save a second time then it fails. 
If you refresh the page and change the selection again it will work.
So the Q is, is there a way to ensure on a second change and save that the `argument/s` are again `POST`ed?

miigotu: Yes absolutely. Without actually being able to test it, I think it might be this:

```diff
---    const anyQualArray = [];
---    const bestQualArray = [];
+++   let anyQualArray = [];
+++   let bestQualArray = [];
```

Those variables are set constant, so they can't be reassigned to an empty array on the second click. You could empty the array or add items too it, but not reassign it. Changing it to `let` allows it to be mutable and reassigned.

miigotu: Unless you mean because the button is greyed out/disabled? It's supposed to be, but I can make that re-enable when something changes I think or just not disable the button. 

Nevermind, it already re-enables the button when changing the qualities, so you must mean the other problem. I'm pushing that const/let change in a minute, check it if you can. https://github.com/SickChill/sickchill/blob/7715e0e51c9d8bd72807ef5c1b0de1f3e1b5c841/sickchill/gui/slick/js/core.js#L4095

BKSteve:
The eslint converted all the let back to const

miigotu:
> The eslint converted all the let back to const

Yeah that's fine, it should work with const because it goes out of scope between calls.

BKSteve:
Still only able to save properly once, second change and save works. Third one fails. (_edited_)

Also the Mass Update > Edit Selected isn't opening the massEdit page and don't see any error or info on why not.

BKSteve:
Ran and tested but note `massUpdate is still broken` did not prove to be so. All working quite nicely.
Adjust font size to of Mass Edit to be more prominent.
Change variable show_list to be edit_show_list as show_list is another variable already used in massEdit to be clean.

miigotu:
On the mass update page, if you checkbox a show to rescan or update or even remove and then click submit, does that thing happen? It didn't for me.

PR #8568, opened 2023-09-27
toDelete combos

BKSteve:
tweak core.js to work with all massUpdates and massEdits, many combos tested

### DIFF
--- a/sickchill/gui/slick/js/core.js
+++ b/sickchill/gui/slick/js/core.js
@@ -3422,6 +3422,7 @@ const SICKCHILL = {
                 });
 
                 if (deleteArray.length > 0) {
+                    submit = false;
                     $.confirm({
                         title: 'Delete Shows',
                         text: 'You have selected to delete ' + deleteArray.length + ' show(s).  Are you sure you wish to continue? All files will be removed from your system.',
@@ -3431,14 +3432,11 @@ const SICKCHILL = {
                         post: false,
                         confirm() {
                             for (const deleteItem of deleteArray) {
-                                form.append($('<input />', {
-                                    name: 'toDelete',
-                                    type: 'hidden',
-                                    value: deleteItem,
-                                }));
+                                form.append($('<input />', {name: 'toDelete', type: 'hidden', value: deleteItem}));
                             }
 
-                            submit = true;
+                            form.appendTo('body');
+                            form.submit();
                         },
                     });
                 }


sickchill/gui/slick/js/core.js
```diff
@@ -3431,14 +3432,11 @@ const SICKCHILL = {
...
-                                    name: 'toDelete',
-                                    type: 'hidden',
-                                    value: deleteItem,
-                                }));
+                                form.append($('<input />', {name: 'toDelete', type: 'hidden', value: deleteItem}));
                             }
 
-                            submit = true;
+                            form.appendTo('body');
+                            form.submit();
```
miigotu, line 3439: The form is not supposed to get submitted here. Delete and all of the other items are done in the same form. There is only one form submit.

sickchill/gui/slick/js/core.js
```diff
@@ -3422,6 +3422,7 @@ const SICKCHILL = {
                 });
 
                 if (deleteArray.length > 0) {
+                    submit = false;
```
miigotu, line 3425: This doesn't make any sense, if previous checks already set it to true it needs to stay true. It's not two separate forms now, it's one combined form. If we don't confirm the delete, it's still supposed to continue on and submit the form later just without the toDelete inputs added. If confirmed we add the toDelete inputs and set it to be submitted. The code was correct before, it's just not submitting how you think it's supposed to. It isn't supposed to cancel the entire submission of the form unless there is nothing else checked on the massUpdate.

miigotu (dismissed):
Removed my spam, I wasn't trying to come at you like that I was tired and frustrated.

miigotu:
Also, I don't even need to create the form there, there is already one on the massUpdate page I can append to that form and submit that. I'll revert this and make that change and do testing later. What browser are you using?

sickchill/gui/slick/js/core.js
```diff
@@ -3431,14 +3432,11 @@ const SICKCHILL = {
                         post: false,
```
miigotu, line 3432: This `post: false;` shouldn't be here

sickchill/gui/slick/js/core.js
```diff
@@ -3431,14 +3432,11 @@ const SICKCHILL = {
...
                         confirm() {
```
miigotu, line 3433: This is probably the problem:

This should not be a `confirm():` call, it should be `confirm: function () {`

Also, I need to check which version of jquery-confirm is being used because v3 is a completely different syntax and settings/config. v2.5.1 is the last version with a `confirm:` parameter.

Going to look at this here in a bit.

miigotu (changes requested):

sickchill/gui/slick/js/core.js
BKSteve: Yeah, was trying a solution. 
By setting to false here it stopped it skipping through the confirm and would then post the toDelete with the other to* items. 
It would never submit the toDelete ever.

sickchill/gui/slick/js/core.js
BKSteve: steep learning curve on this stuff.

miigotu:
Yeah it's ok, I get what you are trying to do, I just get frustrated when something doesn't work. It needs to be restructured more so all of the inputs are already handled. I have an almost finished commit to fix it, there should only be one path to submit.

BKSteve:
closed

PR #8573, opened 2023-10-03
fix manual snatch of jackett results mistakenly being sent using sab/…

miigotu:
…nzbget settings

smarter dupescore for nzbget, if its preferred give highest scores, allowed higher score, not None regular score, or result quality + 10 for propers smarter priority/top adding for nzbget, set them based on newest result in the list of episodes, rather than the last episode in the list. they arent always sorted add method to test nzbget connection (not implemented in UI yet)

check if results are torrent or nzb right in the result class rename/refactor sbtimezone,sbdatetime,sbdate,sbf* to sc* pep8 a bunch of variable names, and use fstrings everywhere I touch use more pathlib.Path

found issue in snatch_episode with marking bad releases as failed, needs fixed


sickchill/oldbeard/search.py
```diff
@@ -125,8 +97,25 @@ def snatch_episode(result: "SearchResult", end_status=SNATCHED):
                 snatched_result = client.sendTORRENT(result)
             else:
                 logger.warning("Torrent file content is empty")
-                History().logFailed(result.episodes, result.name, result.provider)
+                # TODO: This is broken!!
+                # History().logFailed(result.episodes, result.name, result.provider)
```
miigotu, line 101: @BKSteve idk if you think you can track down how this should work lol, I think the history junk needs refactored/rewritten.

BKSteve: taking a look.
`[0]` seems to enter data into failed db

BKSteve:
You'll also need to clone https://sickchill.github.io/sb_network_timezones/ to https://sickchill.github.io/sc_network_timezones/ if you want to transition all sb items out.

In `config_search.mako` the added hidden `951 <div id="content_use_torrents" hidden>` is not showing data when going to page and tab if the box is checked. You need to uncheck and check to show settings.

miigotu:
> You'll also need to clone https://sickchill.github.io/sb_network_timezones/ to https://sickchill.github.io/sc_network_timezones/ if you want to transition all sb items out.
> 
> In `config_search.mako` the added hidden `951 <div id="content_use_torrents" hidden>` is not showing data when going to page and tab if the box is checked. You need to uncheck and check to show settings.

Yeah I know that part with the sb_network_tomezones

The use_torrents I need to change the hidden to use the config setting like `${("hidden", "")[settings.USE_TORRENTS]` Maybe I'll make a filter to call there that won't look so ugly and it will be like `${'USE_TORRENTS' | check_hidden}`

About the history item, I added the one with the episodes[0], the other one just passed        `History().logFailed(result.episodes, result.name, result.provider)` which were all the wrong type of argument. 
I think that part of history needs done like logSnatch where it loops through all of the episodes in the result, but have two ways to pass a failed item. Either as an episode_object, or as a result object (like logSnatch)

miigotu:
For those enabler items, maybe I could add a `data-hidden="USE_TORRENTS"``

And then in js, something like:

```javascript
let enabler_items = $('div[data-hidden!=""]:[data-hidden]');
enabler_items.each(function () {
    let item = $(this);
    getMeta(item.data('hidden')) ? item.show() : item.hide();
});
```

I'll have to think about this one, I'm trying to make some of this code be able to be used in the new UI.  I was trying to make the html do the work and cut down on the js

miigotu:
The logFailed should probably be something like:
```
for episode in result.episodes:
    History().logFailed(episode, result.name, result.provider.name)
```

BKSteve:
With History, I'm not familiar and tested only one episode, so no need to iterate, by doing a single episode search.
At line 93 I did this `result.content = ""`

BKSteve:
All functions of massUpdate work, config too, except the `checkbox` when fist populating the page a checked box doesn't display the fields, they remain hidden. Uncheck and check makes them display, this is throughout with the `hidden` item in the mako.

PR #8606, opened 2023-10-28
hidden reversal

BKSteve:
Fixes the one `hidden` that needed to be a `selected` 
The only one found.
Also hooked the `Associated file extensions` list to be hidden if unselected.

## Summary by CodeRabbit

# Release Notes

- Refactor: Updated the Post Processing settings page to improve visibility of the option to move associated files based on user settings.
- Bug Fix: Adjusted the Providers settings page to correctly enable or disable a checkbox based on the provider's daily capabilities.
- Bug Fix: Modified the Edit Show page to correctly disable the 'Season Folders' checkbox based on the user's naming settings.

These changes enhance the user interface's intuitiveness by ensuring settings and options are displayed or hidden based on relevant conditions, improving overall user experience.

### DIFF
--- a/sickchill/gui/slick/views/config_postProcessing.mako
+++ b/sickchill/gui/slick/views/config_postProcessing.mako
@@ -186,7 +186,7 @@
                             </div>
                         </div>
 
-                        <div class="field-pair row" id="content_move_associated_files" ${hidden(settings.MOVE_ASSOCIATED_FILES)}>
+                        <div class="field-pair row" id="content_move_associated_files" ${selected(settings.MOVE_ASSOCIATED_FILES)}>
                             <div class="col-lg-3 col-md-3 col-sm-4 col-xs-12">
                                 <label class="component-title">${_('Rename .nfo file')}</label>
                             </div>
@@ -194,9 +194,6 @@
                                 <input type="checkbox" name="nfo_rename" id="nfo_rename" ${checked(settings.NFO_RENAME)}/>
                                 <label for="nfo_rename">${_('rename the original .nfo file to .nfo-orig to avoid conflicts?')}</label>
                             </div>
-                        </div>
-
-                        <div class="field-pair row">
                             <div class="col-lg-3 col-md-3 col-sm-4 col-xs-12">
                                 <label class="component-title">${_('Associated file extensions')}</label>
                             </div>
--- a/sickchill/gui/slick/views/config_providers.mako
+++ b/sickchill/gui/slick/views/config_providers.mako
@@ -197,7 +197,7 @@
                                 % endif
 
                                 % if hasattr(curNewznabProvider, 'enable_backlog'):
-                                    <div class="field-pair row${hidden(curNewznabProvider.supports_backlog)}">
+                                    <div class="field-pair row">
                                         <div class="col-lg-3 col-md-4 col-sm-5 col-xs-12">
                                             <label class="component-title">${_('Enable backlog searches')}</label>
                                         </div>
@@ -310,7 +310,7 @@
                                 % endif
 
                                 % if hasattr(curNzbProvider, 'enable_backlog'):
-                                    <div class="field-pair row${hidden(curNzbProvider.supports_backlog)}">
+                                    <div class="field-pair row">
                                         <div class="col-lg-3 col-md-4 col-sm-5 col-xs-12">
                                             <label class="component-title">${_('Enable backlog searches')}</label>
                                         </div>
@@ -669,7 +669,7 @@
                                 % endif
 
                                 % if hasattr(curTorrentProvider, 'enable_backlog'):
-                                    <div class="field-pair row${hidden(curTorrentProvider.supports_backlog)}">
+                                    <div class="field-pair row">
                                         <div class="col-lg-3 col-md-4 col-sm-5 col-xs-12">
                                             <label class="component-title">${_('Enable backlog searches')}</label>
                                         </div>
--- a/sickchill/gui/slick/views/config_providers.mako
+++ b/sickchill/gui/slick/views/config_providers.mako
@@ -658,7 +658,7 @@
                                             <input type="checkbox" name="${curTorrentProvider.get_id("_enable_daily")}"
                                                    id="${curTorrentProvider.get_id("_enable_daily")}"
                                                    ${checked(curTorrentProvider.daily_enabled)}
-                                                   ${disabled(not curTorrentProvider.can_daily)}
+                                                   ${disabled(curTorrentProvider.can_daily)}
                                             />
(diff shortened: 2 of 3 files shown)


sickchill/gui/slick/views/config_providers.mako
```diff
@@ -658,7 +658,7 @@
                                             <input type="checkbox" name="${curTorrentProvider.get_id("_enable_daily")}"
                                                    id="${curTorrentProvider.get_id("_enable_daily")}"
                                                    ${checked(curTorrentProvider.daily_enabled)}
-                                                   ${disabled(not curTorrentProvider.can_daily)}
+                                                   ${disabled(curTorrentProvider.can_daily)}
```
miigotu, line 661: this goes the other way. if true, do not disable, if false then disable, so if it cant daily, disable the not has to stay here

sickchill/gui/slick/views/editShow.mako
```diff
@@ -247,7 +247,7 @@
                                         <span class="component-title">${_('Season folders')}</span>
                                     </div>
                                     <div class="col-lg-9 col-md-8 col-sm-7 col-xs-12 component-desc">
-                                        <input type="checkbox" id="season_folders" name="season_folders" title="season_folders" ${checked(show.season_folders or settings.NAMING_FORCE_FOLDERS)} ${disabled(settings.NAMING_FORCE_FOLDERS)} />
+                                        <input type="checkbox" id="season_folders" name="season_folders" title="season_folders" ${checked(show.season_folders or settings.NAMING_FORCE_FOLDERS)} ${disabled(not settings.NAMING_FORCE_FOLDERS)} />
```
miigotu, line 250: good find on this one

sickchill/gui/slick/views/config_providers.mako
BKSteve: That's not the way it's reacting. I'll look at this some more later.
can_daily seems to be actually cannot and the * ! matching of prior page.

miigotu: Hmm you might be right, this method is confusing because it does the opposite of what you expect.

disabled(True) = ""
disabled(False) = "disabled"

So can_daily = True (you want to be able to select it)

`disabled(provider.can_daily())`

So it might be the other one that's backwards, let me look again.

Also, probably because this is so confusing I need to rename the function so it isn't so confusing and backwards.

BKSteve: Propose in below `no provider disable` (should have called hide and not disable) to not hide the backlog search, leave it show as disabled.
This way anybody wishing to advise otherwise can see the backlog function and daily function disabled.
This matches best with current `master` and as UI changing fix any naming logic later.

Add: If you agree with this path then all have been checked and its ready. Unless you want another look at IMDB functionality.

sickchill/gui/slick/views/config_providers.mako
```diff
@@ -197,7 +197,7 @@
                                 % endif
 
                                 % if hasattr(curNewznabProvider, 'enable_backlog'):
-                                    <div class="field-pair row${hidden(curNewznabProvider.supports_backlog)}">
+                                    <div class="field-pair row">
```
miigotu, line 200: This one needs to be there, it just needed a space between row and $

sickchill/gui/slick/views/config_providers.mako
```diff
@@ -310,7 +310,7 @@
                                 % endif
 
                                 % if hasattr(curNzbProvider, 'enable_backlog'):
-                                    <div class="field-pair row${hidden(curNzbProvider.supports_backlog)}">
+                                    <div class="field-pair row">
```
miigotu, line 313: Same here

sickchill/gui/slick/views/config_providers.mako
miigotu: This one is correct, I think the other one is wrong, looking now

sickchill/gui/slick/views/config_providers.mako
```diff
@@ -669,7 +669,7 @@
                                 % endif
 
                                 % if hasattr(curTorrentProvider, 'enable_backlog'):
-                                    <div class="field-pair row${hidden(curTorrentProvider.supports_backlog)}">
+                                    <div class="field-pair row">
```
miigotu, line 672: Same here, needs to be there just add the space between row and $

sickchill/gui/slick/views/editShow.mako
miigotu: This one is correct.

sickchill/gui/slick/views/config_postProcessing.mako
```diff
@@ -186,17 +186,14 @@
                             </div>
                         </div>
 
-                        <div class="field-pair row" id="content_move_associated_files" ${hidden(settings.MOVE_ASSOCIATED_FILES)}>
+                        <div class="field-pair row" id="content_move_associated_files" ${selected(settings.MOVE_ASSOCIATED_FILES)}>
```
miigotu, line 189: Good here

sickchill/gui/slick/views/config_postProcessing.mako
```diff
@@ -186,17 +186,14 @@
...
                             <div class="col-lg-3 col-md-3 col-sm-4 col-xs-12">
                                 <label class="component-title">${_('Rename .nfo file')}</label>
                             </div>
                             <div class="col-lg-9 col-md-9 col-sm-8 col-xs-12 pull-right component-desc">
                                 <input type="checkbox" name="nfo_rename" id="nfo_rename" ${checked(settings.NFO_RENAME)}/>
                                 <label for="nfo_rename">${_('rename the original .nfo file to .nfo-orig to avoid conflicts?')}</label>
                             </div>
-                        </div>
```
miigotu, line 197: These should stay in

miigotu (changes requested):

sickchill/gui/slick/views/config_providers.mako
BKSteve: I like the idea that it's shown and disabled rather than hiding it.
More consistent with `can_daily`

sickchill/gui/slick/views/config_postProcessing.mako
BKSteve: why show the list if its not enabled?

sickchill/gui/slick/views/config_providers.mako
BKSteve: there's 3 of these so marked the other conversation resolved.

sickchill/gui/slick/views/config_postProcessing.mako
miigotu: idk, its fine I guess.

PR #8612, opened 2023-11-04
Some Possible Updates

BKSteve:
PyCharm error list.
Going through some items and clearing errors and warnings.

Reject, comment, accept, open house.

## Summary by CodeRabbit

- **Refactor**
	- Improved thread handling in certain classes for better performance.
	- Updated URL parsing method in the `Client` class for more accurate host and port extraction.
- **Chores**
	- Removed unused `validators` module imports across multiple provider files.
	- Updated type imports in `tv.py` for better type checking.

### DIFF
--- a/sickchill/adba/__init__.py
+++ b/sickchill/adba/__init__.py
@@ -87,7 +87,7 @@ class Connection(threading.Thread):
 
         # thread keep alive stuff
         self.keepAlive = keep_alive
-        self.setDaemon(True)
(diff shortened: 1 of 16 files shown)


BKSteve:
> Changed the way daemon threads are set from self.setDaemon(True) to self.daemon = True.

This one is a python 3.8+ as far as I know. Now more 3.7

miigotu:
These are all good, thanks!

PR #8616, opened 2023-11-07
Directories setting

BKSteve:
New, Edit, Delete, Default settings of directories.

## Summary by CodeRabbit

- **New Features**
	- Modified the data transmission method for root directories, enhancing server communication.
- **Bug Fixes**
	- Adjusted the handling of root directories for improved data retrieval.
	- Updated the filtering condition for directory list items, ensuring only new items are processed.

### DIFF
--- a/sickchill/gui/slick/js/rootDirs.js
+++ b/sickchill/gui/slick/js/rootDirs.js
@@ -203,7 +203,7 @@ $(document).ready(() => {
         }
 
         refreshRootDirs();
-        $.get(scRoot + '/config/general/saveRootDirs', {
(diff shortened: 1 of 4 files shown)


miigotu:
I'll get around to this soon I hope, I'm not seeing what this changes from the diff here I'm assuming that some places used get and others post for the root dirs?

BKSteve:
Correct of get and post. I aligned them so all were post (2 gets changed) - rootDirs.js
Inverted the logic of the lambda on add existing shows to show those `not` in the db rather than are in the db. - home_massAddTable.mako
Moved when you get `whichSeries` so it gets after you've selected a show and not before the skip show button - add_shows.py

PR #8628, opened 2023-11-19
QBT remove false link to torrent folder.

BKSteve:
Fixes a QBT folder error where BlackHole path passed to QBT rather than using QBT defined folders.
https://discord.com/channels/502612977271439372/1173806065427697775

## Summary by CodeRabbit

- **New Features**
  - Enhanced qBittorrent integration with new options for torrent path and incomplete path settings.
  - Improved file browser visibility for torrent path selection.

- **Refactor**
  - Updated logging practices for better clarity and debugging, including a shift from `info` to `debug` level and adoption of f-string formatting.

- **Bug Fixes**
  - Fixed issues with torrent path settings to align with user-configured directories.

### DIFF
--- a/sickchill/oldbeard/clients/qbittorrent.py
+++ b/sickchill/oldbeard/clients/qbittorrent.py
@@ -47,8 +47,8 @@ class Client(GenericClient):
             download_path=settings.TORRENT_PATH_INCOMPLETE or None,
             category=(settings.TORRENT_LABEL, settings.TORRENT_LABEL_ANIME)[result.show.is_anime] or settings.TORRENT_LABEL,
             is_paused=settings.TORRENT_PAUSED,
-            ratio_limit=float(result.ratio) or None,
-            seeding_time_limit=(None, 3600 * int(settings.TORRENT_SEED_TIME))[bool(settings.TORRENT_SEED_TIME)]
+            ratio_limit=(None, float(result.ratio))[float(result.ratio) > 0],
+            seeding_time_limit=(None, 3600 * int(settings.TORRENT_SEED_TIME))[int(settings.TORRENT_SEED_TIME) > 0]
             tags=("sickchill", "sickchill-anime")[result.show.is_anime]
         )
 
--- a/sickchill/oldbeard/clients/qbittorrent.py
+++ b/sickchill/oldbeard/clients/qbittorrent.py
@@ -41,23 +41,38 @@ class Client(GenericClient):
         return True, "Success: Connected and Authenticated"
 
     @staticmethod
-    def __torrent_args(result: "TorrentSearchResult"):
+    def __torrent_args(result: "TorrentSearchResult") -> dict:
         return dict(
             save_path=settings.TORRENT_PATH or None,
             download_path=settings.TORRENT_PATH_INCOMPLETE or None,
             category=(settings.TORRENT_LABEL, settings.TORRENT_LABEL_ANIME)[result.show.is_anime] or settings.TORRENT_LABEL,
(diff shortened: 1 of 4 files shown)


miigotu:
I don't agree with this. I think that the folder needs to be shown in the js, where we hide and show different settings depending on what client it is. If they don't want it in a folder either they should empty that field, or we need to add an option to set whether they want to use labeler or our path setting . This was a new feature, we weren't able to send a path in the past. I'm curious if there is an incomplete_path option for qBitTorrent API also, it would be nice to be able to pass that.

miigotu:
Need to change this line: https://github.com/SickChill/sickchill/blob/fa39f385b7edf8247b4d24f0b54eb1f114249435/sickchill/gui/slick/js/core.js#L1931

```diff
--- $('#torrent_path_option').hide();
+++ $('#torrent_path_option').show();
+++ $('#torrent_path_option').find('.fileBrowser').show();
```

What I'm also curious about is if qBitTorrent xmlrpc can accept a `$('#torrent_path_incomplete_option').show()`

miigotu:
torrent_path_incomplete_option can be coded in for qBitTorrent:
Screenshot_20231120-000310

BKSteve:
Yes. That's a more complete solution giving the options. The new other torrent items can probably do this too. 
Will do a more detailed push after review

BKSteve:
`save_path` can be set, 
setting `temp_path` no success, ~~if setting enable in qbt or not, still unable to get qbt to accept temp location. Looks to be a read only parameter or I'm doing something wrong.~~

Add: https://qbittorrent-api.readthedocs.io/en/latest/apidoc/torrents.html
`download_path` test soon.

BKSteve:
clean up too many info logs for DownloadStation, converted to debug.

sickchill/oldbeard/clients/qbittorrent.py
BKSteve, line 1: add this debug logging to others?

BKSteve:
@coderabbitai review

BKSteve:
@coderabbitai pause

sickchill/oldbeard/clients/qbittorrent.py
```diff
@@ -41,20 +41,38 @@ def test_client_connection(self):
...
+        logger.debug(f"Posting as file with {self.__torrent_args(result)}")
+        action = self.api.torrents_add(torrent_files=[result.content], **self.__torrent_args(result))
+        self.__add_trackers(result)
+        return action
 
     def _set_torrent_priority(self, result: "TorrentSearchResult"):
         if not self.api.app_preferences().get("queueing_enabled"):
             return True
-        return (self.api.torrents_decrease_priority, self.api.torrents_increase_priority)[result.priority == 1](result.hash.lower)
+        return (self.api.torrents_decrease_priority, self.api.torrents_increase_priority)[result.priority == 1](result.hash.lower())
```
miigotu, line 77: These missing parenthesis are a really old bug o can't believe I missed this all those years.

sickchill/oldbeard/clients/qbittorrent.py
miigotu: Sure, this was added though when their API changes between dsm5 and 6 I think though. Up to you.

BKSteve: I meant add `logging.debug` for uT and rT etc. so there is some data if any of them aren't working.

miigotu: By all means, if you think it might be helpful at another time for debugging go for it.

Issue #8688, opened 2024-01-14
Zamunda Provider not working


joncevski:
2024-01-14 19:22:24 INFO :: SEARCHQUEUE-BACKLOG-366924 :: [Zamunda] :: No NZB/Torrent providers found or enabled in the sickchill config for backlog searches. Please check your settings.
AA TypeError: Provider.search() got an unexpected keyword argument 'episode_object'
AA ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AA items_list += self.search(search_string, episode_object=episode)
AA File "/opt/sickchill/lib/python3.11/site-packages/sickchill/providers/GenericProvider.py", line 162, in find_search_results
AA ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AA search_results = curProvider.find_search_results(show, episodes, search_mode, manual, downCurQuality)
AA File "/opt/sickchill/lib/python3.11/site-packages/sickchill/oldbeard/search.py", line 459, in search_providers
2024-01-14 19:22:24 ERROR :: SEARCHQUEUE-BACKLOG-366924 :: [Zamunda] :: Exception while searching Zamunda. Error: Provider.search() got an unexpected keyword argument 'episode_object'

BKSteve:
Start with rectifying this
`No NZB/Torrent providers found or enabled in the sickchill config for backlog searches. Please check your settings.`

miigotu:
It's saying none are enabled because that one errored out, I missed removing something in a refactor

miigotu:
@BKSteve sickchill/providers/GenericProvider.py", line 162, remove the episode_object parameter and this one is fixed

PR #8691, opened 2024-01-16
Generic Provide + Notifications

BKSteve:
Fix Generic Provider `episode_object` error
#8688

## Summary by CodeRabbit

- **New Features**
	- Added `DISCORD_WEBHOOK` variable to settings for enhanced Discord integration.
	- Updated URL for "opensubtitles" source to use HTTPS for improved security.
- **Bug Fixes**
	- Corrected the logic in the `get_coming_episodes` function to ensure it always returns results.
- **Refactor**
	- Simplified Discord notification logic by removing the `tts` parameter and related settings.
	- Improved error logging in Discord notification sending.
- **Chores**
	- Removed the "Discord TTS" checkbox from the GUI to streamline notification settings.
- **Style**
	- Adjusted search results retrieval logic in `GenericProvider.py` for clarity and efficiency.


BKSteve:
Remove Discord TTS setting and code as it doesn't TTS and message rather than speech is sufficient notification.
Otherwise requires much bigger rewrite.

BKSteve:
Add in change of `http://www.opensubtitles.org` to `https://www.opensubtitles.com`

sickchill/views/config/notifications.py
```diff
@@ -368,7 +367,6 @@ def saveNotifications(
         settings.DISCORD_WEBHOOK = discord_webhook
         settings.DISCORD_NAME = discord_name
         settings.DISCORD_AVATAR_URL = discord_avatar_url
-        settings.DISCORD_TTS = discord_tts
```
miigotu, line 371: This is the only line needed changed for discord_tts, to `settings.DISCORD_TTS = config.checkbox_to_value(discord_tts)`

Tts is for blind people, and we have sight-impaired users. They just don't talk much.

BKSteve:
`Discord TTS big change` fixes the TTS setting and realigns it and improves selection to enable the whole message or just the bot name to be sent.
The Bot name can be anything you want as actual bot is based on web hook. It just can't be blank so it gets forced to SickChill if blank.

With this testing have added in `index.py` try around `await self.finish(results)` as was throwing an exception with all shows regarding `prowl_notify_list` which core.js says is an `Unresolved variable` on 1228.
Need some assistance with this js please.

sickchill/gui/slick/js/core.js
```diff
@@ -716,11 +716,11 @@ const SICKCHILL = {
 
                 discord.name = $('#discord_name').val();
                 discord.avatar = $('#discord_avatar_url').val();
-                discord.tts = $('#discord_tts').val();
+                discord.tts = $('#discord_tts').is(':checked') ? 1 : 0;
```
miigotu: You don't need the ternary here, you can just do:
```javascript
$('#discord_tts').is(':checked');
```

`checkbox_to_value` will handle the correct conversion.

BKSteve: Saw this used elsewhere in core.js and as I couldn't get tts to work with the returned value and store this aligned with the same `true` or `false` values.  
I therefore went with 1 or 0 like many other settings and just simple int when passed from mako, js, python and
`settings.DISCORD_TTS = check_setting_bool(settings.CFG, "Discord", "discord_tts")` as stored are all aligned.

PR #8718, opened 2024-02-18
root dirs, trakt, addshow, bdecode

BKSteve:
Fixes 

- rootDir reference numbering (must have come with the js) aligns with actual position of full list no longer needing +1, removes startup error if only one entry
- trakt interaction and data sharing
- addshow bunch of cleans and trakt link.
- torrent_bdecode b"info not found so change to "info"

## Summary by CodeRabbit

- **Refactor**
	- Updated the `traktChecker` class to enhance the show tracking process without requiring a `force` update.
	- Improved show search and addition flow by streamlining parameter handling and variable naming.
	- Corrected video root directory assignment in server settings for accurate path configuration.

### DIFF
--- a/sickchill/views/server_settings.py
+++ b/sickchill/views/server_settings.py
@@ -70,7 +70,7 @@ class SCWebServer(threading.Thread):
         # video root
         if settings.ROOT_DIRS:
             root_dirs = settings.ROOT_DIRS.split("|")
-            self.video_root = root_dirs[int(root_dirs[0]) + 1]
+            self.video_root = root_dirs[int(root_dirs[0])]
         else:
             self.video_root = None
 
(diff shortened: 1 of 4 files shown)


sickchill/views/server_settings.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, options=None):
         # video root
         if settings.ROOT_DIRS:
             root_dirs = settings.ROOT_DIRS.split("|")
-            self.video_root = root_dirs[int(root_dirs[0]) + 1]
+            self.video_root = root_dirs[int(root_dirs[0])]
```
miigotu, line 73: why change this one? we dont use video_root

BKSteve: The first directory location no longer stored as 0 but 1 so adding 1 to 1 when only 1 entry errors at startup

miigotu: Ah yeah I was seeing the diff in reverse. That little section needs removed, video_root was a strange thing echelon added at some point where your videos could be watched remotely without a password even, it didn't require authentication. Idk if he was scraping other people's videos or what.

PR #8756, opened 2024-05-01
Pytest 8.2.0 errors

BKSteve:
Per https://github.com/pytest-dev/pytest/issues/12263

Work around to avoid `AttributeError: property 'is_active' of 'Provider' object has no setter`
in test_snatch.

Plus ignore the `Warning` about `telnetlib` whilst looking why the build system was erroring

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ flake8-quotes = "^3.3.0"
 flake8-docstrings = "^1.6.0"
 flake9 = "^3.8.3"
 codecov = "^2.1.12"
-pytest = ">=7.0.1,<9.0.0"
+pytest = ">=7.0.1,<9.0.0,!=8.2.0"
 isort = "^5.10.1"
 poethepoet = ">=0.16,<0.27"
 po2json = "^0.2.2"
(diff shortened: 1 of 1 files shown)


BKSteve:
Seriously anybody still got a Popcorn Hour 200 series? It's like a 2011 media player.

Should just deprecate this one.

twinden:
@BKSteve Yelp !
Hi Steve !
I can’t find where to report an issue with SickChill on this all new GitHub.
I installed the new SC package yesterday, and it won’t become running on my Synology Evansport DS214play on DSM 6.2.

Can you register a new issue for this and mention me there ?!
Can you please look into this ?

I’ll delete this here when read. Sorry for posting here, I just can’t find “new issue” anywhere.

miigotu:
Instead of blocking pytest8.2.0, why not add the setter?

Under
```python
@attribute
def is_active(self): ...
```
Add something like
```python
@is_active.setter
def is_active(self, value):
    self.active = value
```

BKSteve:
added to Develop via `guess2` merge. 
Closed

miigotu:
> Instead of blocking pytest8.2.0, why not add the setter?
> 
> Under
> 
> ```python
> @attribute
> def is_active(self): ...
> ```
> 
> Add something like
> 
> ```python
> @is_active.setter
> def is_active(self, value):
>     self.active = value
> ```

I still want this fix, we don't want to block pytest 8.2.0, we want to fix the error (can't setattr to properties that don't have setters in testing, which is a good thing)

PR #8758, opened 2024-05-07
Support list of names passed to search in tvdb.py

canepan:
Currently, if a `list` is passed instead of a single `str`, it will fail the regex match

Fixes #

- process each item if a list is passed


BKSteve:
What was SC doing at the time the `list` error was thrown?

sickchill/show/indexers/tvdb.py
```diff
@@ -110,38 +110,42 @@ def search(self, name, language=None, exact=False, indexer_id=False):
...
-                    names.append(test.group(1).strip())
-                # Name with spaces
-                if re.match(r"[. -_]", name):
-                    names.append(re.sub(r"[. -_]", " ", name).strip())
-                    if test:
-                        # Name with spaces and without year
-                        names.append(re.sub(r"[. -_]", " ", test.group(1)).strip())
-
-            for attempt in set(n for n in names if n.strip()):
+        if isinstance(name, str):
```
miigotu, line 111: Shouldn't this join the list like the other guessit fixes instead of just checking if it's a string?

On line 1 add:
```python
try:
    from collections import Iterable
except (ModuleNotFoundError, ImportError):
    # Iterable moved in python3.10
    from collections.abc import Iterable
```
After line 106 add:
```python
if isinstance(name, Iterable) and not isinstance(name, (str, bytes);
    name = " ".join(name)
```

Then remove the rest of the changes here (we know it's a string then)

miigotu (changes requested):

sickchill/show/indexers/tvdb.py
```diff
@@ -108,38 +108,42 @@ def search(self, name, language=None, exact=False, indexer_id=False):
...
+                            names.append(re.sub(r"[. -_]", " ", test.group(1)).strip())
+    
+                for attempt in set(n for n in names if n.strip()):
+                    try:
+                        result = self._search(attempt, language=language)
+                        if result:
+                            break
+                    except (requests.exceptions.RequestException, requests.exceptions.HTTPError, Exception):
+                        logger.debug(traceback.format_exc())
+        else:
```
miigotu, line 144: Also this else needs removed. We don't want to search for ["fast", "&", "furious"] each individually, we want to join that list and search for the full string.

canepan:
What about instead enabling mypy type checking and avoid a series of checks used probably by a single caller?

⁣Nicola Canepa​

Il giorno 12 mag 2024, 14:41, alle ore 14:41, miigotu ***@***.***> ha scritto:

miigotu:
This has been fixed in #8759

PR #8772, opened 2024-06-20
Sanitize more args

BKSteve:
continue with mako - python code work

This needs some checking for `retry` items to ensure `query` and not `body` but looks consistent.


BKSteve:
@coderabbitai review

PR #8786, opened 2024-08-10
fix: use new function containsAllWords to check for required_words

Tensai75:
- includes new function containsAllWords
- required words are checked with the new function containsAllWords
- the release is only accepted if all required words are present

## Summary by CodeRabbit

- **New Features**
	- Introduced a new word filtering function that checks if all specified words are present in a release name, enhancing the filtering process.

- **Improvements**
	- Updated logging to provide clearer feedback when required words are missing, improving user understanding of filtering decisions.

### DIFF
--- a/sickchill/oldbeard/search.py
+++ b/sickchill/oldbeard/search.py
@@ -178,7 +178,7 @@ def pick_best_result(results, show):
     picked_result = None
 
     # order the list so that preferred releases are at the top
-    results.sort(key=lambda ep: show_name_helpers.hasPreferredWords(ep.name, ep.show), reverse=True)
+    results.sort(key=lambda ep: show_name_helpers.has_preferred_words(ep.name, ep.show), reverse=True)
 
     # find the best result for the current episode
     for result in results:
--- a/sickchill/oldbeard/show_name_helpers.py
+++ b/sickchill/oldbeard/show_name_helpers.py
@@ -34,6 +34,26 @@ def containsAtLeastOneWord(name, words):
             return word
     return False
 
+def containsAllWords(name, words):
+    """
+    Filters out results based on filter_words
+
+    name: name to check
+    words : string of words separated by a ',' or list of words
+
+    Returns: True if all words from the list are present in name, or the first word from the list not found in name.
+    """
+    if isinstance(words, str):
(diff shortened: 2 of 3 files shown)


BKSteve:
@coderabbitai review

miigotu:
This is a breaking change. Required words is not supposed to be ALL words matching, it's one of them. Which is why if you want to match all words you use a regex in the setting. You would have to change the UI and add a 3rd option to change that functionality. There are many instances where people want to ignore or require only 1 word out of their list of words. The setting allows for that, or using vertical bars to separate the words instead of commas if you want it to match ANY of those words. "Sickchill|media|movie" vs "sickchill,media,movie" 

Maybe just adding better examples and documentation would be better than this PR imo.

Tensai75:
> This is a breaking change. Required words is not supposed to be ALL words matching, it's one of them. Which is why if you want to match all words you use a regex in the setting.

Are you serious?
1. you can't currently use regular expressions. In the `containsAtLeastOneWord()` function the words are escaped using `re.escape(word)` (see https://github.com/SickChill/sickchill/blob/a1d1b3ec619376b147b7ee21d0fd8917e779cabe/sickchill/oldbeard/show_name_helpers.py#L35)
2. even if you could, there is no simple regex to make sure that all of a list of words are present, especially if you don't know the order in which they appear.
3. I bet, if you ask 10 ordinary people, 9 out of them would say a list of 'required words' means that all words must be present. This is how everyone would understand it.
4. Even the comment in the code says so: `# if any of the good strings aren't in the name then say no`. It clearly states "__if _any_ of [...] is _not_ present__" (see https://github.com/SickChill/sickchill/blob/a1d1b3ec619376b147b7ee21d0fd8917e779cabe/sickchill/oldbeard/show_name_helpers.py#L79C5-L79C64)

The PR just implements the expected behaviour which was supposed to be there from the beginning.

P.S.: It migth be breaking, but only for those people who relaid on this faulty behaviour of this setting. But they can probably achieve the same result by using 'preferred words'.

BKSteve:
Per the edit show text 
`search results with no words from this list will be ignored.`
and settings episode search
`results with no word from this list will be ignored`
This is clearly an OR.

Closing

PR #8811, opened 2024-11-18
Fix GHSL-2024-288

sylwia-budzynska:
- change the login page to redirect to `settings.DEFAULT_PAGE` instead of to the `next` parameter, which can be user-controlled. Please see GHSL-2024-288 (issue 5 in the report) for more information.

## Summary by CodeRabbit

- **New Features**
	- Simplified login redirect logic to always direct users to the default page after login attempts.

- **Bug Fixes**
	- Maintained error message handling for login attempts without changes.

### DIFF
--- a/sickchill/views/authentication.py
+++ b/sickchill/views/authentication.py
@@ -28,8 +28,7 @@ class LoginHandler(BaseHandler):
             logger.warning(_("User attempted a failed login to the SickChill web interface from IP: ") + self.request.remote_ip)
             login_error = _("Incorrect username or password! Both username and password are case sensitive!")
 
-        next_ = self.get_query_argument("next", next_)
-        self.redirect(next_ or "/" + settings.DEFAULT_PAGE + "/")
+        self.redirect("/" + settings.DEFAULT_PAGE + "/")
 
 
 class LogoutHandler(BaseHandler):


miigotu:
User controllable variables are fine. No matter what you put in there you can't access SC without logging in. It's a single user system. 

If I'm wrong about that, show me how you can access anything without logging in or using a cookie or bypass method outside of this parameter. IMHO this is a false positive.

sylwia-budzynska:
>User controllable variables are fine. No matter what you put in there you can't access SC without logging in. It's a single user system.

This issue _does not_ lead to an attacker accessing anything without logging in. This issue allows for redirecting a user to an attacker-controlled website and could be used in phishing attempts. The impact and PoC for this issue is described in the report I sent to you via email in issue number 5. I don't want to share more details in here, because it would be not in line with coordinated disclosure policy. 

> If I'm wrong about that, show me how you can access anything without logging in or using a cookie or bypass method outside of this parameter. IMHO this is a false positive.

I take that SC gets a lot of bogus reports. This is a real bug (see PoC in the report), but it _does not_ lead to accessing anything without logging in. That said, if you don't consider this a bug, feel free to close this PR.

miigotu:
I'll accept the PR, but there is no way an attacker can exploit this without a login. If you can't login, the next parameter can't be used by anyone.

PR #8901, opened 2026-04-24
fix: fix Docker container crash (enzyme pkg_resources) and improve CI diagnostics

MattPark:
## Summary

The Docker container has been crashing immediately on startup in CI with:
```
ModuleNotFoundError: No module named 'pkg_resources'
```

**Root cause:** When building from source (`SOURCE=1`), the Dockerfile ran `pip install --find-links=./dist "sickchill[speedups]"`. Because the locally-built wheel and the published PyPI wheel share the same version (`2024.3.1`), pip resolved dependencies against the PyPI package's metadata (`enzyme<0.5.0`) instead of the local wheel's (`enzyme<0.6.0`). This pulled enzyme 0.4.1, which imports `pkg_resources` at runtime without declaring `setuptools` as a dependency.

**Fixes:**
- Bump enzyme floor from `>=0.4.1` to `>=0.5.0` — enzyme 0.5.0+ replaced `pkg_resources` with `importlib.resources`
- Install local wheel by explicit path (`$(ls ./dist/sickchill-*.whl)[speedups]`) to force pip to use the locally-built metadata
- Remove `--rm` from `docker run` in CI so container logs survive crashes
- Add early container exit detection with `docker logs` capture on failure
- Increase `SLEEP` from 60s to 120s for the readiness polling loop

## Testing

Verified in fork CI (3 iterations):
1. First run: captured crash logs confirming `pkg_resources` error (previously invisible due to `--rm`)
2. Second run: confirmed enzyme constraint alone wasn't enough (pip still used PyPI metadata)
3. Third run: **container starts in ~3s, both health checks pass** ✅

CI run: https://github.com/MattPark/sickchill/actions/runs/24891154955

## Summary by CodeRabbit

Infrastructure and dependency updates to improve build reliability and system compatibility.

* **Chores**
  * Enhanced Docker test container handling with improved error detection, automatic failure logging, and consistent cleanup procedures across success and failure paths
  * Updated enzyme package version constraint to 0.5.0+ for improved compatibility and stability


MattPark:
@BKSteve got it I was able to check this on my fork and it solves it.

PR #8926, opened 2026-05-08
fix: use video.subtitles.extend for external subtitle discovery

MattPark:
## Summary
- Fix `AttributeError` when scanning external subtitles with subliminal 2.x

## Problem
`subliminal.core.search_external_subtitles()` changed its return type in subliminal 2.x:
- **Old** (subliminal 1.x): returns `dict[str, Language]`
- **New** (subliminal 2.x): returns `dict[str, ExternalSubtitle]`

Additionally, `video.subtitle_languages` changed from a mutable `set` attribute to a **read-only property** computed from `video.subtitles`. The old code:

```python
video.subtitle_languages |= set(search_external_subtitles(...).values())
```

fails with `AttributeError` because you can't `|=` into a property, and the values are `ExternalSubtitle` objects not `Language` objects.

## Fix
```python
video.subtitles.extend(search_external_subtitles(...).values())
```

This adds the `ExternalSubtitle` objects to `video.subtitles` (a mutable list), and the `subtitle_languages` property correctly derives the `Language` set from them.

## Context
Reported via [Discord](https://discord.com/channels/502612977271439372/1132477941436125256), flagged by @BKSteve on #8924.

## Summary by CodeRabbit

* **Bug Fixes**
  * External subtitle discovery now correctly registers discovered subtitle files so they appear alongside a video's existing subtitles.

* **Tests**
  * Added unit tests to verify external subtitles are detected and attached, support multiple matches, and are skipped when external subtitle scanning is disabled.

### DIFF
--- a/sickchill/oldbeard/subtitles.py
+++ b/sickchill/oldbeard/subtitles.py
@@ -303,7 +303,7 @@ def get_video(video_path, subtitles_path=None, subtitles=True, embedded_subtitle
 
         # external subtitles
         if subtitles:
-            video.subtitle_languages |= set(subliminal.core.search_external_subtitles(video_path, directory=subtitles_path).values())
+            video.subtitles.extend(subliminal.core.search_external_subtitles(video_path, directory=subtitles_path).values())
 
         if embedded_subtitles is None:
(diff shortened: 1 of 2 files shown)


MattPark:
Superseded by #8927 which includes this fix along with 5 other subliminal 2.x compatibility fixes and 14 regression tests.

PR #8932, opened 2026-05-10
fix: add existing shows search term becomes '1' on second show

MattPark:
## Summary
- Fix bug where adding multiple existing shows from disk causes the search term to be "1" for the second and subsequent shows
- Root cause: `get_body_argument("other_shows")` (singular) returns a single string instead of a list when the form has multiple hidden inputs with the same name
- `other_shows[0]` on a string returns the first character ("1" from the pipe-delimited format `1|/path|id|name`) instead of the first show entry
- Changed to `get_body_arguments("other_shows")` (plural) which correctly returns a list of all form values

## Changes
- `sickchill/views/manage/add_shows.py:433` — `addNewShow()`: `get_body_argument` → `get_body_arguments`
- `sickchill/views/manage/add_shows.py:176` — `newShow()`: same fix for the alternate code path

Reported via [Discord thread](https://discord.com/channels/502612977271439372/1502753012752449637).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of multiple shows when adding new shows to the application.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SickChill/sickchill/pull/8932)


BKSteve:
lol, what a fundamental faux pas. Probably worked for the first 2 then lost it.

